PR #72, opened 2014-09-05
[JCLOUDS-702] JCloud ProfitBricks provider implementation

devcsrj:
Hi! This implements most of ProfitBrick's API. I look forward for your comments, suggestions, and revisions.


profitbricks/src/test/java/org/jclouds/profitbricks/compute/internal/ProvisioningStatusPollingPredicateTest.java
```diff
@@ -0,0 +1,73 @@
...
+
+      Predicate<String> waitUntilAvailable = Predicates2.retry(
+              new ProvisioningStatusPollingPredicate(pbApi, ProvisioningStatusAware.DATACENTER, ProvisioningState.AVAILABLE),
+              30l, 1l, TimeUnit.SECONDS);
+
+      String id = "aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee";
+      try {
+         waitUntilAvailable.apply(id);
+         ProvisioningState finalState = pbApi.dataCenterApi().getDataCenterState(id);
+         assertEquals(finalState, ProvisioningState.AVAILABLE);
```
nacx: Also verify the requests that have been made?

profitbricks/src/main/java/org/jclouds/profitbricks/features/DataCenterApi.java
```diff
@@ -0,0 +1,132 @@
...
+    * further request can be performed on it.
+    * <p>
+    * @param identifier Identifier of the virtual data center
+    * @return Returns a boolean indicating whether delete operation was made
+    */
+   @POST
+   @Named("datacenter:delete")
+   @Payload("<ws:deleteDataCenter><dataCenterId>{id}</dataCenterId></ws:deleteDataCenter>")
+   @Fallback(Fallbacks.FalseOnNotFoundOr404.class)
+   Boolean deleteDataCenter(@PayloadParam("id") String identifier);
```
nacx: Better return a primitive type, as `null` does not make sense here.

profitbricks/src/test/java/org/jclouds/profitbricks/features/DataCenterApiMockTest.java
```diff
@@ -0,0 +1,236 @@
...
+      try {
+         List<DataCenter> dataCenters = api.getAllDataCenters();
+         assertRequestHasCommonProperties(server.takeRequest());
+         assertNotNull(dataCenters);
+         assertEquals(dataCenters.size(), 2);
+      } finally {
+         pbApi.close();
+         server.shutdown();
+      }
+   }
```
nacx: Add also a test that verifies what happens when the server returns a 404. There is an explicit fallback in the method definition, so we should test it is taking place when needed.

profitbricks/src/test/java/org/jclouds/profitbricks/features/DataCenterApiMockTest.java
```diff
@@ -0,0 +1,236 @@
...
+   public void testGetNonExistingContainer() throws Exception {
+      MockWebServer server = mockWebServer();
+      server.enqueue(new MockResponse().setResponseCode(500).setBody(payloadFromResource("/fault-404.xml")));
+
+      ProfitBricksApi pbApi = api(server.getUrl(rootUrl));
+      DataCenterApi api = pbApi.dataCenterApi();
+
+      String id = "random-non-existing-id";
+      try {
+         DataCenter dataCenter = api.getDataCenter(id);
```
nacx: Fail the test here, otherwise the test will succeed even if the exception is not thrown.

profitbricks/src/test/java/org/jclouds/profitbricks/features/DataCenterApiMockTest.java
```diff
@@ -0,0 +1,236 @@
...
+
+   @Test
+   public void testCreateDataCenterWithIllegalArguments() throws Exception {
+      String[] names = {"JCl@ouds", "JC|ouds", "^clouds", ""};
+      for (String name : names)
+         try {
+            DataCenter.Request.CreatePayload.create(name, Location.US_LAS);
+            fail("Should have failed for name: ".concat(name));
+         } catch (Exception ex) {
+            assertTrue(ex instanceof IllegalArgumentException);
```
nacx: Why not just catching the `IllegalArgumentException`?

profitbricks/src/test/java/org/jclouds/profitbricks/internal/BaseProfitBricksMockTest.java
```diff
@@ -0,0 +1,87 @@
...
+         return toStringAndClose(getClass().getResourceAsStream(resource)).getBytes(Charsets.UTF_8);
+      } catch (IOException e) {
+         throw Throwables.propagate(e);
+      }
+   }
+
+   protected static void assertRequestHasCommonProperties(final RecordedRequest request) {
+      assertEquals(request.getMethod(), "POST");
+      assertEquals(request.getHeader(HttpHeaders.AUTHORIZATION), authHeader);
+      assertEquals(request.getHeader(HttpHeaders.ACCEPT), MediaType.TEXT_XML);
```
nacx: We want to make sure that every request is made to the right endpoint. In this case, just add here a common assertion to verify all requests are done to "/"?

nacx:
Good job @devcsrj! There just a few minor comments about the tests. Once they are addressed this will be good to go!

profitbricks/src/test/java/org/jclouds/profitbricks/features/DataCenterApiMockTest.java
devcsrj: Hmm. I was thinking that if (for some reason) <code>CreatePayload#create</code> throws a different kind of exception, it would skip the <code>fail()</code> and <code>catch</code> clause; thus, the test becoming a success. So... should just <code>IllegalArgumentException</code> be okay?

profitbricks/src/test/java/org/jclouds/profitbricks/features/DataCenterApiMockTest.java
```diff
@@ -0,0 +1,257 @@
...
+
+      ProfitBricksApi pbApi = api(server.getUrl(rootUrl));
+      DataCenterApi api = pbApi.dataCenterApi();
+
+      String id = "random-non-existing-id";
+      try {
+         DataCenter dataCenter = api.getDataCenter(id);
+         assertRequestHasCommonProperties(server.takeRequest());
+         assertNull(dataCenter);
+      } catch (Exception ex) {
```
nacx: [minor] This is redundant. Test will already fail if the method throws an exception.

nacx:
Thanks! There is just one minor comment. If you squash the commits I'll merge the PR. Great job!

pongstr:
:+1:  Great job @devcsrj

nacx:
Pushed to master as [6e1a037](https://git-wip-us.apache.org/repos/asf?p=jclouds-labs.git;a=commit;h=6e1a0376f837e2dfe765758333e09761ff39b9b4). Thanks @devcsrj!

PR #113, opened 2014-11-25
support for docker 1.3.2

andreaturli:
- update value objects (Container, Config)
- add new value object (Resource, StatusCode)
- add more options for ContainerApi
- add support for pause/unpause, restart, kill, copy and attach APIs + Mock and Live tests


docker/src/main/java/org/jclouds/docker/config/DockerOkHttpClientSupplier.java
```diff
@@ -0,0 +1,50 @@
...
+import javax.inject.Inject;
+import javax.inject.Singleton;
+
+@Singleton
+public class DockerOkHttpClientSupplier implements OkHttpClientSupplier {
+
+   private final SSLContextWithKeysSupplier sslContextWithKeysSupplier;
+
+   @Inject
+   private DockerOkHttpClientSupplier(SSLContextWithKeysSupplier sslContextWithKeysSupplier) {
```
nacx: Make it package private by removint the `private` modifier:

```java
@Inject
DockerOkHttpClientSupplier(SSLContextWithKeysSupplier sslContextWithKeysSupplier) {
```

docker/src/main/java/org/jclouds/docker/suppliers/SSLContextWithKeysSupplier.java
```diff
@@ -18,60 +18,154 @@
...
+   private static X509Certificate getCertificate(String certificate) {
+      try {
+         return (X509Certificate) CertificateFactory.getInstance("X.509").generateCertificate(
+                 new ByteArrayInputStream(certificate.getBytes(Charsets.UTF_8)));
+      } catch (CertificateException ex) {
+         throw new RuntimeException("Invalid certificate", ex);
+      }
+   }
+
+   private static PrivateKey getKey(String privateKey, String... password) {
```
nacx: Hmmm still an array. Does it really need to be an array?

nacx:
There just a couple remaining minors. lgtm once they're fixed. Thanks @andreaturli !

andreaturli:
@nacx I think I need to rebase and squash so that you can merge it, yes? 
Thanks!

andreaturli:
@nacx I think you are right, better to remove the password as we don't suggest to use encrypted PEM file

nacx:
The build fails due to checkstyle violations. Address them while squashing and +1!

andreaturli:
argh! @nacx 
Fixing them now

andreaturli:
pushed to master as 74cefc97d0d62a921bcb1635be45153037d4c233

PR #116, opened 2014-12-11
JCLOUDS-782: init for shipyard provider

cdancy:
Initial commit for shipyard provider. My intent was to have the entire provider coded and in something like "good working shape" before I submitted a PR but @nacx strongly suggested I not do this.

This code base is a fork of the work @andreaturli is doing with docker so there may be more than a few similarities. Shipyard itself is essentially a docker manager allowing users to connect N number of engines (docker daemons) to a single source allowing you to work with them from a single location. A few points to note:

-There are literally no testers at this point. Working on this...
-Around 10 files or so are json that gets handed back from shipyard. Used for testing purposes and do not need to be looked over.
-The idea here is to implement the entirety of ComputeServiceAdapter. All methods except 'createNodeWithGroupIntoName' have been coded. Working on this...
-All 'listNodes*' methods are being ignored. Not entirely sure why...

shipyard API can be found here: http://shipyard-project.com/docs/api/

So with that said please take a peek and let me know what you think!


shipyard/src/main/java/org/jclouds/shipyard/ShipyardApiMetadata.java
```diff
@@ -0,0 +1,84 @@
...
+      this(new Builder());
+   }
+
+   protected ShipyardApiMetadata(Builder builder) {
+      super(builder);
+   }
+
+   public static Properties defaultProperties() {
+      Properties properties = BaseHttpApiMetadata.defaultProperties();
+      properties.setProperty(ShipyardProperties.SHIPYARD_CREDENTIAL, "<remote-service-key-given-by-shipyard-cli>");
```
nacx: What is this property used for?

nacx:
Just a couple comments, and the code looks good. Once the tests mentioned in my previous comment are added, this will be good to go. Thanks @cdancy!

shipyard/src/main/java/org/jclouds/shipyard/ShipyardApiMetadata.java
cdancy: It's not being used just a leftover artifact from cloning the docker project. Removing now.

cdancy:
Very good. Been on vacation, not to mention a new baby, and just getting back into the swing of things. I'll look into getting those tests submitted now that this initial set of files is in good shape.

cdancy:
@nacx thoughts on latest commit? Want to get the Ok with this latest round so that I can then push live tests so we can close this PR and work on the next set of commits for fleshing out Shipyard API (in a subsequent PR of course).

nacx:
This is a good start with the Mock tests. But there are a couple things you have to address in all them:

* Add tests for failed responses (404, etc), for all methods where you explicitly define a `@Fallback`. Mock tests also have to verify that the fallback takes place when needed.
* In all tests that procude a body, you also need to verify that the generated body is the right one.
* In the `assertSent` method, add a check for the authentication header, to make sure the filter configured in each API is behaving properly too.

nacx:
rebuild please

nacx:
The squashed commit removes the .gitignore and the pom.xml file (causing the build failure). Can you fix that?

cdancy:
@nacx rebased against master. All is working well now ;)

shipyard/src/main/java/org/jclouds/shipyard/config/ShipyardProperties.java
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+package org.jclouds.shipyard.config;
+
+public class ShipyardProperties {
+
+   private ShipyardProperties(){};
+   
+   // shipyard remote api access key
+   public static final String SHIPYARD_CREDENTIAL = "shipyard.credential";
```
nacx: Remove this property (and the class) if unused.

shipyard/src/main/java/org/jclouds/shipyard/features/ImagesApi.java
```diff
@@ -0,0 +1,41 @@
...
+import org.jclouds.rest.annotations.Fallback;
+import org.jclouds.shipyard.domain.images.ImageInfo;
+
+@Consumes(MediaType.APPLICATION_JSON)
+@Path("/images/json")
+public interface ImagesApi {
+
+   @Named("images:list")
+   @GET
+   @Fallback(EmptyListOnNotFoundOr404.class)
```
nacx: Should this fallback be removed too, according to the other apis?

nacx:
Just two minors left. Nice job @cdancy! Can you fix them and squash again? Then I'll merge the PR.

shipyard/src/main/java/org/jclouds/shipyard/config/ShipyardProperties.java
cdancy: Doing so now

shipyard/src/main/java/org/jclouds/shipyard/features/ImagesApi.java
cdancy: yes it should ;)

nacx:
Merged to master as [02bc515d](https://git1-us-west.apache.org/repos/asf?p=jclouds-labs.git;a=commit;h=02bc515d). Thanks @cdancy!

PR #119, opened 2015-01-14
JCLOUDS-702: JCloud ProfitBricks provider - Server & Image API

devcsrj:
Hi! This PR is the implementation of ProfitBricks' API for:
* **Server** - [live test](https://gist.github.com/devcsrj/6a17a3c1347f740b0715)
* **Image** - [live test](https://gist.github.com/devcsrj/e4ab53864b75afa9655b)

As usual, looking forward to awesome code reviews!

*bump* 
Hi @nacx ! I'm back. Haha! :)
@adriancole


profitbricks/src/main/java/org/jclouds/profitbricks/binder/BaseProfitBricksRequestBinder.java
```diff
@@ -56,6 +56,8 @@ protected BaseProfitBricksRequestBinder(String paramName) {
    protected abstract String createPayload(T payload);
 
    protected String formatIfNotEmpty(String pattern, Object param) {
+      if (param == null )
+         return "";
       return Strings.isNullOrEmpty(param.toString()) ? "" : String.format(pattern, param);
```
nacx: [minor] You could do it in a more elegant way as:
```java
return Strings.emptyToNull(param) == null ? "" : String.format(pattern, param);
```

profitbricks/src/main/java/org/jclouds/profitbricks/domain/Image.java
```diff
@@ -0,0 +1,207 @@
...
+            return UNRECOGNIZED;
+         }
+      }
+   }
+
+   public abstract String id();
+
+   public abstract String name();
+
+   public abstract Float size(); // MB
```
nacx: Use the primitive type for non `@Nullable` variables.

profitbricks/src/main/java/org/jclouds/profitbricks/domain/Image.java
```diff
@@ -0,0 +1,207 @@
...
+
+   public abstract boolean isRamHotUnPlug();
+
+   public abstract boolean isNicHotPlug();
+
+   public abstract boolean isNicHotUnPlug();
+
+   public abstract boolean isDiscVirtioHotPlug();
+
+   public abstract boolean isDiscVirtioHotUnPlug();
```
nacx: Are all properties nullable?

profitbricks/src/main/java/org/jclouds/profitbricks/domain/Server.java
```diff
@@ -0,0 +1,424 @@
...
+   @Nullable
+   public abstract Date lastModificationTime();
+
+//   public abstract List<Storage> storages();
+//   public abstract List<Nic> storages();
+   public static Server create( String id, String name, int cores, int ram, boolean hasInternetAccess, ProvisioningState state,
+           Status status, OsType osType, AvailabilityZone availabilityZone, Date creationTime, Date lastModificationTime, Boolean isCpuHotPlug,
+           Boolean isRamHotPlug, Boolean isNicHotPlug, Boolean isNicHotUnPlug, Boolean isDiscVirtioHotPlug, Boolean isDiscVirtioHotUnPlug ) {
+      return new AutoValue_Server( isCpuHotPlug, isRamHotPlug, isNicHotPlug, isNicHotUnPlug, isDiscVirtioHotPlug, isDiscVirtioHotUnPlug,
+              cores, ram, id, name, hasInternetAccess, state, status, osType, availabilityZone, creationTime, lastModificationTime );
```
nacx: Take care with inheritance, as the order here is very important, as noted in the [google auto warnings](https://github.com/google/auto/tree/master/value#warnings). Make sure we have a test to cover the proper object construction so we don't break this accidentally by changing the implemented interface.

profitbricks/src/main/java/org/jclouds/profitbricks/domain/Server.java
```diff
@@ -0,0 +1,424 @@
...
+
+   @Nullable
+   public abstract String id();
+
+   @Nullable
+   @Override
+   public abstract String name();
+
+   @Nullable
+   public abstract boolean hasInternetAccess();
```
nacx: A primitive type can't be null.

profitbricks/src/main/java/org/jclouds/profitbricks/features/ServerApi.java
```diff
@@ -0,0 +1,164 @@
...
+@Consumes( MediaType.TEXT_XML )
+@Produces( MediaType.TEXT_XML )
+public interface ServerApi {
+
+   /**
+    * @return Returns information about all virtual server, such as configuration, provisioning status, power status, etc.
+    */
+   @POST
+   @Named( "server:getall" )
+   @Payload( "<ws:getAllDataCenters/>" )
```
nacx: Is this the right payload?

profitbricks/src/main/java/org/jclouds/profitbricks/features/ImageApi.java
```diff
@@ -0,0 +1,65 @@
...
+
+   /**
+    * @return Outputs a list of all HDD and/or CD-ROM/DVD images existing on or uploaded to the ProfitBricks FTP server.
+    */
+   @POST
+   @Named("image:getall")
+   @Payload("<ws:getAllImages/>")
+   @XMLResponseParser(ImageListResponseHandler.class)
+   @Fallback(Fallbacks.EmptyListOnNotFoundOr404.class)
+   List<Image> getAllImages();
```
nacx: Does the Profitbricks API return paginated results?

profitbricks/src/test/java/org/jclouds/profitbricks/features/ImageApiMockTest.java
```diff
@@ -0,0 +1,114 @@
...
+         pbApi.close();
+         server.shutdown();
+      }
+
+   }
+
+   @Test
+   public void testGetNonExistingImage() throws Exception {
+      MockWebServer server = mockWebServer();
+      server.enqueue(new MockResponse().setResponseCode(500).setBody(payloadFromResource("/image/image-not-found.xml")));
```
nacx: Hmmm this test depends on the `ResponseStatusFromPayloadHttpCommandExecutorService` fixing the response code. Even if the *real* provider returns a 500, I'd just configure a 404 response to make sure the API works, without coupling it to that class.
The `ResponseStatusFromPayloadHttpCommandExecutorService` tests should take care the status is properly fixed and that should be enough.

profitbricks/src/test/java/org/jclouds/profitbricks/http/parser/RequestIdOnlyResponseHandlerTest.java
```diff
@@ -0,0 +1,76 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jclouds.profitbricks.http.parser;
+
+import org.jclouds.http.functions.ParseSax;
+import static org.testng.Assert.assertEquals;
+import org.testng.annotations.Test;
+
+@Test( groups = "unit", testName = "RequestIdOnlyResponseHandler" )
```
nacx: [minor] `testName = "RequestIdOnlyResponseHandlerTest"`

nacx:
Thanks! Most comments are minors.
Given [this comment](https://github.com/jclouds/jclouds-labs/pull/119#discussion_r22944097), though, I realized that the mock tests are not verifying that the requests we generate have the expected body. This passed unnoticed in the previous review, but we should add those checks too.

profitbricks/src/main/java/org/jclouds/profitbricks/features/ServerApi.java
devcsrj: Woah nice catch. Will add tests to verify correct requests.

profitbricks/src/main/java/org/jclouds/profitbricks/domain/Image.java
devcsrj: So far, the [API](https://www.profitbricks.com/apidoc/1_3/APIDocumentation.html?GetAllImages.html) is always returning values for those properties and it's docs don't say that they're nullable. I think it's safe to assume that they're not.

profitbricks/src/main/java/org/jclouds/profitbricks/features/ImageApi.java
devcsrj: Nah it doesn't. Returns a <i>reaaally</i> large response containing all of the available images.

profitbricks/src/main/java/org/jclouds/profitbricks/binder/BaseProfitBricksRequestBinder.java
devcsrj: I can't exactly use it, since the param expects a String. (Doing <code>#toString()</code> won't do on null, cast to String throws a <code>ClassCastException</code> for <code>Boolean</code>) Anyway, see the change I did instead

devcsrj:
Hi @nacx . I addressed the comments and is now ready for another review cycle. :)

nacx:
Really nice test coverage @devcsrj. Thanks for the effort! Mind squashing the commits so I can merge it?

nacx:
Pushed to master as [3b6ddf0](https://git-wip-us.apache.org/repos/asf?p=jclouds-labs.git;a=commit;h=3b6ddf0d74bb7f4d5b7376b6e7f00b7336b139e8). Thanks @devcsrj!

PR #122, opened 2015-01-15
Split Auto dependencies

jdaggett:
Splitting the auto-service and auto-value dependencies allows us to control individual versions of these libraries.


jdaggett:
Pushed to **master** f2da7f2

PR #123, opened 2015-01-19
Add missing dependency to Profitbricks

nacx:
The build was broken since the last auto value commits were merged. This provider was missing.
/cc @jdaggett take care!


jdaggett:
@nacx It looks like I created my branch prior to commit 6e1a037. No wonder it failed! Thanks! :sunglasses:

PR #128, opened 2015-01-26
JCLOUDS-702: JClouds ProfitBricks provider - Storage API

devcsrj:
Following PRs [72](https://github.com/jclouds/jclouds-labs/pull/72) and [119](https://github.com/jclouds/jclouds-labs/pull/119), here is another PR that implements ProfitBricks' [Storage API](https://www.profitbricks.com/apidoc/1_3/APIDocumentation.html?StorageOperations.html).


profitbricks/src/main/java/org/jclouds/profitbricks/compute/internal/ProvisioningStatusPollingPredicate.java
```diff
@@ -50,6 +50,8 @@ public boolean apply(String input) {
             return expect == api.dataCenterApi().getDataCenterState(input);
          case SERVER:
             return expect == api.serverApi().getServer( input ).state();
+         case STORAGE:
+            return expect == api.storageApi().getStorage( input ).state();
```
nacx: Add the server and storage cases to the `ProvisioningStatusPollingPredicateTest`.

profitbricks/src/main/java/org/jclouds/profitbricks/domain/DataCenter.java
```diff
@@ -38,11 +40,13 @@
...
-//   @Nullable public abstract List<Storage> storages();
+   @Nullable public abstract List<Server> servers();
+   
+   @Nullable public abstract List<Storage> storages();
 //   @Nullable public abstract List<LoadBalancer> loadBalancers();
-   public static DataCenter create(String id, String name, int version, ProvisioningState state, Location location) {
-      return new AutoValue_DataCenter(id, name, version, state, location);
+   public static DataCenter create(String id, String name, int version, ProvisioningState state, Location location, List<Server> servers,
+           List<Storage> storages) {
+      return new AutoValue_DataCenter(id, name, version, state, location, servers, storages);
```
nacx: To enforce immutability it would be better to set the servers and storages with `ImmutableList.copyOf(servers)`.

profitbricks/src/main/java/org/jclouds/profitbricks/handlers/ProfitBricksHttpErrorHandler.java
```diff
@@ -59,7 +59,11 @@ public void handleError(final HttpCommand command, final HttpResponse response)
 	       break;
 	    case 413:
 	    case 503:
-	       exception = new InsufficientResourcesException(response.getMessage(), exception);
+               // if nothing (default message was OK) was parsed from command executor, assume it was an 503 (Maintenance) html response.
+               if (response.getMessage().equals( "OK" )) 
+                  exception = new RuntimeException("The ProfitBricks team is currently carrying out maintenance." );
```
nacx: Better propagate an `HttpResponseException` (with your custom message) so users have a better context to determine the cause of the failure.

profitbricks/src/main/java/org/jclouds/profitbricks/http/ResponseStatusFromPayloadHttpCommandExecutorService.java
```diff
@@ -119,4 +129,9 @@ private static boolean hasPayload(final HttpResponse response) {
       return response.getPayload() != null && response.getPayload().getRawContent() != null;
    }
 
+   private static boolean isSoapPayload( final InputStream is ) throws IOException {
+      String content = CharStreams.toString( new InputStreamReader( is, Charsets.UTF_8 ) );
+      is.reset();
+      return endSoapTag.matcher( content ).find();
+   }
```
nacx: Hmmm, so now we're buffering the input stream in memory twice: first to check if it is a SOAP payload, and then to actually parse the payload and see if it is an error. Could the same be done without this extra stream processing? Would the `faultHandler.parse(in);` fail, or can we use the parser (directly or indirectly) to cover that case too?

profitbricks/src/test/java/org/jclouds/profitbricks/features/StorageApiLiveTest.java
```diff
@@ -0,0 +1,167 @@
...
+   public void testUberStorage() {
+      api.storageApi().createStorage(
+              Storage.Request.creatingBuilder()
+              .dataCenterId( dataCenter )
+              .name( "Uber Storage" )
+              .size( 9999999f )
+              .build() );
+   }
+
+   @Test( dependsOnMethods = "testUberStorage" )
```
nacx: There is no need to depend on that test.

profitbricks/src/test/java/org/jclouds/profitbricks/features/StorageApiMockTest.java
```diff
@@ -0,0 +1,265 @@
...
+
+      try {
+         boolean result = api.deleteStorage( storageId );
+         assertRequestHasCommonProperties( server.takeRequest(), content );
+         assertTrue( result );
+      } finally {
+         pbApi.close();
+         server.shutdown();
+      }
+   }
```
nacx: Add a test to verify the configured delete fallback.

profitbricks/src/test/java/org/jclouds/profitbricks/http/ResponseStatusFromPayloadHttpCommandExecutorServiceTest.java
```diff
@@ -95,4 +100,43 @@ public void testUnauthorized() throws Exception {
...
+      } finally {
+         pbApi.close();
+         server.shutdown();
+      }
+   }
+   
+   @Test
+   public void testServiceUnderMaintenance() throws Exception {
+      MockWebServer server = mockWebServer();
+      for ( int i = 0; i <= MAX_RETRIES; i++ )  // jclouds retries 5 times
```
nacx: You can set the `Constants.PROPERTY_MAX_RETRIES` property when building the context to a lower value if you don't need all those 5 retries.

nacx:
Thanks @devcsrj!
Just one last minor (style) comment: prefer `if (foo) {` (wihtout spaces) instead of `if ( foo ) {`.

nacx:
BTW, can you share the live test results?

profitbricks/src/main/java/org/jclouds/profitbricks/http/ResponseStatusFromPayloadHttpCommandExecutorService.java
devcsrj: The reason behind this change is this: when I encountered a `503 Maintenance` from ProfitBricks, I noticed CommandExecutorService hung on the `faultHandler.parse(in);` itself. This was because (presumably) the HTML payload is actually a valid XML (it's [an XHTML](https://github.com/devcsrj/jclouds-labs/blob/master/profitbricks/src/test/resources/maintenance-503.html) even). It took forever, and doesn't timeout even if I override properties `Constants.PROPERTY_REQUEST_TIMEOUT` and `Constants.PROPERTY_CONNECTION_TIMEOUT` to `1000` when [testing this](https://github.com/devcsrj/jclouds-labs/blob/master/profitbricks/src/test/java/org/jclouds/profitbricks/http/ResponseStatusFromPayloadHttpCommandExecutorServiceTest.java#L125).

Further debugging the [ServiceFaultResponseHandler](https://github.com/devcsrj/jclouds-labs/blob/master/profitbricks/src/main/java/org/jclouds/profitbricks/http/parser/ServiceFaultResponseHandler.java), it shows that it hungs just before it even reaches the `<html>` tag - hangs on the `~<!DOCTYPE...` line.

Any thoughts?

devcsrj:
Here are the Storage API [live tests](https://gist.github.com/devcsrj/d0f3cb9d7c0f2e6a5988).

profitbricks/src/main/java/org/jclouds/profitbricks/http/ResponseStatusFromPayloadHttpCommandExecutorService.java
nacx: Understood.

The issue I see in your code is that you read the *original* InputStream into a String, and then your *reset* it, but that stream might not be replayable. That is the reason why we buffer it into a ByteArrayInputStream: we need to read it to see if there is an error, and if it is an OK response, we'll have to read it again to deserialize it to the right object. To do that, we *must* have a replayable inputstream and that's why we buffer it into a ByteArrayInputStream.

To make sure everything works, I'd change the code si the `isSoapPayload` method reads the already buffered stream. Makes sense?

devcsrj:
Hi @nacx , is this change better?

nacx:
>is this change better?

Yes, it is better, but take into account that if the `isSoapPayload` returns false you are not updating the payload in the response, and the original input stream might be already read. I'd put the soap check inside the try/finally block to make sure the response always has a readable payload when exiting the method.

devcsrj:
Oh this streams. Haha! Anyway, @nacx I amended the previous commit and [here's the change](https://github.com/devcsrj/jclouds-labs/commit/e16878763e51f249a428335f11279393ee5deabc#diff-7772dc9ed46441061f1e9d43596e1866R96).

nacx:
Pushed to master as [67cae0ca](https://git1-us-west.apache.org/repos/asf?p=jclouds-labs.git;a=commit;h=67cae0ca). Thanks @devcsrj!

PR #131, opened 2015-02-04
azurecompute: improvements to the domain objects

andreaturli:
@hsbhathiya could you please have a look?


nacx:
Just one comment: In the domain objects, google auto already guarantees immutability, but when there are lists, we have to make sure they are immutable too. Change all `create` methods that receive lists, and use `ImmutableList.copyOf` instead of assigning the lists directly to make sure they are immutable too.
+1 to merge once this is done. Thanks @andreaturli and @hsbhathiya!

andreaturli:
@nacx mind merging it?

nacx:
Changes lgtm. let's squash the commits and merge it yourself :) Thanks!

andreaturli:
Merged at https://git1-us-west.apache.org/repos/asf?p=jclouds-labs.git;a=blob;f=azurecompute/src/test/java/org/jclouds/azurecompute/xml/NetworkSecurityGroupHandlerTest.java;hb=a34e62ca

PR #132, opened 2015-02-09
In progress implementation of JClouds for Azure

eduardkoller:
Initial contribution for the Azure provider.
https://issues.apache.org/jira/browse/JCLOUDS-818


azure/compute/src/main/java/org/jclouds/azurecompute/AzureComputeApi.java
```diff
@@ -0,0 +1,67 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jclouds.azurecompute;
+
+import java.io.Closeable;
+import java.io.IOException;
+
+import org.jclouds.azurecompute.features.*;
```
nacx: [minor] Avoid wildcard imports.

azure/compute/src/main/java/org/jclouds/azurecompute/AzureComputeApi.java
```diff
@@ -0,0 +1,67 @@
...
+
+public class AzureComputeApi implements Closeable {
+    private final VirtualMachineApi virtualMachineApi;
+    private final ImageApi imageApi;
+    private final LocationApi locationApi;
+    private final HostedServiceApi hostedServiceApi;
+    private final RoleSizeApi roleSizeApi;
+
+    @Inject
+    public AzureComputeApi(VirtualMachineApi virtualMachineApi, ImageApi imageApi, LocationApi locationApi,
```
nacx: General: Following the Guice best practices, make all injection constructors package-private. This way only the injector (and the test classes) will be able to call it and you won't need to check for nulls, and other preconditions in the constructor code.

azure/compute/src/main/java/org/jclouds/azurecompute/AzureComputeApi.java
```diff
@@ -0,0 +1,67 @@
+/**
+ * Copyright (c)Microsoft Open Technologies, Inc.  All rights reserved.
```
nacx: All headers must be changed to the ASF license header.

azure/compute/src/main/java/org/jclouds/azurecompute/compute/AzureComputeService.java
```diff
@@ -0,0 +1,100 @@
...
+   }
+
+   /**
+    * Returns template options, except of type {@link org.jclouds.azurecompute.compute.options.AzureTemplateOptions}.
+    */
+   @Override
+   public AzureTemplateOptions templateOptions() {
+      return AzureTemplateOptions.class.cast(super.templateOptions());
+   }
+}
```
nacx: There is no need to subclass the BaseComputeService just to override this. No one will directly use this concrete implementation, and the injector will already provide the right instance of the options class, as a binding for it should be configured in the ComputeserviceContextModule.
This class could be removed.

azure/compute/src/main/java/org/jclouds/azurecompute/compute/config/ConfigurationSupplier.java
```diff
@@ -0,0 +1,48 @@
...
+public class ConfigurationSupplier implements Supplier<Configuration> {
+    private final Configuration configuration;
+
+    @Inject
+    public ConfigurationSupplier(@Provider Supplier<Credentials> creds) {
+        Configuration configuration ;
+        try {
+            configuration = PublishSettingsLoader.createManagementConfiguration(creds.get().credential, creds.get().identity);
+        } catch (IOException e) {
+            e.printStackTrace();
```
nacx: Use a logger if there is something to log, but remove this.

azure/compute/src/main/java/org/jclouds/azurecompute/compute/config/ConfigurationSupplier.java
```diff
@@ -0,0 +1,48 @@
...
+
+@Singleton
+public class ConfigurationSupplier implements Supplier<Configuration> {
+    private final Configuration configuration;
+
+    @Inject
+    public ConfigurationSupplier(@Provider Supplier<Credentials> creds) {
+        Configuration configuration ;
+        try {
+            configuration = PublishSettingsLoader.createManagementConfiguration(creds.get().credential, creds.get().identity);
```
nacx: Suppliers are used to defer the work until it is actually needed, so this logic should be moved to the `get` method. If you need to cache the results of the call to get, consider decorating the supplier with `Suppliers.memoize` or similar when configuring it in the corresponding module.

azure/compute/src/main/java/org/jclouds/azurecompute/compute/functions/DeploymentResponseToVirtualMachine.java
```diff
@@ -0,0 +1,53 @@
...
+
+import javax.annotation.Resource;
+import javax.inject.Inject;
+import javax.inject.Singleton;
+
+@Singleton
+public class DeploymentResponseToVirtualMachine implements Function<DeploymentGetResponse, VirtualMachine> {
+
+    @Resource
+    protected Logger logger = Logger.NULL;
```
nacx: Remove if unused.

azure/compute/src/main/java/org/jclouds/azurecompute/compute/functions/OSImageListResponseToOSImage.java
```diff
@@ -0,0 +1,70 @@
...
+import javax.annotation.Resource;
+import javax.inject.Inject;
+import javax.inject.Singleton;
+
+
+@Singleton
+public class OSImageListResponseToOSImage implements Function<VirtualMachineOSImageListResponse.VirtualMachineOSImage, OSImage> {
+
+    @Resource
+    protected Logger logger = Logger.NULL;
```
nacx: Same here. remove if unused.

azure/compute/src/main/java/org/jclouds/azurecompute/compute/functions/OSImageResponseToOSImage.java
```diff
@@ -0,0 +1,72 @@
...
+
+import javax.annotation.Resource;
+import javax.inject.Inject;
+import javax.inject.Singleton;
+
+@Singleton
+public class OSImageResponseToOSImage implements Function<VirtualMachineOSImageGetResponse, OSImage> {
+
+    @Resource
+    protected Logger logger = Logger.NULL;
```
nacx: Remove this.

azure/compute/src/main/java/org/jclouds/azurecompute/compute/functions/OSImageToImage.java
```diff
@@ -0,0 +1,88 @@
...
+//                (from.getDescription() != null && from.getDescription().contains("64")) ||
+//                (from.getLabel() != null && from.getLabel().contains("64")) || osFamily == OsFamily.WINDOWS;
+
+        return new OperatingSystem(osFamily, osName, osVersion, osArch, osDescription, is64Bit);
+    }
+
+    @Override
+    public Image apply(OSImage from) {
+        ImageBuilder builder = new ImageBuilder();
+        builder.ids(from.getName());
```
nacx: This sets two fields: the `id` and the `providerId`. The former is the ID jclouds will use to identify uniquely the image across all regions, and the latter the *real* ID of the resource in the provider. Is this true in this case?
Usually jclouds uses a different ID and prepends the region to the provider ID. Take this into accound to make sure the providerId and the id (which is the one used when calling getImage(id) in the computeservice) are configured properly.

azure/compute/src/main/java/org/jclouds/azurecompute/compute/functions/VirtualMachineToNodeMetadata.java
```diff
@@ -0,0 +1,109 @@
...
+      this.hardwares = checkNotNull(hardwares, "hardwares");
+      this.locations = checkNotNull(locations, "locations");
+      this.deploymentStatusToStatus = checkNotNull(deploymentStatusToStatus);
+      this.instanceStatusToStatus = instanceStatusToStatus;
+   }
+
+   @Override
+   public NodeMetadata apply(VirtualMachine from) {
+      NodeMetadataBuilder builder = new NodeMetadataBuilder();
+      builder.ids(from.getServiceName() + ":" + from.getRoleName());
```
nacx: Same comment about the image IDs. You might want to set separately the `id` and the `providerId`.

azure/compute/src/main/java/org/jclouds/azurecompute/compute/functions/VirtualMachineToNodeMetadata.java
```diff
@@ -0,0 +1,109 @@
...
+      builder.name(from.getServiceName() + ":" + from.getRoleName());
+//      Location location = Iterables.find(locations.get(), LocationPredicates.idEquals(from.getDatacenter().getId() + ""));
+//      builder.location(location);
```
nacx: The location must be set in the node.

azure/compute/src/main/java/org/jclouds/azurecompute/compute/functions/VirtualMachineToNodeMetadata.java
```diff
@@ -0,0 +1,109 @@
...
+         Image image = from(images.get()).firstMatch(ImagePredicates.idEquals(imageId)).orNull();
+         if (image != null) {
+             builder.operatingSystem(image.getOperatingSystem());
+         }
+      }
+      builder.status(deploymentStatusToStatus.get(from.getDeploymentStatus()));
+
+//      builder.status(serverStateToNodeStatus.get(from.getState()));
+      builder.publicAddresses(ImmutableList.copyOf(Collections2.transform(from.getPublicIpAddresses(), toIpAddress)));
+      return builder.build();
```
nacx: The node should also try to populate the credentials. If the API does not provide them, it is usually a good idea to read the credentials from the jclouds credential store as done in [DigitalOcean](https://github.com/jclouds/jclouds-labs/blob/master/digitalocean/src/main/java/org/jclouds/digitalocean/compute/functions/DropletToNodeMetadata.java#L115-L120).

azure/compute/src/main/java/org/jclouds/azurecompute/compute/internal/AzureTemplateBuilderImpl.java
```diff
@@ -0,0 +1,140 @@
...
+            @Override
+            public Image apply(Iterable<? extends Image> input) {
+                // filter private images
+                Predicate<Image> statusPredicate = new Predicate<Image>() {
+
+                    @Override
+                    public boolean apply(Image image) {
+                        return image.getStatus() == Image.Status.AVAILABLE;
+                    }
+                };
```
nacx: Is this predicate needed? According to [your image transformation function](https://github.com/MSOpenTech/jclouds-labs/blob/azure/azure/compute/src/main/java/org/jclouds/azurecompute/compute/functions/OSImageToImage.java#L83) images are always available. If that's not true, then that function should be improved to populate the right status.

azure/compute/src/main/java/org/jclouds/azurecompute/compute/functions/OSImageToImage.java
```diff
@@ -0,0 +1,88 @@
...
+        builder.name(from.getLabel());
+        builder.description(from.getDescription() == null ? "" : from.getDescription());
+        builder.operatingSystem(parseOs(from));
+        builder.uri(from.getMediaLink());
+//        builder.status(toPortableImageStatus.get(from.getState()));
+        builder.status(Status.AVAILABLE);
+        builder.status(PUBLIC_CATEGORY.equals(from.getCategory()) ? Status.AVAILABLE : Status.PENDING);
+        builder.userMetadata(ImmutableMap.<String, String> of(OSImage.LOCATION, from.getLocation()));
```
nacx: Can this be transformed into a proper location and set the location field properly?

azure/compute/src/main/java/org/jclouds/azurecompute/compute/strategy/AzureComputeServiceAdapter.java
```diff
@@ -0,0 +1,326 @@
...
+/**
+ * defines the connection between the {@link org.jclouds.azurecompute.AzureComputeApi} implementation and the jclouds
+ * {@link org.jclouds.compute.ComputeService}
+ */
+@Singleton
+public class AzureComputeServiceAdapter implements ComputeServiceAdapter<VirtualMachine, RoleSizeListResponse.RoleSize, OSImage,
+        LocationsListResponse.Location> {
+
+    @Resource
+    protected Logger logger = Logger.NULL;
```
nacx: This should be qualified as the compute logger.

azure/compute/src/main/java/org/jclouds/azurecompute/compute/strategy/AzureComputeServiceAdapter.java
```diff
@@ -0,0 +1,326 @@
...
+//            timeouts.nodeRunning * 9l / 10l);
+//      this.serverLatestJobCompletedShort = retry(new ServerLatestJobCompleted(client.getJobServices()),
+//            timeouts.nodeRunning * 1l / 10l);
+    }
+
+
+    @Override
+    public NodeAndInitialCredentials<VirtualMachine> createNodeWithGroupEncodedIntoName(String group, String name,
+                                                                                    Template template) {
+        logger.info("createNodeWithGroupEncodedIntoName");
```
nacx: Move this kind of logs to trace level.

azure/compute/src/main/java/org/jclouds/azurecompute/compute/strategy/AzureComputeServiceAdapter.java
```diff
@@ -0,0 +1,326 @@
...
+        checkNotNull(template, "template was null");
+        checkNotNull(template.getOptions(), "template options was null");
+        checkArgument(template.getOptions().getClass().isAssignableFrom(AzureTemplateOptions.class),
+                "options class %s should have been assignable from AzureTemplateOptions", template.getOptions().getClass());
```
nacx: These three precondition checks are redundant. Will always be OK, as the injector and the compute service module configuration will enforce this.

azure/compute/src/main/java/org/jclouds/azurecompute/compute/strategy/AzureComputeServiceAdapter.java
```diff
@@ -0,0 +1,326 @@
...
+        AzureTemplateOptions templateOptions = template.getOptions().as(AzureTemplateOptions.class);
+        checkNotNull(templateOptions.getStorageAccount(), "storage account should be specified in order to create node");
```
nacx: This mandatory field needs to be removed. Users must be able to spawn new nodes by just doing:
```java
compute.templageBuilder().build();
```
Consider configuring the storage account as part of the credentials

azure/compute/src/main/java/org/jclouds/azurecompute/compute/strategy/AzureComputeServiceAdapter.java
```diff
@@ -0,0 +1,326 @@
...
+        ArrayList<ConfigurationSet> configlist = new ArrayList<ConfigurationSet>();
+        ConfigurationSet configset = new ConfigurationSet();
+        configset.setConfigurationSetType(template.getImage().getOperatingSystem().getFamily() == OsFamily.LINUX ?
+                ConfigurationSetTypes.LINUXPROVISIONINGCONFIGURATION : ConfigurationSetTypes.WINDOWSPROVISIONINGCONFIGURATION);
+//        configset.setConfigurationSetType(ConfigurationSetTypes.WINDOWSPROVISIONINGCONFIGURATION);
+        String vmName = name.substring(group.length() + 1);
+//        required
+        configset.setComputerName(vmName);
+        String adminuserPassword = "Pamir5463";
+        String adminUserName = "farlen";
```
nacx: This can't be hardcoded here and must be read from the template options, as it provides fields to override the login username, password etc. Default credentials can be set in the image transformation function or as default properties in the api metadata, where you can specify credentials for each os family.

azure/compute/src/main/java/org/jclouds/azurecompute/compute/strategy/AzureComputeServiceAdapter.java
```diff
@@ -0,0 +1,326 @@
...
+            parameters.setOSVirtualHardDisk(oSVirtualHardDisk);
+            parameters.setProvisionGuestAgent(true);
+            parameters.setConfigurationSets(configlist);
+
+            client.getVirtualMachineApi().addRole(group, hostedService.getDeployments().get(0).getName(), parameters);
+        }
+
+        VirtualMachine virtualMachine = getNode(name);
+        virtualMachine.setServiceName(group);
+        return new NodeAndInitialCredentials<VirtualMachine>(virtualMachine, name, null);
```
nacx: The node credentials must be returned, as they will be used by the compute service to bootstrap the node, run scripts, etc.

azure/compute/src/main/java/org/jclouds/azurecompute/compute/strategy/AzureComputeServiceAdapter.java
```diff
@@ -0,0 +1,326 @@
...
+                    virtualMachines.add(virtualMachine);
+                }
+            }
+//            addDeploymentIfExists(deployments, hostedService.getServiceName(), DeploymentSlot.Staging);
+//            addDeploymentIfExists(deployments, hostedService.getServiceName(), DeploymentSlot.Production);
+        }
+        return virtualMachines;
+    }
+
+    private void addDeploymentIfExists(List<VirtualMachine> virtualMachines, String serviceName, DeploymentSlot deploymentSlot) {
```
nacx: Remove if unused.

azure/compute/src/main/java/org/jclouds/azurecompute/compute/strategy/AzureComputeServiceAdapter.java
```diff
@@ -0,0 +1,326 @@
...
+        }
+    }
+
+    @Override
+    public Iterable<VirtualMachine> listNodesByIds(final Iterable<String> ids) {
+        Set<Long> idsAsLongs = FluentIterable.from(ids)
+                .transform(toLong())
+                .toSet();
+        logger.info("listNodesByIds()");
+        throw new UnsupportedOperationException("Should not be here");
```
nacx: This must be implemented. If there is not an API call to get N nodes by id, then just call the existing `listNodes()` method and filter afterwards.

azure/compute/src/main/java/org/jclouds/azurecompute/compute/strategy/AzureComputeServiceAdapter.java
```diff
@@ -0,0 +1,326 @@
...
+    @Override
+    public void suspendNode(String id) {
+        String service = groupNamingConvention.createWithoutPrefix().extractGroup(id);
+        String name = service == null ? id : id.substring(service.length() + 1);
+        String deploymentName = client.getVirtualMachineApi().getDeploymentBySlot(service, DeploymentSlot.Production).getName();
+        VirtualMachineShutdownParameters parameters = new VirtualMachineShutdownParameters();
+        client.getVirtualMachineApi().shutdownVirtualMachine(service, deploymentName, name, parameters);
+    }
+
+    private Function<String, Long> toLong() {
```
nacx: Remove if unused.

azure/compute/src/main/java/org/jclouds/azurecompute/compute/suppliers/AzureHardwareSupplier.java
```diff
@@ -0,0 +1,71 @@
...
+            *//*.volumes(ImmutableList.<Volume> of(new VolumeImpl(25.0f, true, true)))*//*.build();
+
+   public static final ImmutableSet<Hardware> H_ALL = ImmutableSet.of(ExtraSmall, Small, Medium, Large, ExtraLarge, A5, A6, A7);
+
+   @Override
+   public Set<? extends Hardware> get() {
+      return H_ALL;
+   }
+}
+*/
```
nacx: Remove this file if not needed.

azure/compute/src/main/java/org/jclouds/azurecompute/compute/suppliers/LocationSupplier.java
```diff
@@ -0,0 +1,63 @@
...
+        this.adapter = adapter;
+        this.transformer = transformer;
+    }
+
+    @Override
+    public Set<? extends Location> get() {
+        return FluentIterable.from(adapter.listLocations()).filter(notNull()).transform(transformer).toSet();
+//            return transformGuardingNull(adapter.listLocations(), transformer);
+    }
+}
```
nacx: There is no need to provide a custom location supplier, as the adapter already has a method that lists them. Just add this binding to the `AzoreComputeServiceContextModule` configure method:
```java
install(new LocationsFromComputeServiceAdapterModule<VirtualMachine, RoleSizeListResponse.RoleSize, OSImage, LocationsListResponse.Location>() {
});
```
And then you can remove this class. jclouds will extract the locations by calling the compute service implementation and the corresponding transformation function.

azure/compute/src/main/java/org/jclouds/azurecompute/config/AzureComputeApiModule.java
```diff
@@ -0,0 +1,100 @@
...
+//      super.configure();
+       bind(new TypeLiteral<ApiContext<AzureComputeApi>>() {
+       }).to(new TypeLiteral<ApiContextImpl<AzureComputeApi>>() {
+       });
+       installLocations();
+   }
+
+   protected void installLocations() {
+      install(new LocationModule());
+      bind(ImplicitLocationSupplier.class).to(AzureDefaultLocationSupplier.class).in(Scopes.SINGLETON);
```
nacx: All this can be removed. jclouds will provide a default location (the first region) by default.

azure/compute/src/main/java/org/jclouds/azurecompute/config/AzureComputeApiModule.java
```diff
@@ -0,0 +1,100 @@
...
+import org.jclouds.rest.config.HttpApiModule;
+
+import com.google.common.base.Supplier;
+import com.google.common.base.Suppliers;
+import com.google.inject.Provides;
+import com.google.inject.Scopes;
+import org.jclouds.rest.internal.ApiContextImpl;
+
+
+@ConfiguresRestClient
```
nacx: Change to `@ConfiguresHttpApi`. The rest client is now deprecated.

azure/compute/src/main/java/org/jclouds/azurecompute/domain/OSImage.java
```diff
@@ -0,0 +1,431 @@
...
+import java.util.Calendar;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * OS images from the image repository
+ *
+ * @see <a href="http://msdn.microsoft.com/en-us/library/jj157191" >api</a>
+ */
+public class OSImage {
```
nacx: We use Google Auto for the domain model. This allows us to:

* Enforce immutable objects easily.
* Remove the boilerplate code introduced in large domain objects.
* Avoid adding (and maintaining) the equals/hashcode/tostring methods.

In general, refactor all these domain objects to use Google auto. You can have a look at the latest OpenStack pull requests, ProfitBricks or GCE providers for further reference.

azure/compute/src/main/java/org/jclouds/azurecompute/domain/OSImage.java
```diff
@@ -0,0 +1,431 @@
...
+
+    public static Builder<?> builder() {
+        return new ConcreteBuilder();
+    }
+
+    public Builder<?> toBuilder() {
+        return new ConcreteBuilder().fromHostedService(this);
+    }
+
+    public abstract static class Builder<T extends Builder<T>> {
```
nacx: If the builder is not going to be subclassed, there is no need for an abstract and a concrete builder. Keep just one regular builder and remove the unnecessary inheritance.

azure/compute/src/main/resources/META-INF/services/org.jclouds.providers.ProviderMetadata
```diff
@@ -0,0 +1 @@
+org.jclouds.azurecompute.AzureComputeProviderMetadata
```
nacx: We use Google Auto for service loader discovery. Remove this file and annotate the `AzureComputeProviderMetadata` accordingly.

nacx:
Thankd @eduardkoller!

I've done an initial review focusing on structural things that must be addressed before this can be merged. Before we can merge it, all the comments must be addressed. I've added some general ones that I've mentioned only once to avoid repeating them:

* All license headers must be the ASF headers.
* Remove all `printStackTrace` references or replace them by proper log messages.
* Remove all wildcard imports.
* Remove all unused loggers, and properly qualify the remaining ones. Also move all debugging messages such as "entering here" to trace level.
* Reduce the constructor visibility of all injection constructors to package-private.
* Refactor the domain model classes to use Google Auto.

Please, when committing fixes don't squash the commits. This way we'll be able to review just the last changes without having to go through the entire PR again.

Once all global and particular comments have been addressed, and the next reviews are fine, I'll merge this. Once that happens, though, there will be several things that must be done:

**Tests**:

There is no single unit test in this code, and we really need them. The following tests must be added:

* Unit tests for each transformation function.
* Live tests for each *Api class method.
* A live test that subclasses the `BaseComputeServiceLiveTest`. This is our contract, so it is extremely important.
* A live test that subclasses the `BaseTemplateBuilderLiveTest`. Same as above, we need this test passing.

Once all tests are in place, we can consider the provider as a "working provider".

Regarding promotion to core, as [discussed in the mailing list](http://markmail.org/message/et6pmzrygbnjvfuw), there are more structural changes to be made, such as use the jclouds HTTP layer. It is not an issue now in labs, but I just want to leave the note now that there are two different implementations of the provider growing.

eduardkoller:
Thank you for the thorough review, Ignasi – much appreciated.

We’ll address the issues you flagged and provide an updated PR soon.

 

Thanks,

Eduard

nacx:
A couple more things to consider that have come to my mind after reading the ComputeService implementation again:

* The `templateOptions.inboundPorts` should be considered to open those ports in the target virtual machine. Is this something that can be done per VM, or the "external" ports must be unique per Cloud Service/ whatever? (I'm still not familiar enough with some of the Azure concepts).
* The template options also provides a way so users can specify a set of network IDs, and jclouds is supposed to assign to the node one IP address in each of those networks. Is this something that makes sense in Azure too?
* Does azure support setting tags and metadata (key/value pairs) in each VM? If so, the compute service adapter should also take into account the tags and userMetadata template options.

ilgrosso:
Hi @eduardkoller,
shall this issue be closed as [this e-mail](http://www.mail-archive.com/dev@jclouds.apache.org/msg05877.html) suggests?

eduardkoller:
Closing, per [this email] (http://www.mail-archive.com/dev%40jclouds.apache.org/msg05877.html)

PR #135, opened 2015-02-13
AzureCompute: initial work to support ComputeServiceAdapter

andreaturli:
improve DeploymentToNodeMetadata
add support for get/set NetworkConfiguration
added AzureComputeTemplateOptions to manage networkConfigurations
add support for create/checkAvailable/list storageAccounts
add support for inboundPorts
add support for SecurityGroupExtension by using NetworkSecurityGroup
fix destroy node


azurecompute/src/main/java/org/jclouds/azurecompute/AzureComputeProviderMetadata.java
```diff
@@ -48,6 +49,7 @@ public AzureComputeProviderMetadata(Builder builder) {
 
    public static Properties defaultProperties() {
       Properties properties = new Properties();
+      properties.setProperty(TEMPLATE, "osFamily=UBUNTU,loginUser=jclouds");
```
nacx: Don't duplicate the property here. It is enough to have it in the ApiMetadata.

azurecompute/src/main/java/org/jclouds/azurecompute/AzureManagementApiMetadata.java
```diff
@@ -57,16 +63,17 @@ public static Properties defaultProperties() {
       protected Builder() {
          id("azurecompute")
          .name("Microsoft Azure Service Management Service API")
-         .version("2014-06-01")
+         .version("2014-10-01")
          .identityName("Path to Management Certificate .p12 file, or PEM string")
          .credentialName("Password to Management Certificate")
          .defaultEndpoint("https://management.core.windows.net/${" + SUBSCRIPTION_ID + "}")
```
nacx: Endpoints are meant to be usable *without* user modification. If there is no endpoint that can be hardcoded, just remove it.

azurecompute/src/main/java/org/jclouds/azurecompute/compute/AzureComputeServiceAdapter.java
```diff
@@ -16,39 +16,275 @@
...
+   @Named(ComputeServiceConstants.COMPUTE_LOGGER)
+   protected Logger logger = Logger.NULL;
+
+   private final AzureComputeApi api;
+   private final String subscriptionId;
+   private final long operationTimeout;
+   private Predicate<String> operationSucceeded;
+
+   @Inject
+   public AzureComputeServiceAdapter(final AzureComputeApi api, @Named(SUBSCRIPTION_ID) String subscriptionId, @Named(OPERATION_TIMEOUT) long operationTimeout) {
```
nacx: Make the constructor package private and remove the null checks.

azurecompute/src/main/java/org/jclouds/azurecompute/compute/AzureComputeServiceAdapter.java
```diff
@@ -16,39 +16,275 @@
...
+      this.api = checkNotNull(api, "api");
+      this.subscriptionId = checkNotNull(subscriptionId, "subscriptionId");
+      this.operationTimeout = checkNotNull(operationTimeout, "operationTimeout");
+      this.operationSucceeded = retry(new Predicate<String>() {
+         public boolean apply(String input) {
+            return api.getOperationApi().get(input).status() == Operation.Status.SUCCEEDED;
+         }
+      }, operationTimeout, 5, 5, SECONDS);
```
nacx: Consider moving this predicate to a provider method in the Guice configuration module so it can be injected and reused (reused in live tests too?).

azurecompute/src/main/java/org/jclouds/azurecompute/compute/AzureComputeServiceAdapter.java
```diff
@@ -16,39 +16,275 @@
...
+
+      final String loginUser = templateOptions.getLoginUser() != null ? templateOptions.getLoginUser() : DEFAULT_LOGIN_USER;
+      final String loginPassword = templateOptions.getLoginPassword() != null ? templateOptions.getLoginPassword() : DEFAULT_LOGIN_PASSWORD;
+      final String location = template.getLocation().getId();
+      final int[] inboundPorts = template.getOptions().getInboundPorts();
+
+      String storageAccountName = templateOptions.getStorageAccountName().or(generateStorageServiceName(DEFAULT_STORAGE_ACCOUNT_PREFIX));
+
+      // get or create storage service
+      StorageService storageService = tryFindExistingStorageServiceAccountOrCreate(location, storageAccountName, storageAccountType);
```
nacx: Take care with `get or create` operations here, as they will introduce race conditions. This method will be executed concurrently when creating more than one node.
This should be done in the upper layer, in a custom implementation of the `CreateNodesWithGroupEncodedIntoNameThenAddToSet` class. Take a look at how [digitalocean creates the key pairs globally](https://github.com/jclouds/jclouds-labs/blob/master/digitalocean/src/main/java/org/jclouds/digitalocean/compute/strategy/CreateKeyPairsThenCreateNodes.java) before creating the nodes to avoid this kind of race conditions.

azurecompute/src/main/java/org/jclouds/azurecompute/compute/AzureComputeServiceAdapter.java
```diff
@@ -16,39 +16,275 @@
...
+      String storageServiceAccountName = storageService.serviceName();
+
+      // check existence or create virtual network
+      checkExistingVirtualNetworkNamedOrCreate(virtualNetworkName, location, subnetName, addressSpaceAddressPrefix, subnetAddressPrefix);
```
nacx: Same comment here. as the templateOptions are shared with all nodes, all "get or create" operations should be done *before* this method is run to avoid race conditions.

nacx: Also, there is the `tempalteOptions.getnetworks()` method that allows users to specify a set of network IDs where the node should be attached. Is this something that can be done in Azure? No need to address this in this PR.

azurecompute/src/main/java/org/jclouds/azurecompute/compute/AzureComputeServiceAdapter.java
```diff
@@ -16,39 +16,275 @@
...
+         logger.debug("Adding a networkSecurityGroup %s is already applied to subnet '%s' of virtual network %s ...", networkSecurityGroupName, subnetName, virtualNetworkName);
+         String addToSubnetId = api.getNetworkSecurityGroupApi().addToSubnet(virtualNetworkName, subnetName,
+                 NetworkSecurityGroup.create(networkSecurityGroupName, null, null, null));
+         String operationDescription = format("Add networkSecurityGroup(%s) to subnet(%s)", networkSecurityGroupName, subnetName);
+         waitForOperationCompletion(addToSubnetId, operationDescription);
+      }
+
+      logger.debug("Creating a cloud service with name '%s', label '%s' in location '%s'", name, name, location);
+      String createCloudServiceRequestId = api.getCloudServiceApi().createWithLabelInLocation(name, name, location);
+      waitForOperationCompletion(createCloudServiceRequestId, format("Create cloud service (%s)", name));
```
nacx: Same here. Move all operations global to the nodes to the upper layer. Keep in mind that you might be deploying N nodes and this method will be executed concurrently for all them.

azurecompute/src/main/java/org/jclouds/azurecompute/compute/AzureComputeServiceAdapter.java
```diff
@@ -16,39 +16,275 @@
...
+      }, 30 * 60, 1, SECONDS).apply(name)) {
+         logger.warn("Instances %s of %s has not reached the status %s within %sms so it will be destroyed.",
+                 Iterables.toString(api.getDeploymentApiForService(name).get(name).roleInstanceList()), name,
+                 READY_ROLE, Long.toString(operationTimeout));
+         api.getDeploymentApiForService(group).delete(name);
+         api.getCloudServiceApi().delete(name);
+         throw new IllegalStateException(format("Deployment %s is being destroyed as its instanceStatus didn't reach " +
+                 "status %s after %ss. Please, try by increasing `jclouds.azure.operation-timeout` and " +
+                 " try again", name, READY_ROLE, 30 * 60));
+      }
```
nacx: The jclouds ComputeService will already perform the active wait until the node is in state RUNNING. Is it really necessary to manually wait for the node being ready here? can't you just return the Deployment without waiting?

azurecompute/src/main/java/org/jclouds/azurecompute/compute/AzureComputeServiceAdapter.java
```diff
@@ -58,21 +294,36 @@ public OSImage getImage(String id) {
...
+
+      // TODO detach disk before deleting node
 
+      String deleteDeploymentRequestId = api.getDeploymentApiForService(cloudService.name()).delete(id);
+      String operationDescription = format("Delete deployment(%s) of cloudService(%s)", id, cloudService.name());
+      waitForOperationCompletion(deleteDeploymentRequestId, operationDescription);
+
+      String cloudServiceDeleteRequestId = api.getCloudServiceApi().delete(cloudService.name());
+      operationDescription = format("Delete cloudService(%s)", cloudService.name(), Long.toString(operationTimeout));
+      waitForOperationCompletion(cloudServiceDeleteRequestId, operationDescription);
```
nacx: When destroying nodes, should we consider removing the storage account at some point?

azurecompute/src/main/java/org/jclouds/azurecompute/compute/AzureComputeServiceAdapter.java
```diff
@@ -88,17 +339,70 @@ public void resumeNode(String id) {
    @Override
    public void suspendNode(String id) {
       // TODO Auto-generated method stub
-
```
nacx: Is this intentionally unimplemented? Should it be better to throw an exception if this is not supported?

azurecompute/src/main/java/org/jclouds/azurecompute/compute/AzureComputeServiceAdapter.java
```diff
@@ -88,17 +339,70 @@ public void resumeNode(String id) {
...
+                 }
+              })
+              .filter(notNull())
+              .toSet();
+      return deployments;
    }
 
    @Override public Iterable<Deployment> listNodesByIds(Iterable<String> ids) {
       // TODO Auto-generated method stub
       return null;
```
nacx: If there is no method in the provider API to get N nodes by their ID, just call the `listNodes()` and filter the results.

azurecompute/src/main/java/org/jclouds/azurecompute/compute/AzureComputeServiceAdapter.java
```diff
@@ -88,17 +339,70 @@ public void resumeNode(String id) {
...
+      String characters = "abcdefghijklmnopqrstuvwxyz";
+      StringBuilder builder = new StringBuilder();
+      builder.append(prefix);
+      int charactersLength = characters.length();
+      for (int i = 0; i < 10; i++) {
+         double index = Math.random() * charactersLength;
+         builder.append(characters.charAt((int) index));
+      }
+      return builder.toString();
+   }
```
nacx: Move this to a Supplier that can be injected wherever it is needed. This way you have a clean way of generating storage service names. Consider using the `GroupNamingConvention` to generate it properly prefixed, if needed.

azurecompute/src/main/java/org/jclouds/azurecompute/compute/AzureComputeServiceAdapter.java
```diff
@@ -88,17 +339,70 @@ public void resumeNode(String id) {
...
+
+   @VisibleForTesting
+   public static URI createMediaLink(String storageServiceName, String diskName) {
+      return URI.create(String.format("https://%s.blob.core.windows.net/vhds/disk-%s.vhd", storageServiceName, diskName));
+   }
+
+   public void waitForOperationCompletion(String operationId, String operationDescription) {
```
nacx: Does this need to be public?

azurecompute/src/main/java/org/jclouds/azurecompute/compute/extensions/AzureComputeSecurityGroupExtension.java
```diff
@@ -0,0 +1,460 @@
...
+ * An extension to compute service to allow for the manipulation of {@link org.jclouds.compute.domain.SecurityGroup}s. Implementation
+ * is optional by providers.
+ *
+ * It considers only the custom rules added by the user and ignores the default rules created by Azure
+ */
+public class AzureComputeSecurityGroupExtension implements SecurityGroupExtension {
+
+   protected final AzureComputeApi api;
+   private final long operationTimeout;
+   private Predicate<String> operationSucceeded;
```
nacx: Make this final too.

azurecompute/src/main/java/org/jclouds/azurecompute/compute/AzureComputeServiceAdapter.java
```diff
@@ -16,39 +16,275 @@
...
+   private static final String DEFAULT_STORAGE_ACCOUNT_PREFIX = "jclouds";
+
+   @Resource
+   @Named(ComputeServiceConstants.COMPUTE_LOGGER)
+   protected Logger logger = Logger.NULL;
+
+   private final AzureComputeApi api;
+   private final String subscriptionId;
+   private final long operationTimeout;
+   private Predicate<String> operationSucceeded;
```
nacx: Make it final.

azurecompute/src/main/java/org/jclouds/azurecompute/compute/extensions/AzureComputeSecurityGroupExtension.java
```diff
@@ -0,0 +1,460 @@
...
+
+   @Resource
+   @Named(ComputeServiceConstants.COMPUTE_LOGGER)
+   protected Logger logger = Logger.NULL;
+
+   @Inject
+   public AzureComputeSecurityGroupExtension(final AzureComputeApi api, @Named(OPERATION_TIMEOUT) long operationTimeout) {
```
nacx: Change the constructor to be package private.

azurecompute/src/main/java/org/jclouds/azurecompute/compute/extensions/AzureComputeSecurityGroupExtension.java
```diff
@@ -0,0 +1,460 @@
...
+      this.api = checkNotNull(api, "api");
+      this.operationTimeout = checkNotNull(operationTimeout, "operationTimeout");
+      this.operationSucceeded = retry(new Predicate<String>() {
+         public boolean apply(String input) {
+            final Operation operation = api.getOperationApi().get(input);
+            return operation.status() == Operation.Status.SUCCEEDED;
+         }
+      }, operationTimeout, 5, 5, SECONDS);
```
nacx: Same comment than above. Extract the predicate to a  provider and get it injected here directly.

azurecompute/src/main/java/org/jclouds/azurecompute/compute/extensions/AzureComputeSecurityGroupExtension.java
```diff
@@ -0,0 +1,460 @@
...
+   private Predicate<String> operationSucceeded;
+
+   @Resource
+   @Named(ComputeServiceConstants.COMPUTE_LOGGER)
+   protected Logger logger = Logger.NULL;
+
+   @Inject
+   public AzureComputeSecurityGroupExtension(final AzureComputeApi api, @Named(OPERATION_TIMEOUT) long operationTimeout) {
+      this.api = checkNotNull(api, "api");
+      this.operationTimeout = checkNotNull(operationTimeout, "operationTimeout");
```
nacx: Remove the null checks once the constructor is package private.

azurecompute/src/main/java/org/jclouds/azurecompute/compute/extensions/AzureComputeSecurityGroupExtension.java
```diff
@@ -0,0 +1,460 @@
...
+
+   /**
+    * @param name it represents both cloudservice and deployment name
+    * @return Set<SecurityGroup>
+    */
+   @Override
+   public Set<SecurityGroup> listSecurityGroupsForNode(String name) {
+      checkNotNull(name, "name");
+
+      final String virtualNetworkName;
```
nacx: Create the variable below. Don't create it here uninitialized.

azurecompute/src/main/java/org/jclouds/azurecompute/compute/extensions/AzureComputeSecurityGroupExtension.java
```diff
@@ -0,0 +1,460 @@
...
+                  }
+               }
+            }
+         }
+      }
+      String deleteRequestId = api.getNetworkSecurityGroupApi().delete(id);
+      if (!operationSucceeded.apply(deleteRequestId)) {
+         return false;
+      }
+      return true;
```
nacx: Just `return operationSucceeded.apply(deleteRequestId);`

azurecompute/src/main/java/org/jclouds/azurecompute/compute/extensions/AzureComputeSecurityGroupExtension.java
```diff
@@ -0,0 +1,460 @@
...
+                 public Deployment apply(CloudService cloudService) {
+                    return api.getDeploymentApiForService(cloudService.name()).get(cloudService.name());
+                 }
+              })
+              .filter(Predicates.notNull())
+              .toSet();
+      // TODO filter deployments
+      for (Deployment deployment : deployments) {
+         Deployment.VirtualIP virtualIP = Iterables.tryFind(deployment.virtualIPs(), Predicates.notNull()).orNull();
+         if (virtualIP == null) throw new IllegalStateException("");
```
nacx: Add some explanatory message to the exception.

azurecompute/src/main/java/org/jclouds/azurecompute/compute/extensions/AzureComputeSecurityGroupExtension.java
```diff
@@ -0,0 +1,460 @@
...
+         for (Role role : deployment.roles()) {
+            for (Role.ConfigurationSet configurationSet : role.configurationSets()) {
+               for (int i = ipPermission.getFromPort(); i <= ipPermission.getToPort(); i++) {
+                  String name = String.format(AzureComputeConstants.TCP_FORMAT, i, i);
+                  configurationSet.inputEndpoints().remove(createInputEndpoint(name, ipPermission.getIpProtocol()
+                          .name().toLowerCase(), virtualIP.address(), i));
+               }
+               String updateRoleRequestId = api.getVirtualMachineApiForDeploymentInService(deployment.name(), deployment.name()).updateRole(role.roleName(), role);
+               if (!operationSucceeded.apply(updateRoleRequestId)) {
+                  // TODO
```
nacx: Throw an exception here to fail? Also, implement the other similar TODOs in this class, to have all execution paths covered.

azurecompute/src/main/java/org/jclouds/azurecompute/compute/extensions/AzureComputeSecurityGroupExtension.java
```diff
@@ -0,0 +1,460 @@
...
+   public void waitForOperationCompletion(String operationId, String operationDescription) {
+      if (!operationSucceeded.apply(operationId)) {
+         final String warnMessage = format("%s) has not been completed within %sms.", operationDescription, Long.toString(operationTimeout));
+         logger.warn(warnMessage);
+         String illegalStateExceptionMessage = format("%s. Please, try by increasing `jclouds.azure.operation-timeout` and try again", warnMessage);
+         throw new IllegalStateException(illegalStateExceptionMessage);
+      } else {
+         logger.info("%s (id: %s) terminated successfully", operationDescription, operationId);
+      }
+   }
```
nacx: This method is already present in the compute service adapter. Extract it somewhere so it can be reused.

azurecompute/src/main/java/org/jclouds/azurecompute/compute/functions/DeploymentToNodeMetadata.java
```diff
@@ -16,16 +16,99 @@
...
+
+	@Inject
+	DeploymentToNodeMetadata(@Memoized Supplier<Set<? extends Location>> locations,
+										GroupNamingConvention.Factory namingConvention, OSImageToImage osImageToImage,
+										RoleSizeToHardware roleSizeToHardware, Map<String, Credentials> credentialStore) {
+		this.nodeNamingConvention = checkNotNull(namingConvention, "namingConvention").createWithoutPrefix();
+		this.locations = checkNotNull(locations, "locations");
+		this.osImageToImage = checkNotNull(osImageToImage, "osImageToImage");
+		this.roleSizeToHardware = checkNotNull(roleSizeToHardware, "roleSizeToHardware");
+		this.credentialStore = checkNotNull(credentialStore, "credentialStore cannot be null");
```
nacx: Since the constructor is package private and only the Guice injector will be able to call it, you can remove all the null checks. The injector will already fail if any parameter is missing.

azurecompute/src/main/java/org/jclouds/azurecompute/compute/functions/DeploymentToNodeMetadata.java
```diff
@@ -16,16 +16,99 @@
...
+	}
+
 	@Override
-	public NodeMetadata apply(Deployment input) {
-		return null;
+	public NodeMetadata apply(Deployment from) {
+		NodeMetadataBuilder builder = new NodeMetadataBuilder();
+		builder.ids(from.name());
```
nacx: Is the deployment name unique? Is it the *real* id of the node int he provider? Otherwise call separately the `id()` and the `providerId()` methods.

azurecompute/src/main/java/org/jclouds/azurecompute/compute/functions/DeploymentToNodeMetadata.java
```diff
@@ -16,16 +16,99 @@
...
+					  LocationPredicates.idEquals(from.getDatacenter().getId() + "")).orNull());
+		}
+		builder.group(nodeNamingConvention.groupInUniqueNameOrNull(from.getHostname()));
+		builder.hardware(roleSizeToHardware.apply(from.instanceSize()));
+		Image image = osImageToImage.apply(from);
+		if (image != null) {
+			builder.imageId(image.getId());
+			builder.operatingSystem(image.getOperatingSystem());
+		}
+		*/
```
nacx: It is important to have the location and image set. Can you uncomment this?

azurecompute/src/main/java/org/jclouds/azurecompute/compute/functions/DeploymentToNodeMetadata.java
```diff
@@ -16,16 +16,99 @@
...
+			}
+			builder.publicAddresses(publicIpAddresses);
+		}
+		Set<String> privateIpAddresses = Sets.newLinkedHashSet();
+		if (from.roleInstanceList() != null) {
+			for (Deployment.RoleInstance roleInstance : from.roleInstanceList()) {
+				privateIpAddresses.add(roleInstance.ipAddress());
+			}
+			builder.privateAddresses(privateIpAddresses);
+		}
```
nacx: Does the API return the credentials? If not, consider setting there from the credential store as a best-effort to get them. Take [DigitalOcean](https://github.com/jclouds/jclouds-labs/blob/master/digitalocean/src/main/java/org/jclouds/digitalocean/compute/functions/DropletToNodeMetadata.java#L115-L120) as an example.

azurecompute/src/main/java/org/jclouds/azurecompute/compute/functions/LocationToLocation.java
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * Converts an Datacenter into a Location.
+ */
+@Singleton
+public class LocationToLocation implements Function<Location, org.jclouds.domain.Location> {
+   private final JustProvider provider;
+
+   // allow us to lazy discover the provider of a resource
+   @Inject
+   public LocationToLocation(JustProvider provider) {
```
nacx: Same here. Make the constructor package private and remove the null checks.

azurecompute/src/main/java/org/jclouds/azurecompute/compute/functions/OSImageToImage.java
```diff
@@ -62,8 +62,8 @@ public Image apply(OSImage image) {
             .description(image.description())
             .status(Image.Status.AVAILABLE)
             .uri(image.mediaLink())
-            .providerId(image.name())
-            .location(createLocation(image.location()));
+            .providerId(image.name());
+            //.location(createLocation(image.location()));
```
nacx: Can we get this back?

azurecompute/src/main/java/org/jclouds/azurecompute/compute/strategy/UseNodeCredentialsButOverrideFromTemplate.java
```diff
@@ -0,0 +1,57 @@
...
+      if (options.getLoginUser() != null)
+         builder.user(template.getOptions().getLoginUser());
+      if (options.getLoginPassword() != null)
+         builder.password(options.getLoginPassword());
+      if (options.getLoginPrivateKey() != null)
+         builder.privateKey(options.getLoginPrivateKey());
+      if (options.shouldAuthenticateSudo() != null && options.shouldAuthenticateSudo())
+         builder.authenticateSudo(true);
+      return builder.build();
+   }
```
nacx: This seems to be something that other providers can benefit from. Consider moving this class to the jclouds-compute project.

azurecompute/src/main/java/org/jclouds/azurecompute/features/NetworkSecurityGroupApi.java
```diff
@@ -0,0 +1,167 @@
...
+   @Fallback(NullOnNotFoundOr404.class)
+   NetworkSecurityGroup getFullDetails(@PathParam("networkSecurityGroupName") String networkSecurityGroupName);
+
+   /**
+    *  Adds a Network Security Group to a subnet.
+    *
+    * @param virtualNetworkName
+    * @return
+    */
+   @Named("AddNetworkSecurityGroupToSubnet.")
```
nacx: Remove the trailing `.`

azurecompute/src/main/java/org/jclouds/azurecompute/features/NetworkSecurityGroupApi.java
```diff
@@ -0,0 +1,167 @@
...
+   @Produces(MediaType.APPLICATION_XML)
+   @ResponseParser(ParseRequestIdHeader.class)
+   String addToSubnet(@PathParam("virtualNetworkName") String virtualNetworkName,
+                      @PathParam("subnetName") String subnetName,
+                      @BinderParam(NetworkSecurityGroupToXML.class) NetworkSecurityGroup networkSecurityGroup);
+
+   /**
+    * Removes a Network Security Group from a subnet
+    */
+   @Named("RemoveNetworkSecurityGroupToSubnet.")
```
nacx: Remove the trailing `.`

azurecompute/src/main/java/org/jclouds/azurecompute/features/StorageAccountApi.java
```diff
@@ -0,0 +1,98 @@
...
+   String create(@BinderParam(StorageServiceParamsToXML.class) StorageServiceParams storageServiceParams);
+
+   /**
+    * https://management.core.windows.net/<subscription-id>/services/storageservices
+    */
+   @Named("GetStorageAccountDetails")
+   @GET
+   @Path("/{storageAccountName}")
+   @XMLResponseParser(StorageServiceHandler.class)
+   @Fallback(Fallbacks.VoidOnNotFoundOr404.class)
```
nacx: Change to `@Fallback(Fallbacks.NullOnNotFoundOr404.class)`.

azurecompute/src/main/java/org/jclouds/azurecompute/features/StorageAccountApi.java
```diff
@@ -0,0 +1,98 @@
...
+   List<StorageService> list(@PathParam("subscriptionId") String subscriptionId);
+
+   /**
+    * The Check Storage Account Name Availability operation checks to see if the specified storage account name is available, or if it has already been taken.
+    */
+   @Named("CheckStorageAccountNameAvailability")
+   @GET
+   @Path("/operations/isavailable/{storageAccountName}")
+   @XMLResponseParser(AvailabilityHandler.class)
+   @Fallback(NullOnNotFoundOr404.class)
```
nacx: What does it mean that a check returns `null`? Shouldn't it return just a boolean? if you want to return an object, shouldn't it be better to throw an exception if the check can not be performed? (I think it is not clear what a null means in this case: not available/the check operation failed).

Consider removing the fallback to propagate an exception.

azurecompute/src/main/java/org/jclouds/azurecompute/features/SubscriptionApi.java
```diff
@@ -0,0 +1,54 @@
...
+public interface SubscriptionApi {
+
+   /**
+    * The List Role Sizes request may be specified as follows. Replace <subscription-id> with the subscription ID.
+    */
+   @Named("ListRoleSizes")
+   @GET
+   @XMLResponseParser(ListRoleSizesHandler.class)
+   @Fallback(EmptyListOnNotFoundOr404.class)
+   List<RoleSize> list(@PathParam("subscriptionId") String subscriptionId);
```
nacx: Change the name to `listRoleSizes`? Otherwise it gives the impression that this method will list subscription objects.

azurecompute/src/main/java/org/jclouds/azurecompute/features/VirtualMachineApi.java
```diff
@@ -80,9 +85,30 @@ String capture(@PathParam("name") String name, @PayloadParam("imageName") String
...
+
+   /**
+    * https://msdn.microsoft.com/en-us/library/azure/jj157193.aspx
+    */
+   @Named("GetRole")
+   @GET
+   @Path("/roles/{roleName}")
+   @Produces(MediaType.APPLICATION_XML)
+   @ResponseParser(ParseRequestIdHeader.class)
+   Role getRole(@PathParam("roleName") String roleName);
```
nacx: Add a fallback to return `null` if the role is not found.

azurecompute/src/main/java/org/jclouds/azurecompute/features/VirtualNetworkApi.java
```diff
@@ -0,0 +1,74 @@
...
+    *
+    * @return The response body is a netcfg.cfg file.
+
+    */
+   @Named("GetVirtualNetworkConfiguration")
+   @Path("/media")
+   @GET
+   @XMLResponseParser(NetworkConfigurationHandler.class)
+   @Fallback(NullOnNotFoundOr404.class)
+   NetworkConfiguration get();
```
nacx: Change to `getNetworkConfiguration`. This returns a `NetworkConfiguration` but the `list` method returns `VirtualNetworkSite`. The naming can be confusing.

azurecompute/src/main/java/org/jclouds/azurecompute/features/VirtualNetworkApi.java
```diff
@@ -0,0 +1,74 @@
...
+
+   @Named("ListVirtualNetworkSites")
+   @Path("/virtualnetwork")
+   @GET
+   @XMLResponseParser(ListVirtualNetworkSitesHandler.class)
+   @Fallback(NullOnNotFoundOr404.class)
```
nacx: Don't return null. Return an empty list here.

azurecompute/src/test/java/org/jclouds/azurecompute/compute/AzureComputeServiceAdapterLiveTest.java
```diff
@@ -0,0 +1,139 @@
...
+
+   protected void doConnectViaSsh(Deployment deployment, LoginCredentials creds) {
+      SshClient ssh = sshFactory.create(HostAndPort.fromParts(deployment.virtualIPs().get(0).address(), 22), creds);
+      try {
+         ssh.connect();
+         ExecResponse hello = ssh.exec("echo hello");
+         assertEquals(hello.getOutput().trim(), "hello");
+         System.err.println(ssh.exec("df -k").getOutput());
+         System.err.println(ssh.exec("mount").getOutput());
+         System.err.println(ssh.exec("uname -a").getOutput());
```
nacx: Consider removing the console output.

azurecompute/src/test/java/org/jclouds/azurecompute/compute/AzureComputeServiceAdapterLiveTest.java
```diff
@@ -0,0 +1,139 @@
...
+   public void testListHardwareProfiles() {
+      Iterable<RoleSize> roleSizes = adapter.listHardwareProfiles();
+      assertFalse(Iterables.isEmpty(roleSizes));
+
+      for (RoleSize roleSize : roleSizes) {
+         assertNotNull(roleSize);
+      }
+   }
+
+   @AfterGroups(groups = "live")
```
nacx: Also configure it with an `alwayrRun = true` to make sure we don't leak resources when the tests fail.

azurecompute/src/test/java/org/jclouds/azurecompute/compute/AzureComputeServiceAdapterLiveTest.java
```diff
@@ -0,0 +1,139 @@
...
+
+   @Override
+   protected Iterable<Module> setupModules() {
+      return ImmutableSet.<Module> of(getLoggingModule(), new SshjSshClientModule());
+   }
+
+   @Override
+   protected Properties setupProperties() {
+      Properties properties = super.setupProperties();
+      properties.setProperty("jclouds.ssh.max-retries", "10");
```
nacx: Users should be able to override this by setting the value from the command line. Set it only if not present.

azurecompute/src/test/java/org/jclouds/azurecompute/compute/AzureComputeServiceContextLiveTest.java
```diff
@@ -0,0 +1,99 @@
...
+   }
+
+   @Override
+   protected Module getSshModule() {
+      return new SshjSshClientModule();
+   }
+
+   @Resource
+   @Named(ComputeServiceConstants.COMPUTE_LOGGER)
+   protected Logger logger = Logger.NULL;
```
nacx: This class is not instantiated by Guice, so the logger won't get injected.

azurecompute/src/test/java/org/jclouds/azurecompute/features/DeploymentApiLiveTest.java
```diff
@@ -0,0 +1,132 @@
...
+   public void testDelete() {
+      String requestId = api().delete(deployment.name());
+      assertTrue(operationSucceeded.apply(requestId), requestId);
+      Logger.getAnonymousLogger().info("operation succeeded: " + requestId);
+
+      assertTrue(deploymentGone.apply(deployment), deployment.toString());
+      Logger.getAnonymousLogger().info("deployment deleted: " + deployment);
+   }
+
+   @Override @AfterClass(groups = "live")
```
nacx: Same comment. Configure it with an `alwaysRun = true`.

azurecompute/src/test/java/org/jclouds/azurecompute/features/SubscriptionApiLiveTest.java
```diff
@@ -0,0 +1,42 @@
...
+ * limitations under the License.
+ */
+package org.jclouds.azurecompute.features;
+
+import static org.testng.Assert.assertNotNull;
+import org.jclouds.azurecompute.domain.RoleSize;
+import org.jclouds.azurecompute.internal.BaseAzureComputeApiLiveTest;
+import org.testng.annotations.Test;
+
+@Test(groups = "live", testName = "LocationApiLiveTest")
```
nacx: Test name should be "SubscriptionApiLiveTest".

azurecompute/src/test/java/org/jclouds/azurecompute/features/VirtualMachineApiLiveTest.java
```diff
@@ -0,0 +1,161 @@
...
+      String requestId = api().restart(roleName);
+      assertTrue(operationSucceeded.apply(requestId), requestId);
+      Logger.getAnonymousLogger().info("operation succeeded: " + requestId);
+
+      RoleInstance roleInstance = getFirstRoleInstanceInDeployment(DEPLOYMENT);
+      assertTrue(roleInstanceReady.apply(DEPLOYMENT), roleInstance.toString());
+      Logger.getAnonymousLogger().info("roleInstance restarted: " + roleInstance);
+   }
+
+   @AfterClass
```
nacx: Same comment about always running the tear down.

azurecompute/src/test/java/org/jclouds/azurecompute/features/VirtualNetworkApiLiveTest.java
```diff
@@ -0,0 +1,112 @@
...
+      operationSucceeded = retry(new Predicate<String>() {
+         public boolean apply(String input) {
+            return api.getOperationApi().get(input).status() == Operation.Status.SUCCEEDED;
+         }
+      }, 600, 5, 5, SECONDS);
+
+      originalNetworkConfiguration = api().get();
+   }
+
+   @AfterClass
```
nacx: Always tear down

nacx:
Thanks @andreaturli! This is quite good. Apart from the inline comments here are some general considerations:

* Move all global operations in the ComputeService adapter to the upper layer to void race conditions.
* In all live tests, add the `alwaysRun = true` to the clenaup method so it is executed even if tests fail. Let's try to not leak resources.
* There are missing mock tests, but those can be added in upcoming PRs.
* There should also be unit tests for all transformation functions.

Good job!

azurecompute/src/main/java/org/jclouds/azurecompute/compute/AzureComputeServiceAdapter.java
andreaturli: I need to be sure that the Role inside the Deployment is in READY_ROLE state, so I think it is needed

nacx: @eduardkoller could you give us a bit of insight? There are a couple conceptual things I don't have clear:

* The current impl uses a Deployment as a Node. Would it be more correct to use a Role? (We can address this in an upcoming PR once this starting one is merged).

Taking into account the current implementation:

* How do we determine the state of a VM? There is the state of the RoleInstance, but there is also the state of the Deployment. Which one should we take into account?
* Finally, jclouds already polls for the status of the node, so there is no need for this class to wait for that. Can we remove this active wait from here and just return the Deployment, regardless of its RoleInstance status?

Thanks!

hsbhathiya: @andreaturli 
 I think this is good place to continue the discussion we had so @nacx and Eduard can share their thoughts

As far as I understand  RoleInstance represents a node in azure API. So InstacneStatus would represent the state of VM. Deployment is more a collections of nodes .However RoleInstance to NodeMetadata would be bit  problematic as  Azure RoleInstance represenation not consist of some important data

My suggestion is to 
 1 - Introduce VirtualMachine in domain 
 2 - Map DeploymentToVirtualMachines where single deployment produces list of virtual machines
 3 - Map VirtualMachineToNodeMetadata
 4 - implement  ComputeServiceAdapter<VirtualMachine, RoleSize, OSImage, Location>

Proposed changes are partly implemented here
https://github.com/hsbhathiya/jclouds-labs/blob/MyServiceAdapter/azurecompute/src/main/java/org/jclouds/azurecompute/compute/AzureComputeServiceAdapter2.java

azurecompute/src/main/java/org/jclouds/azurecompute/compute/extensions/AzureComputeSecurityGroupExtension.java
andreaturli: Added that but for some reason github is still showing that

azurecompute/src/main/java/org/jclouds/azurecompute/compute/extensions/AzureComputeSecurityGroupExtension.java
```diff
@@ -0,0 +1,423 @@
...
+/**
+ * An extension to compute service to allow for the manipulation of {@link org.jclouds.compute.domain.SecurityGroup}s. Implementation
+ * is optional by providers.
+ *
+ * It considers only the custom rules added by the user and ignores the default rules created by Azure
+ */
+public class AzureComputeSecurityGroupExtension implements SecurityGroupExtension {
+
+   protected final AzureComputeApi api;
+   private Predicate<String> operationSucceededPredicate;
```
nacx: Make this final

azurecompute/src/main/java/org/jclouds/azurecompute/compute/extensions/AzureComputeSecurityGroupExtension.java
```diff
@@ -0,0 +1,423 @@
...
+
+   @Resource
+   @Named(ComputeServiceConstants.COMPUTE_LOGGER)
+   protected Logger logger = Logger.NULL;
+
+   @Inject
+   public AzureComputeSecurityGroupExtension(final AzureComputeApi api, Predicate<String> operationSucceededPredicate) {
```
nacx: Make the injection constructor package private.

azurecompute/src/main/java/org/jclouds/azurecompute/compute/extensions/AzureComputeSecurityGroupExtension.java
```diff
@@ -0,0 +1,423 @@
...
+
+   @Override
+   public SecurityGroup createSecurityGroup(final String name, Location location) {
+      checkNotNull(name, "name");
+      checkNotNull(location, "location");
+
+      final NetworkSecurityGroup networkSecurityGroup = NetworkSecurityGroup.create(name, name, location.getId(), null);
+      String createNSGRequestId =
+              api.getNetworkSecurityGroupApi().create(networkSecurityGroup);
+      operationSucceededPredicate.apply(createNSGRequestId);
```
nacx: What if the predicate returns `false`? Should we throw an exception instead of silently continuing? Apply the same reasoning to the rest of operations in this file.

nacx:
The ComputeService implementation looks much cleaner now. Goog job! There are just a few minor comments to be addressed.
There are also several comments that have not been considered or been answered that can be addressed in subsequenc PRs (but let's discuss them so we don't forget):

* As said [here](https://github.com/jclouds/jclouds-labs/pull/135#discussion-diff-24976948), should we use the `GroupNamingConvention` to generate the names?
* These comments regarding the `DeploymentToNodeMetadata`: [ids](https://github.com/jclouds/jclouds-labs/pull/135#discussion-diff-24977372), [commented code](https://github.com/jclouds/jclouds-labs/pull/135#discussion-diff-24977411) and [credentials](https://github.com/jclouds/jclouds-labs/pull/135#discussion-diff-24977465).

andreaturli:
@nacx thanks also for the summary!
Are we good in merging it now and then address the other issue later on?

andreaturli:
I've created sub-tasks on jira [JCLOUDS-664](https://issues.apache.org/jira/browse/JCLOUDS-664) so we don't forget your valuable commets not yet addressed

andreaturli:
@nacx squashed and rebased, let's wait for the builder and I'll merge it, ok?

andreaturli:
Merged at [e97ddaeeb59acc516e3d4d8398cf2da5e5492487](https://git1-us-west.apache.org/repos/asf?p=jclouds-labs.git;a=commit;h=e97ddaeeb59acc516e3d4d8398cf2da5e5492487)

PR #136, opened 2015-02-18
JCLOUDS-702: JClouds ProfitBricks provider - Snapshots API

baldwinSPC:
@nacx Here's the newest PR. We'll see if it passes build once submitted. This cleans it up so there are only 25 files. 

Let us know if there are any issues or changes you'd like made. Sorry about that last PR. 

Thanks.
@baldwinSPC & @jasminSPC


nacx:
Thanks, this is much cleaner. However, the files you changed (not the new ones) appear to be changed entirely, instead of only the relevant lines. Could this be due to file encoding or similar? Can you check that, so the diff *only* affects the real changed lines?

jasminSPC:
Snapshots were implemented from the scratch. So these are new files

nacx:
I know, but the following files were modified:
https://github.com/jclouds/jclouds-labs/pull/136/files#diff-fc9e881289bcbc04129422f4a9f3083e
https://github.com/jclouds/jclouds-labs/pull/136/files#diff-c589bd6e543e8b2e27e2c7bf60ada0e3
https://github.com/jclouds/jclouds-labs/pull/136/files#diff-a06995fa1f69af08ae2873a75d09814d
https://github.com/jclouds/jclouds-labs/pull/136/files#diff-43353bb4b23a894db0ae2d061e7cd863

And it appears that the entire files where modified while actually only few lines were added. Can you fix that?

jasminSPC:
I'll pull previous version and add my changes. Note this is going to
reflect only in three files. Other ones in the diff are whole new files.

profitbricks/src/test/java/org/jclouds/profitbricks/binder/snapshot/UpdateSnapshotRequestBinderTest.java
```diff
@@ -0,0 +1,59 @@
...
+package org.jclouds.profitbricks.binder.snapshot;
+
+import org.jclouds.profitbricks.domain.OsType;
+import org.jclouds.profitbricks.domain.Snapshot;
+import org.testng.annotations.Test;
+
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertNotNull;
+
+public class UpdateSnapshotRequestBinderTest {
```
nacx: Configure the test in the "unit" group.

profitbricks/src/main/java/org/jclouds/profitbricks/binder/snapshot/RollbackSnapshotRequestHandler.java
```diff
@@ -0,0 +1,42 @@
...
+ * limitations under the License.
+ */
+package org.jclouds.profitbricks.binder.snapshot;
+
+import org.jclouds.profitbricks.binder.BaseProfitBricksRequestBinder;
+import org.jclouds.profitbricks.domain.Snapshot;
+
+import static java.lang.String.format;
+
+public class RollbackSnapshotRequestHandler extends BaseProfitBricksRequestBinder<Snapshot.Request.RollbackPayload> {
```
nacx: Shouldn't this class be names `RollbackSnapshotRequestBinder`?

profitbricks/src/test/java/org/jclouds/profitbricks/features/SnapshotApiLiveTest.java
```diff
@@ -0,0 +1,132 @@
...
+        assertTrue(snapshots.size() > 0);
+    }
+
+    @Test
+    public void testGetSnapshot() {
+        Snapshot snapshot = api.snapshotApi().getSnapshot(snapshotId);
+
+        assertNotNull(snapshot);
+        assertEquals(snapshot.id(), snapshotId);
+    }
```
nacx: Test execution order is not guaranteed. Configure the tests with the `dependsOnMethod` attribute to make sure they run at the right time.

profitbricks/src/main/java/org/jclouds/profitbricks/features/SnapshotApi.java
```diff
@@ -0,0 +1,90 @@
...
+    @POST
+    @Named( "snapshot:delete" )
+    @Payload( "<ws:deleteSnapshot><snapshotId>{id}</snapshotId></ws:deleteSnapshot>" )
+    @Fallback( Fallbacks.FalseOnNotFoundOr404.class )
+    boolean deleteSnapshot( @PayloadParam( "id" ) String id );
+
+    @POST
+    @Named( "snapshot:rollback" )
+    @MapBinder( RollbackSnapshotRequestHandler.class )
+    @Fallback( Fallbacks.FalseOnNotFoundOr404.class )
```
nacx: Is this fallback really needed? Does the API return a 404?

profitbricks/src/test/java/org/jclouds/profitbricks/http/parser/snapshot/SnapshotListResponseHandlerTest.java
```diff
@@ -0,0 +1,99 @@
...
+import org.jclouds.profitbricks.domain.Location;
+import org.jclouds.profitbricks.domain.OsType;
+import org.jclouds.profitbricks.domain.ProvisioningState;
+import org.jclouds.profitbricks.domain.Snapshot;
+import org.jclouds.profitbricks.http.parser.BaseResponseHandlerTest;
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertNotNull;
+import org.testng.annotations.Test;
+
+public class SnapshotListResponseHandlerTest extends BaseResponseHandlerTest<List<Snapshot>> {
```
nacx: Configure the test in the "unit" group.

nacx:
Thanks @baldwinSPC & @jasminSPC! Code looks pretty good. Apart from the inline comments:

* Add a test class for the `RollbackSnapshotRequestHandler` (or `RollbackSnapshotRequestBinder` once renamed).
* Reformat your code according to the [jclouds coding standards](https://cwiki.apache.org/confluence/display/JCLOUDS/Coding+Standards). Basically use a 3 space indent and a 120 line length.

jasminSPC:
Should I do formatting in a separate commit?

nacx:
>Should I do formatting in a separate commit?

There's no need for that. Just add a new commit with all the fixes if you prefer.

jasminSPC:
Alright

jasminSPC:
I did the changes as suggested and I also formatted code as you suggested.

profitbricks/src/main/java/org/jclouds/profitbricks/features/SnapshotApi.java
```diff
@@ -0,0 +1,86 @@
...
+   @POST
+   @Named("snapshot:delete")
+   @Payload("<ws:deleteSnapshot><snapshotId>{id}</snapshotId></ws:deleteSnapshot>")
+   @Fallback(Fallbacks.FalseOnNotFoundOr404.class)
+   boolean deleteSnapshot(@PayloadParam("id") String id);
+
+   @POST
+   @Named("snapshot:rollback")
+   @MapBinder(RollbackSnapshotRequestBinder.class)
+   boolean rollbackSnapshot(@PayloadParam("snapshot") Snapshot.Request.RollbackPayload payload);
```
nacx: Does this method actually return a boolean? what does the API return, and how does it get transformed into a boolean?

profitbricks/src/main/java/org/jclouds/profitbricks/compute/internal/ProvisioningStatusPollingPredicate.java
```diff
@@ -24,36 +24,41 @@
...
+            case STORAGE:
+                return expect == api.storageApi().getStorage(input).state();
+            case SNAPSHOT:
+                return expect == api.snapshotApi().getSnapshot(input).state();
+            default:
+                throw new IllegalArgumentException("Unknown domain '" + domain + "'");
+        }
+    }
 
 }
```
nacx: Format this class according to the style guide.

profitbricks/src/main/java/org/jclouds/profitbricks/features/SnapshotApi.java
jasminSPC: It doesn't nor does delete but we use it the same way in delete. 

 <ns2:rollbackSnapshotResponse xmlns:ns2="http://ws.api.profitbricks.com/">
      <return>
        <requestId>request-id</requestId>
        <dataCenterId>datacenter-id</dataCenterId>
        <dataCenterVersion>datacenter-version</dataCenterVersion>
      </return>
    </ns2:rollbackSnapshotResponse>

I can switch the return type to return RequestId but I don't see any use of that.

nacx: So, if I understand this properly, this method will never return false. If it suceeds returns "true", and fails with an exception otherwise. It sounds better to me to return the requestId or just `void`. WDYT?

Regarding the delete method: does the API actually return a 404? If the API never returns a 404, then remove the fallback and apply the same criteria applied to the rollback method.

jasminSPC: Actually it doesn't return 404 but error message if SnapshotId is not provided. Which is a required field.
I will return RequestId then

nacx: Regarding the other methods with the `@Fallback` annotation, does the API ever return a 404?

jasminSPC: Actually I'm looking right now into my .net library and it never returns 404. Since it is SOAP service it is returning 500 and fault xml.
The thing is I was following how it is done in other features like servers https://github.com/StackPointCloud/jclouds-labs/blob/master/profitbricks/src/main/java/org/jclouds/profitbricks/features/ServerApi.java

nacx: Oh, right. For delete and list methods, the fallback should be OK. jclouds already [handles the fault xml](https://github.com/jclouds/jclouds-labs/blob/master/profitbricks/src/main/java/org/jclouds/profitbricks/http/ResponseStatusFromPayloadHttpCommandExecutorService.java#L95-L100) and returns a response with the appropriate response code, so forget about delete and the other existing methods.
I just didn't find the boolean return value that "meaningful" in the rollback method. Returning the request id looks more appropriate.

nacx:
There are again Windows line terminations. Can you fix them? Apart from that changes lgtm. After fixing the line endings, mind squashing the commits into a single one, so this can be cleanly merged?

jasminSPC:
Squashed.

nacx:
Hmmm The squash should leave this PR with just one single commit on top of the latest master.

PR #137, opened 2015-02-20
[JCLOUDS-664] Fixing live test execution on Azure

ilgrosso:
This PR provides fix for running live tests on Azure.

Warning: some white noise (mostly formatting) is also included, but in separate commit.


ilgrosso:
Still working on this, I'll open new one once all live tests will be passing.

PR #138, opened 2015-02-25
JClouds ProfitBricks provider - Snapshot API Vol.2

jasminSPC:
Here is the repeated PR for Profitbricks Snapshot API.

Can you please close the previous one https://github.com/jclouds/jclouds-labs/pull/136


profitbricks/src/test/java/org/jclouds/profitbricks/features/SnapshotApiLiveTest.java
```diff
@@ -0,0 +1,132 @@
...
+
+      snapshotId = snapshot.id();
+   }
+
+   @Test
+   public void testGetAllSnapshots() {
+      List<Snapshot> snapshots = api.snapshotApi().getAllSnapshots();
+
+      assertNotNull(snapshots);
+      assertTrue(snapshots.size() > 0);
```
nacx: This can fail if this test is executed *before* the `testCreateSnapshot`. Configure the test dependency properly.

nacx:
This looks good now, but apart from the last comment, there are several things that must be changed:

* The `SnapshotApiMockTest` is again in windows format.
* The `SnapshotApiLiveTest` is again in windows format.
* Remove the `ServerApiLiveTest` from this PR, as it is not related to this PR and only changes spaces by tabs (which is against the jclouds style guide).
* Remove the `ServerApiMockTest` from this PR, as it is not related to this PR and only changes spaces by tabs (which is against the jclouds style guide).

Please, fix this and the inline comment, in a new commit in the PR. I'll take care of squashing before merging.

jasminSPC:
I'm done with this.

nacx:
I think you misunderstood what I wanted to say (apologies if I didn't explain it properly). I said to remove the ServerApi mock and live tests from the PR, because they contained only formatting changes not related to this PR, but I didn't pretend to say "remove the files from the code". This PR now removes the files, but we want them. I just wanted to say: "undo the accidental changes that have been made to those files".

PR #141, opened 2015-02-26
Profitbricks IpBlock API

jasminSPC:
Profitbricks IpBlock API - Cleaned up

[Live test results](https://gist.github.com/jasminSPC/ae25152f013337119e45)


profitbricks/src/main/java/org/jclouds/profitbricks/ProfitBricksApi.java
```diff
@@ -21,11 +21,11 @@
 import org.jclouds.profitbricks.features.DataCenterApi;
 import org.jclouds.profitbricks.features.FirewallApi;
 import org.jclouds.profitbricks.features.ImageApi;
+import org.jclouds.profitbricks.features.IpBlockApi;
 import org.jclouds.profitbricks.features.NicApi;
 import org.jclouds.profitbricks.features.ServerApi;
 import org.jclouds.profitbricks.features.StorageApi;
 import org.jclouds.rest.annotations.Delegate;
-
```
nacx: Add this line back

profitbricks/src/main/java/org/jclouds/profitbricks/domain/IpBlock.java
```diff
@@ -0,0 +1,125 @@
...
+
+   public abstract Location location();
+
+   public abstract List<PublicIp> publicIps();
+
+   @Nullable
+   public abstract List<String> ips();
+
+   public static IpBlock create(String id, Location location, List<PublicIp> publicIps, List<String> ips) {
+      return new AutoValue_IpBlock(id, location, publicIps, ips);
```
nacx: To enforce the immutability of the list, if not null you should pass an `ImmutableList.copyOf(ips)` to the auto value constructor.

profitbricks/src/main/java/org/jclouds/profitbricks/http/parser/publicip/BasePublicIpResponseHandler.java
```diff
@@ -0,0 +1,40 @@
...
+
+import com.google.inject.Inject;
+import org.jclouds.profitbricks.domain.IpBlock;
+import org.jclouds.profitbricks.http.parser.BaseProfitBricksResponseHandler;
+
+public abstract class BasePublicIpResponseHandler<T> extends BaseProfitBricksResponseHandler<T> {
+
+   protected IpBlock.PublicIp.Builder builder;
+
+   @Inject
```
nacx: This has no effect here. Remove it and delegate the configuration to the subclasses.

profitbricks/src/test/java/org/jclouds/profitbricks/features/IpBlockApiLiveTest.java
```diff
@@ -0,0 +1,92 @@
...
+import static org.testng.Assert.assertFalse;
+import static org.testng.Assert.assertNotNull;
+import org.testng.annotations.Test;
+
+@Test(groups = "live", testName = "IpBlockApiLiveTest", singleThreaded = true)
+public class IpBlockApiLiveTest extends BaseProfitBricksLiveTest {
+
+   String nicid;
+   IpBlock newIpBlock;
+   String ip;
```
nacx: Any reason why these variables are package private and not private?

profitbricks/src/test/java/org/jclouds/profitbricks/features/IpBlockApiLiveTest.java
```diff
@@ -0,0 +1,92 @@
...
+
+   @Override
+   public void initialize() {
+      super.initialize();
+
+      List<Nic> nics = api.nicApi().getAllNics();
+
+      assertFalse(nics.isEmpty(), "At least one NIC is requred to test IpBlocks");
```
nacx: Is this something that can be setup or is it a precondition that will be met by default?
I mean, a user with a default and clean account, will need to perform a manual configuration to be able to run thses live tests?

profitbricks/src/test/java/org/jclouds/profitbricks/features/IpBlockApiLiveTest.java
```diff
@@ -0,0 +1,92 @@
...
+      assertNotNull(newIpBlock);
+      assertNotNull(newIpBlock.ips());
+      assertFalse(newIpBlock.ips().isEmpty());
+   }
+
+   @Test
+   public void testGetAllIpBlocks() {
+      List<IpBlock> ipBlocks = api.ipBlockApi().getAllIpBlock();
+
+      assertNotNull(ipBlocks);
```
nacx: The jclouds fallback always returns a list (or an empty one). Should we better add some assertion to verify that the list has the expected elements?

profitbricks/src/test/java/org/jclouds/profitbricks/features/IpBlockApiMockTest.java
```diff
@@ -0,0 +1,181 @@
...
+
+      try {
+         List<IpBlock> ipBlocks = api.getAllIpBlock();
+         assertRequestHasCommonProperties(server.takeRequest());
+         assertNotNull(ipBlocks);
+      } finally {
+         pbApi.close();
+         server.shutdown();
+      }
+   }
```
nacx: The get all blocks method also defines a 404 fallback. Add the corresponding test.

profitbricks/src/main/java/org/jclouds/profitbricks/http/parser/publicip/PublicIpListResponseHandler.java
```diff
@@ -0,0 +1,51 @@
...
+   public void reset() {
+      this.publicIps = Lists.newArrayList();
+   }
+
+   @Override
+   public List<PublicIp> getResult() {
+      return publicIps;
+   }
+
+}
```
nacx: Add a unit test for this class.

profitbricks/src/test/java/org/jclouds/profitbricks/features/IpBlockApiLiveTest.java
baldwinSPC: @nacx Thanks for reviewing all of this. This live test assumes the user has already built out their environment via API or UI with a datacenter, server, volume, and NIC. So, this is a precondition that will be met by default. This is very similar to the Server live tests where a datacenter is required before running the test.

profitbricks/src/main/java/org/jclouds/profitbricks/http/parser/publicip/PublicIpListResponseHandler.java
jasminSPC: This is the same case like this one Firewall.Rule
https://github.com/jclouds/jclouds-labs/blob/master/profitbricks/src/main/java/org/jclouds/profitbricks/http/parser/firewall/rule/FirewallRuleListResponseHandler.java

jasminSPC: In our IRC conversation we said to ignore this one.

profitbricks/src/main/java/org/jclouds/profitbricks/domain/IpBlock.java
```diff
@@ -0,0 +1,126 @@
...
+@AutoValue
+public abstract class IpBlock {
+
+   public abstract String id();
+
+   public abstract Location location();
+
+   public abstract List<PublicIp> publicIps();
+
+   @Nullable
```
nacx: This is no longer nullable since the factory method sets it to an empty list if the input parameter is null. Remove the annotation or keep the null value when the input is null.

nacx:
Thanks @jasminSPC! Just one comment remaining. Feel free to squash the commits into a single one when addressing it, and I'll merge the PR. Thanks!

jasminSPC:
This has been squashed and ready for merge.

nacx:
Pushed to master as [f5605dd4](https://git1-us-west.apache.org/repos/asf?p=jclouds-labs.git;a=commit;h=f5605dd4). Thanks @jasminSPC!

PR #145, opened 2015-03-03
JCLOUDS-702: JClouds ProfitBricks provider - ComputeServiceAdapter

devcsrj:
Hi,

This implements `ProfitBricksComputeServiceAdapter`. The PR is expected to fail, since I have yet to resolve issues (see comments on diff). Thoughts?

Aside from the mentioned issues, the `Function` classes are ready to be reviewed.

/cc @nacx


profitbricks/src/main/java/org/jclouds/profitbricks/config/ProfitBricksComputeServiceContextModule.java
```diff
@@ -0,0 +1,69 @@
...
+
+public class ProfitBricksComputeServiceContextModule extends
+	ComputeServiceAdapterContextModule<Server, Hardware, Image, DataCenter> {
+
+   @Override
+   protected void configure() {
+      super.configure();
+
+      // FIXME Not working
+      install(new LocationsFromComputeServiceAdapterModule<Server, Hardware, Image, DataCenter>() {
```
devcsrj: For some reason, this module doesn't get installed. Guice still complains that there are no bound implementation for `Supplier<java.util.Set<? extends org.jclouds.domain.Location>>` when used in the [ServerToNodeMetadata](https://github.com/jclouds/jclouds-labs/pull/145/files#diff-56ae535b8c57334bd467e33f42c740b9R63) constructor.

profitbricks/src/main/java/org/jclouds/profitbricks/compute/ProfitBricksComputeServiceAdapter.java
```diff
@@ -0,0 +1,267 @@
...
+
+      Server server = getNode(serverId);
+
+      // TODO Consult jclouds committers. Server credentials in profitbricks are sent 
+      // via email used in credentials.identity. We can set a password during 
+      // storage provisioning, maybe generate random characters?
+      LoginCredentials serverCredentials = LoginCredentials.builder()
+	      .user("root")
+	      .password("<password_here>")
+	      .build();
```
devcsrj: (on comment //) Recommendation?

nacx: jclouds needs the node credentials to be able to connect to the node when using the "runScript" methods. If credentials are configured per-node (instead of having default credentials per image), we should return them here.

If the password can be set during the node provisioning, that's perfect. If the user has set a password using the `templateOptions.overrideLoginPassword`, let's use that one, otherwise auto-generate a random one. The same applies to the user.

profitbricks/src/main/java/org/jclouds/profitbricks/config/ProfitBricksComputeServiceContextModule.java
nacx: I think you have to qualify the supplier as `@Memoized` in the `ServerToNodeMetadata`. Something like:
```java
@Inject
 ServerToNodeMetadata(Function<Storage, Volume> fnVolume,
      @Memoized Supplier<Set<? extends Location>> locationsSupply) {
```

nacx: That's because it is qualified when bound in the module that configures it:
https://github.com/jclouds/jclouds/blob/master/core/src/main/java/org/jclouds/location/config/LocationModule.java#L102-L106

profitbricks/src/main/java/org/jclouds/profitbricks/compute/ProfitBricksComputeServiceAdapter.java
```diff
@@ -0,0 +1,269 @@
...
+   protected Logger logger = Logger.NULL;
+
+   private final ProfitBricksApi api;
+   private final Predicate<String> waitDcUntilAvailable;
+
+   @Inject
+   ProfitBricksComputeServiceAdapter(ProfitBricksApi api) {
+      this.api = api;
+      this.waitDcUntilAvailable = Predicates2.retry(new ProvisioningStatusPollingPredicate(
+              api, ProvisioningStatusAware.DATACENTER, ProvisioningState.AVAILABLE), 10l * 60l, 2l, TimeUnit.SECONDS);
```
nacx: Consider making the predicate injectable. That would facilitate making the timeouts configurable so users can customise them. Take a look at [Azure](https://github.com/jclouds/jclouds-labs/blob/master/azurecompute/src/main/java/org/jclouds/azurecompute/compute/config/AzureComputeServiceContextModule.java#L95-L101) to see an example.

profitbricks/src/main/java/org/jclouds/profitbricks/compute/ProfitBricksComputeServiceAdapter.java
```diff
@@ -0,0 +1,269 @@
...
+
+      Double cores = 0d;
+      for (Processor processor : hardware.getProcessors())
+         cores += processor.getCores();
+
+      String storageBootDeviceId = Iterables.getFirst(storageIds, "");
+      Server.Request.CreatePayload serverRequest = Server.Request.creatingBuilder()
+              .dataCenterId(dataCenterId)
+              .name(name)
+              .bootFromStorageId(storageBootDeviceId)
```
nacx: What happens when the value is an empty string?

profitbricks/src/main/java/org/jclouds/profitbricks/compute/ProfitBricksComputeServiceAdapter.java
```diff
@@ -0,0 +1,269 @@
...
+
+      Server server = getNode(serverId);
+
+      LoginCredentials serverCredentials = LoginCredentials.builder()
+              .user("root") // always root user
+              .password(password)
+              .build();
+
+      return new NodeAndInitialCredentials<Server>(server, serverId, serverCredentials);
+   }
```
nacx: This method creates several resources in the provider. If it fails at some point, we should take care of rolling back the created stuff (at least we should try), in order to avoid leaking resources.

profitbricks/src/main/java/org/jclouds/profitbricks/compute/ProfitBricksComputeServiceAdapter.java
```diff
@@ -0,0 +1,269 @@
...
+
+   @Override
+   public void destroyNode(String id) {
+      ServerApi serverApi = api.serverApi();
+      Server server = serverApi.getServer(id);
+      if (server != null) {
+         String dataCenterId = server.dataCenter().id();
+         for (Storage storage : server.storages()) {
+            waitDcUntilAvailable.apply(dataCenterId);
+            api.storageApi().deleteStorage(storage.id());
```
nacx: try/catch each operation to delete as much as we can if something fails?

profitbricks/src/main/java/org/jclouds/profitbricks/compute/function/ImageToImage.java
```diff
@@ -0,0 +1,85 @@
...
+	      .description(osFamily.value())
+	      .family(osFamily)
+	      .version(parseVersion(desc))
+	      .is64Bit(is64Bit(desc))
+	      .build();
+
+      return new ImageBuilder()
+	      .ids(from.id())
+	      .name(desc)
+	      .location(mapLocation(from.location()))
```
nacx: Images are available in all regions by default? (Just asking as the mapLocation function always sets the PROVIDER scope.

profitbricks/src/main/java/org/jclouds/profitbricks/compute/function/ServerToNodeMetadata.java
```diff
@@ -0,0 +1,173 @@
...
+	    return NodeMetadata.Status.UNRECOGNIZED;
+      }
+   }
+
+   static OperatingSystem mapOsType(OsType osType) {
+      OperatingSystem os = OperatingSystem.builder()
+	      .description(OsFamily.UNRECOGNIZED.value())
+	      .family(OsFamily.UNRECOGNIZED)
+	      .build();
+      if (os != null)
```
nacx: This is never null at this point.

profitbricks/src/test/java/org/jclouds/profitbricks/compute/ProfitBricksComputeServiceAdapterLiveTest.java
```diff
@@ -0,0 +1,162 @@
...
+import com.google.inject.Injector;
+import com.google.inject.Module;
+import org.jclouds.compute.ComputeService;
+import org.jclouds.compute.ComputeServiceAdapter.NodeAndInitialCredentials;
+import org.jclouds.compute.domain.OsFamily;
+import org.jclouds.compute.domain.Template;
+import org.jclouds.compute.domain.TemplateBuilder;
+
+@Test(groups = "live", singleThreaded = true, testName = "ProfitBricksComputeServiceAdapterLiveTest")
+public class ProfitBricksComputeServiceAdapterLiveTest extends BaseProfitBricksLiveTest {
```
nacx: You don't have to build a custom test for this. You must extend the [BaseComputeServiceLiveTest](https://github.com/jclouds/jclouds/blob/master/compute/src/test/java/org/jclouds/compute/internal/BaseComputeServiceLiveTest.java). That is the jclouds contract. Once that test passes, the ComputeService implementation is working as expected. You can take a look at the [DigitalOceanComputeServiceLiveTest](https://github.com/jclouds/jclouds-labs/blob/master/digitalocean/src/test/java/org/jclouds/digitalocean/compute/DigitalOceanComputeServiceLiveTest.java) for an example.

nacx:
This is a pretty good start @devcsrj! Just commented some things I've seen in a quick review, and added an important comment about the live test. If you change it to extend the BaseComputeServiceLive test you will see what we expect, and you'll be able to build the ComputeService keeping that test as a goal.

profitbricks/src/main/java/org/jclouds/profitbricks/compute/function/ImageToImage.java
devcsrj: Hmm. I couldn't tell if it it qualifies to the __available in all regions__ statement. Let me describe it instead.

So given is an image called `Ubuntu-14.04-LTS-server-2015-03-01`. This exists in [Location](https://github.com/jclouds/jclouds-labs/pull/145/files#diff-8dba16965654b3b80ba6cc57f45b8278L19)s: `de/fra`, `de/fkb` but with different `id`s. There's also a _similar_ image called `Ubuntu-14.04-LTS-server-2015-03-02` in `us/las` and `us/lasdev` also different IDs. Does this qualify to the statement? (Does this present an issue?)

Also, I thought of `LocationScope` as a hierarchy. So `Provider` -> `Region` -> `Zone`, etc., and so I mapped it like `Location`(=`Provider`) -> `DataCenter`(`=Region`). Correct me if I assumed incorrectly

devcsrj:
[Live test results](https://gist.github.com/devcsrj/828e5be44a54f5217cbd). `Status=FAILING`

profitbricks/src/main/java/org/jclouds/profitbricks/compute/function/ImageToImage.java
nacx: The location hierarchy makes sense. jclouds, however, already provides the PROVIDER location through the `JustProvider` supplier. You can inject it in the DataCenterToLocation function and use it. Have a look at how [digitalocean](https://github.com/jclouds/jclouds-labs/blob/master/digitalocean/src/main/java/org/jclouds/digitalocean/compute/functions/RegionToLocation.java) does this.

Regarding the images that don't exist in all locations:

* The ComputeServiceAdapter must return each existing image, for each location. One image for each different location.
* The providerId of the image must be set to the *real* id of the image in the provider.
* The "id" of the image is the one jclouds will use to identify it (for example when calling the `geImage(id)` in the ComputeService or when using the `imageId` template option. If the same image id is available in multiple locations, usually providers encode the region in this field. Something like: "us-west-1/image-id". If ids are unique it is ok to use the same id for the providerId and the id (what the `ids` method in the builder does). Otherwise you'll have to make the "id" field able to distinguish between each image in each region. You can have a look at how the OpenStack provider uses the [RegionAndId](https://github.com/jclouds/jclouds/blob/master/apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/domain/regionscoped/RegionAndId.java) class.
* Finally, if images are by default available in **all** locations, the "location" field should be left null, otherwise it should be set to the right location (region) instead of being set to the PROVIDER scoped one. You can do this by injecting the locations supplier here as you do in the `ServerToNodeMetadata` to find the location where the image is available. This assumes that the ComputeService returns one copy of the image for each location where it is available, so you can make sure you return there all you need to build this properly here.

devcsrj:
[Live test results](https://gist.github.com/devcsrj/828e5be44a54f5217cbd/5c7509e5e20aa2ddb43dd48ca570b3ba9e1e8754)
`Tests run: 25, Failures: 4, Errors: 0, Skipped: 10` (48:46.720s, ugh.)

devcsrj:
Hi @nacx !
So 11 of the live tests were green; the failed tests were related to the ssh flow like:
<pre>
org.jclouds.ssh.SshException: (root:pw[ca4cb85a391831b122e3277b88393d3f]@46.16.73.251:22) (root:pw[ca4cb85a391831b122e3277b88393d3f]@46.16.73.251:22) error acquiring ExecResponse(command=[rm /tmp/init-configure-jetty]) (out of retries - max 5): Read timed out
</pre>
I tried to ssh to the said nodes, and it turns out that this was due to the operating system requesting a password change after logging in; thus, hanging for so long waiting for an input. It's a really weird behavior IMO, since I thought this was only applicable if I let the API generate it's own password. (I emailed PB tech support regards this.)

Assuming this was by design, my next step would be is to connect via SSH right away (to complete the prerequisites first) before returning the `NodeAndInitialCredentials`. I'm not sure if there are _hooks_ for something like this, but WDYT? Suggestions?

nacx:
Yeah, live tests take their time :)

So that password change is an issue, as it is interactive and we can't workaround that easily (but if you find a way, it would be very welcome!). The same happened with CloudSigma, and we ended up using a template that supported cloud-init and did not request the password change.

You can pass the "jclouds.template" property to set the template that the live tests will use, and you should configure the property with a template that works in the provider's metadata default properties (look at CloudSigma or DigitalOcean).

Let's see if there is an image that does not require the password change.

baldwinSPC:
@devcsrj @nacx Hey guys, is there any assistance we can provide on this PR or is it ready to go? Thanks.

devcsrj:
Hi @baldwinSPC ! 

I'm still working on passing the live tests for the compute service adapter. Right now, the issue is that the default images from ProfitBricks are built with password change a requirement upon first login, and so the ssh commands executed by tests are actually left hanging. I'm thinking of uploading a custom one that we can use for the live test. ([see also](https://devops.profitbricks.com/community/an-image-without-password-change/))

With that said, if you could get the live test pass with a viable approach, that'll be much much welcome. :)

baldwinSPC:
Hey @devcsrj 

So, a couple ways we could handle this:

1. We could do the image and push it to our Virtual Data Center, then we perform the live tests against our account. 
2. We could setup a dedicated jclouds account using the new Developer Program at PB. 
3. You could spin up a VM using their DCD tool, set the password, then create a snapshot and use that snapshot as the base image for any build requests. This would remove the password change requirement. 

I think long term (2) would be ideal. Short term (3) might be pretty do-able depending on your time. Let me know if you'd like us to do (1) and then just pass you the creds or have us walk through the steps for the validation. 

-matt

devcsrj:
Hi @baldwinSPC ,

(2 + 3) :+1:  Sounds about the right step, since for this provider to be promoted out of labs, it'll need a testing account. Btw, can't seem to find the _new developer program_. Is there some link or was it via email or something?

(3) This was the suggestion from Flavius (PB S.Engr) as well. Although I'm thinking: AFAIK jclouds doesn't necessarily have a concept for _snapshots_. I initially thought of maybe, categorizing _snapshots_ as an **image** (i.e.: a `Provisionable`, a resource where a node can be created). This way, the provider will search for both snapshots and image when creating nodes. I'll ask for the commit-ers' opinion after they release the 1.9.0.

In any case, feel free to continue where I left off. Kinda busy on my end, so I haven't put much time here yet. :) With that said, thanks for the suggestions!

baldwinSPC:
@devcsrj We've been able to score a test account for jclouds at ProfitBricks. We're waiting on the confirmation so we can log in. If we can go the snapshots route then we'll build a server and snapshot it. If not, then we'll take a different approach.

baldwinSPC:
@devcsrj I should also ask how you'd like us to convey the creds to you for testing. :)

devcsrj:
@baldwinSPC SCORE? AWESOME. :+1: 

As soon as you get the confirmation, go ahead and do the snapshot route. :)

Creds via email will do fine. :)

nacx:
Option 2 sounds like the way to go. If we do that, then any user that wants to automate stuff when deploying to ProfitBricks with jclouds will have to use a custom image that should have been created before, right?
Out of curiosity, how are your users (PB users) automating stuff?

Thanks for joining the discussion and for your help @baldwinSPC!

baldwinSPC:
Correct, if you wish to automate against PB you would need to leverage a
custom image that precludes the password change requirement; however, I
believe they are moving towards supporting things such as SSH keys,
user-data via cloud-init, and other things that would alleviate this long
term. It certainly puts a wrinkle in any full automation.

At the moment, we're building the tooling that would allow automation to
start to happen -- working in things like Ansible, Libcloud, Fog, along
with base libraries for their current API. Currently, users build custom
images and then automate against the API if doing infrastructure
automation.

devcsrj:
TODO: Live test

devcsrj:
TODO: Live test

baldwinSPC:
@devcsrj Just checking in on the live testing. How's it going?

devcsrj:
Yo @baldwinSPC . Haven't tried in the non-dev cluster yet, but will, tomorrow. :)

Also, will apply more changes based on conversation with Ignasi at IRC.

devcsrj:
[Live test results](https://gist.github.com/devcsrj/828e5be44a54f5217cbd/60eca3d2eb1fe1d07bfdec1c9c41ec03d0458059).
`Tests run: 25, Failures: 3, Errors: 0, Skipped: 10, at 38:40.672s`

profitbricks/src/main/java/org/jclouds/profitbricks/compute/ProfitBricksComputeServiceAdapter.java
```diff
@@ -0,0 +1,395 @@
...
+            logger.trace(">> storage connected.");
+         } catch (Exception ex) {
+            // delete unconnected storage
+            logger.warn(ex, ">> failed to connect storage '%s'. deleting..", storageId);
+            waitDcUntilAvailable.apply(dataCenterId);
+            destroyStorage(storageId, dataCenterId);
+         }
+      }
+
+      waitDcUntilAvailable.apply(dataCenterId);
```
nacx: Is this wait needed? Each step in the loop/catch will already perform an active wait.

nacx:
Test failures seem to be due to auth issues. Are you using the mentioned custom image?

profitbricks/src/main/java/org/jclouds/profitbricks/compute/ProfitBricksComputeServiceAdapter.java
devcsrj: Hmm. Yeah it's no longer needed. I was too paranoid of the datacenter not being ready I guess. Haha

devcsrj:
Yes, I'm using a custom image. Not exactly an auth issue (because the `testCreateAndRunAService` succeeded; was able to install configure jetty, etc). 

Upon further investigation, the servers become inaccessible via ssh after some `n` servers are created. Currently, the implementation [creates a lan for every server](https://github.com/jclouds/jclouds-labs/pull/145/files#diff-1a27cebd5920714ae4c207026edd5255R143). I'll try testing afterwards by explicitly setting which lan to connect. Might not be ideal though, thoughts?

devcsrj:
TODO: Live test

baldwinSPC:
@nacx is this a PR that can be merged yet? @devcsrj would you like me or @jasminSPC to do anything on this to close it out?

nacx:
The only thing I'm missing in this PR is a test that extends the [BaseTemplateBuilderLiveTest](https://github.com/jclouds/jclouds/blob/master/compute/src/test/java/org/jclouds/compute/internal/BaseTemplateBuilderLiveTest.java). This test, in conjunction with the adapter one are the compute service contract. Once both are passing, the provider is ready to go.

> Encountering what seem to be networking issues in the provider; thus, making jclouds unable to connect via ssh. Needs further investigation.

@devcsrj Is this the cause of the current live test failures? In that case, would it be safe to say that the current code should be good to go?

devcsrj:
Hi @nacx ! While this provider can already provision nodes successfully, I haven't been able to pass all tests required by the `BaseComputeServiceLiveTest`. The network issue is resolved by increasing the timeout. The current issue I need to resolve is that the tests [require](https://github.com/jclouds/jclouds/blob/master/compute/src/test/java/org/jclouds/compute/internal/BaseComputeServiceLiveTest.java#L232) that the `hostname` be retrieved via API (which isn't possible). Yet to think a workaround.

Hi @baldwinSPC , @jasminSPC ! I haven't put much time in this PR in a while, but I'd appreciate if you could run the live tests on your end as well and post a link to the results as gist (or workaround for the issues too!). Here's the command I'm using:
$ `mvn clean integration-test -Plive -Dtest=org.jclouds.profitbricks.compute.ProfitBricksTemplateBuilderLiveTest -Dtest.profitbricks.identity='<username>' -Dtest.profitbricks.credential='<password>' -Dtest.profitbricks.template='locationId=16c7ffee-70d8-4505-8793-d2b32795e4fb,imageNameMatches=jclouds-ubuntu14.10-template,loginUser=root:asdfghjk'`

nacx:
>The current issue I need to resolve is that the tests require that the hostname be retrieved via API (which isn't possible). Yet to think a workaround.

Oh, I see. In this case, I'd say don't worry and go ahead overwriting the hostname check, as we do in [cloudsigma](https://github.com/jclouds/jclouds-labs/blob/master/cloudsigma2/src/test/java/org/jclouds/cloudsigma2/compute/CloudSigma2ComputeServiceLiveTest.java#L51-L54). Not all providers offer the same feature set (for example not all providers have user metadata, etc), so it is OK to relax/remove some checks. The important thing is that the node provisioning flow works as expected and that all supported template options are also taken into account.

devcsrj:
(TODO: Live test)

devcsrj:
[Live test results.](https://gist.github.com/devcsrj/828e5be44a54f5217cbd/088324b581f7f528144fc43d2c42cb1f529d60fb)
`Tests run: 25, Failures: 5, Errors: 0, Skipped: 11`
Weirdly enough, when doing debug mode (adding `-Dmaven.surefire.debug` and breakpoint at [testGet](https://github.com/jclouds/jclouds/blob/master/compute/src/test/java/org/jclouds/compute/internal/BaseComputeServiceLiveTest.java#L521)) the test doesn't stop, as opposed from previous comment.)

As it turns out, it's not possible to concurrently execute methods in a datacenter. (i.e.: when using `client.rebootNodesMatching(..)`, other nodes would actually fail to receive the request because the API complains that a `PROVISIONING_IN_PROCESS`.) Need to refactor `ProfitBricksComputeServiceAdapter` to use a single-thread worker per datacenter approach.

nacx:
Interesting. IIRC a similar thing happens in Azure. It is resolved there by using a [ConflictManagementPredicate](https://github.com/jclouds/jclouds-labs/blob/master/azurecompute/src/main/java/org/jclouds/azurecompute/util/ConflictManagementPredicate.java) that tries to complete the operation until it fails or times out. You can see an example usage [here](https://github.com/jclouds/jclouds-labs/blob/master/azurecompute/src/main/java/org/jclouds/azurecompute/compute/AzureComputeServiceAdapter.java#L146-L159). Just a tip!

devcsrj:
(TODO: Live test)

devcsrj:
[Live test results.](https://gist.github.com/devcsrj/828e5be44a54f5217cbd/8f63cd1be9ff89f5c5959316258d75d7805da49f)
`Tests run: 25, Failures: 0, Errors: 0, Skipped: 0`

Hi @nacx ! With the live tests finally passing, you can review the changes again. With the latest impl, I chose to delegate *provisioning methods* in  a [ProvisioningManager](https://github.com/devcsrj/jclouds-labs/commit/a6ff7dd4f465d94a8063a3ac7a743c1a169406e5#diff-c67d5d74a6059f235c249f2e9a00a176R42), which holds a single-threaded executor per *logical group*, in this case, a datacenter. WDYT?

profitbricks/pom.xml
```diff
@@ -36,7 +36,11 @@
         <test.profitbricks.identity>FIXME</test.profitbricks.identity>
         <test.profitbricks.credential>FIXME</test.profitbricks.credential>
         <test.profitbricks.api-version>1.3</test.profitbricks.api-version>
-        <test.profitbricks.template />
+        <test.profitbricks.template>
+            locationId=45a8667e-edfe-4f64-8a4d-79d53c8fec1f,
+            imageNameMatches=jclouds-ubuntu14.10-template,
+            loginUser=root:asdfghjk
+        </test.profitbricks.template>
```
nacx: Is this a public template anyone will be able to use, or is it a custom one in your account? If it is a private one and the ones provided by default by Profitbricks don't provide remote access, I'd add a README with the instructions to build a  *jclouds friendly* image and how to run the live tests using it.

nacx:
I've had a look at the adapter and looks good. The ProvisioningManager approach looks clean! A couple questions:

* When is the ProvisioningManager closed? I haven't seen a call to its close method and we need to make sure its executors are closed.

I'd suggest to take a slightly different approach, to bind the ProvisioningManager to the jclouds context lifecycle, but that would require some (minor) changes to make it better managed by Guice:

* Configure the ProvisioningManager in a *provider* method in the Compute module, so it can be added to the jclouds Closer. Have a look at the [ExecutorServiceModule](https://github.com/jclouds/jclouds/blob/master/core/src/main/java/org/jclouds/concurrent/config/ExecutorServiceModule.java#L135-L138) to see an example. This way it will be closed then the jclouds context is closed.
* Given that the ProvisioningJob is always about the datacenter, I would specialise it, and configure it with Guice so it gets the predicate already injected. That would leave the adapter code cleaner. To do that, you'll have to use the Guice assisted inject so you can provide the operation to run. It could be something like:

```java
public class ProvisioningJob implements Job {
   public interface Factory {
      ProvisioningJob create(String group, Supplier<Object> operation);
   }

   private final Predicate<String> waitDataCenterUntilReady;
   private final String group;
   private final Supplier<Object> operation;
   
   @Inject ProvisioningJob(@Named(POLL_PREDICATE_DATACENTER) Predicate<String> waitDataCenterUntilReady,
         @Assisted String group, @Assisted Supplier<Object> operation) {
      this.waitDataCenterUntilReady = waitDataCenterUntilReady;
      this.group = checkNotNull(group, "group cannot be null");
      this.operation = checkNotNull(operation, "operation cannot be null");
   }

   @Override
   public Object call() throws Exception {
      waitDataCenterUntilReady.apply(group);
      Object obj = operation.get();
      waitDataCenterUntilReady.apply(group);
      return obj;
   }
}
```
Then, you should configure it in the ComputeService module *configure* method as follows:
```java
install(new FactoryModuleBuilder().build(ProvisioningJob.Factory.class));
```
And finally, use it in the compute service adapter, by injecting the job factory:
```java
// Inject the job factory
private final ProvisioningJob.Factory jobFactory;
(...)
// Use it to schedule a new job
provisioningManager.provision(jobFactory.create(dataCenterId, new Supplier<Object>() {
   @Override public Object get() {
      return api.serverApi().createServer(serverRequest)
   })
);
```

It's just a suggestion to keep the job creation more self-contained and injectable, and to make the code a bit more concise.

Great job btw! I'll finish the review later today. Meanwhile, feel free to squash the commits and apply any changes you might have.

devcsrj:
Yo @nacx ! You're approach is. simply. too. awesome. +1 

So I refactored the adapter to follow your approach. It worked well on `6d71bbb`. On commit  `199a90f` however, re-running the live test throws a `java.util.concurrent.RejectedExecutionException` ([Stacktrace](http://pastie.org/private/yxap1nx09xrmk3rkeaieg)).

I think I misconfigured the *provider* somehow. Notice anything wrong?

nacx:
I can't see anything wrong in the configuration. Is it consistent that the tests suceed on `6d71bbb` and fail on `199a90f`?

Does it fail if you **don't** add it to the jclouds closer? That would indicate that the context could be being closed while there are still tasks being submitted or in the execution queue. Could that be the case?

nacx:
@devcsrj I've seen a couple of things that are not OK in the Provisioning manager. I've done some minor changes and fixed some stuff, but haven't tried running the live tests. You can see the diff here: https://github.com/nacx/jclouds-labs/compare/199a90fd53526c8a6e7803ea51be157da77154c3...profitbricks

Things I've found:

* Using the `putIfAbsent` to put a new ExecutorService to the map, could leak executor services when the value was absent. They were created but not added.
* Once the manager was closed, one could still enqueue jobs, as the executors are created on demand.
* I've also changed the manager to be properly typed, but I'm not 100% sure that will be guice friendly :) Just discard that part if it does not work.

Could you try running the live tests with my changes?

devcsrj:
Hi @nacx ! Sorry for the delay, I was preoccupied with other things.

- *Does it fail if you don't add it to the jclouds closer?* I tried commenting out the `closer.add(..)` line, but still threw the error.
- *"..not 100% sure that will be guice friendly."* Yeah it wasn't. Too bad. Kinda messy type casts.

But more importantly, because of the `ProvisioningManagerTest`, the exception was reproduced. It was because of the [executors map](https://github.com/devcsrj/jclouds-labs/commit/0e783857c09d94b3770583bb52a23fbee1d894ca#diff-c67d5d74a6059f235c249f2e9a00a176L47) being a static variable. The live tests creates multiple contexts; and afterwards closed. Since it's a static variable, the rest of the instances held a closed executor. Phew, you're a code wizard. \m/

Anyway, before I squash them (and fix conflicts), I'll re-run a live test in a few moments and re-post the latest results. :)

nacx:
>The live tests creates multiple contexts

True, there is an explicit tests in the base compute service test class that creates multiple contexts to make sure there is no shared state between them. Good to see you figured out what was going on. Having unit tests always helps (that's why I consistently ask for a good coverage).

Once the live tests pass, mind squashing the commits and rebasing to the latest master version (there are several conflicts)? I'll then do a final review and merge it.

devcsrj:
Hi @nacx !

The PR's been rebased and squashed. :)
[Live test results](https://gist.github.com/devcsrj/828e5be44a54f5217cbd/b6055d1d6c84848154d0a0f94279121140ace03a) `Tests run: 25, Failures: 0, Errors: 0, Skipped: 0` at `1:01:56.334s`

P.S.: The part left I'm not sure of is my use of [@Named("DEFAULT2")](https://github.com/jclouds/jclouds-labs/commit/97aa7a2f7d268f167403ffe47c3edd5a2a444306#diff-d60e9024736ca1880e1eb4a045763065R128) here. There seems to be a prettier approach.

nacx:
Thanks @devcsrj!

>The part left I'm not sure of is my use of `@Named("DEFAULT2")` here. There seems to be a prettier approach.

Having a look at the method, I don't really get why you need to override it (and I think you shouldn't). The TemplateBuilderSpec already considers the "locationId" property, so there is no need to provide a separate property to configure it to be taken into account to configure the default template (to be used if none is provided when calling the createNodes method).

Also, there is no need to explicitly returning the first Datacenter, as the tempalte builder will fallback to the [default location](https://github.com/jclouds/jclouds/blob/master/compute/src/main/java/org/jclouds/compute/domain/internal/TemplateBuilderImpl.java#L700-L703) if none is provided. That default location supplier is configured by the [ImplicitLocationSupplier](https://github.com/jclouds/jclouds/blob/master/core/src/main/java/org/jclouds/location/suppliers/ImplicitLocationSupplier.java), which default implementation returns the first region or zone. 

The default jclouds behaviour should cover what you are trying to do by overriding this method, if I have understood it properly. Does this make sense?

devcsrj:
The reason for this override is because, when I do:
```
ComputeService compute = ContextBuilder.newBuilder("profitbricks")
    .credentials("identity", "credential")
    .buildView(ComputeServiceContext.class)
    .getComputeService();
Template template = compute.templateBuilder().build();
```
The `template` produced has a location of type `REGION` (`us/lasdev`, `us/las`, etc.,). ProfitBricks provisions a node to a datacenter though (`ZONE`). Then, there's no method like `TemplateBuilder#getLocationId()` to check whether the locationId was already set, so I was *forced* to parse again the property.

Both `BaseTemplateBuilderLiveTest` and `BaseComputeServiceLiveTest` passes with or without this method, but then again I am passing a template, so the context uses the location of the image.

What'ya think?

P.S.: I removed the `ProfitBricksTemplateBuilder#build()`. It turns out, it wasn't needed. Also I added this [catch](https://github.com/devcsrj/jclouds-labs/commit/f4b1311c3cb4f4caf8b17c31e6cb6597d151bce9#diff-11ea89320c87f906e05816c31c5773a1R90), since weirdly enough, it threw `Unparseable date: "2015-06-18T14:16:51+0000"`, even when the actual value returned by the server is `2015-06-18T14:16:51Z`.

nacx:
>The template produced has a location of type REGION (us/lasdev, us/las, etc.,). ProfitBricks provisions a node to a datacenter though (ZONE). Then, there's no method like TemplateBuilder#getLocationId() to check whether the locationId was already set, so I was forced to parse again the property.

Then the right approach is to change the default implicit location supplier. Something like this in the compute service context module should work:

```java
bind(ImplicitLocationSupplier.class).to(OnlyLocationOrFirstZone.class).in(Scopes.SINGLETON);
```
The `OnlyLocationOrFirstZone` already exists in jclouds and will only take into account zone locations.

>P.S.: I removed the ProfitBricksTemplateBuilder#build(). It turns out, it wasn't needed. Also I added this catch, since weirdly enough, it threw Unparseable date: "2015-06-18T14:16:51+0000", even when the actual value returned by the server is 2015-06-18T14:16:51Z.

Perhaps you have more control on how the date is parsed if you inject the `DateService` instead?

devcsrj:
> The OnlyLocationOrFirstZone already exists in jclouds and will only take into account zone locations.

AH. Because of this, I could remove the `ProfitBricksTemplateBuilder` altogether. :+1: 

One last thing though: the [BaseTemplateBuilderLiveTest.testTemplateBuilderWithImageIdSpecified](https://github.com/jclouds/jclouds/blob/master/compute/src/test/java/org/jclouds/compute/internal/BaseTemplateBuilderLiveTest.java#L72) fails, with the result of:

```
Failed tests: 
  ProfitBricksTemplateBuilderLiveTest>BaseTemplateBuilderLiveTest.testTemplateBuilderWithImageIdSpecified:159 expected [{image={id=a380aec9-080b-11e5-b78a-52540066fee9, providerId=a380aec9-080b-11e5-b78a-52540066fee9, name=Ubuntu-12.04-LTS-server-2015-06-01, location={scope=REGION, id=us/lasdev, description=USA Developer cluster, parent=profitbricks}, os={family=ubuntu, version=12.04, description=ubuntu, is64Bit=true}, status=AVAILABLE, loginUser=root}, hardware={id=cpu=1,ram=1024,disk=10.000000, providerId=cpu=1,ram=1024,disk=10.000000, name=cpu=1,ram=1024,disk=10.000000, processors=[{cores=1.0, speed=1.0}], ram=1024, volumes=[{type=LOCAL, size=10.0, bootDevice=true, durable=true}], hypervisor=kvm, supportsImage=ALWAYS_TRUE}, location={scope=ZONE, id=abc433ce-6d21-47c9-a7eb-5342fde68dbf, description=DCtest001, parent=us/lasdev, metadata={version=2, state=AVAILABLE}}, options={}}] but found [{image={id=a380aec9-080b-11e5-b78a-52540066fee9, providerId=a380aec9-080b-11e5-b78a-52540066fee9, name=Ubuntu-12.04-LTS-server-2015-06-01, location={scope=REGION, id=us/lasdev, description=USA Developer cluster, parent=profitbricks}, os={family=ubuntu, version=12.04, description=ubuntu, is64Bit=true}, status=AVAILABLE, loginUser=root}, hardware={id=cpu=1,ram=1024,disk=10.000000, providerId=cpu=1,ram=1024,disk=10.000000, name=cpu=1,ram=1024,disk=10.000000, processors=[{cores=1.0, speed=1.0}], ram=1024, volumes=[{type=LOCAL, size=10.0, bootDevice=true, durable=true}], hypervisor=kvm, supportsImage=ALWAYS_TRUE}, location={scope=REGION, id=us/lasdev, description=USA Developer cluster, parent=profitbricks}, options={}}]
```
The difference was that, the default template used a more specific location (ZONE), while the template with assigned image used the image location instead (REGION). What's the best way to approach this?

nacx:
I think the assertion in that test is not accurate. Given [this](https://github.com/jclouds/jclouds/blob/master/compute/src/main/java/org/jclouds/compute/domain/internal/TemplateBuilderImpl.java#L689-L690), the location of the TemplateBuilder is set to the most concrete location. If the belongs to a location more concrete than the one specified in the ID, the location is changed.

The test does not take that into account, but given how the TemplateBuilder works, it is possible that the selected image has the same location, or a location "contained" in the specified location.

The way to go would be to override that method and replace the assertion by N assertions doing the same, but when asserting the location, check that the location is the same or contained in the given location. Even better, could you do this fix in a PR in the main repo to fix the test in the base class?

devcsrj:
Alright! With the template equals assertion fixed, the `BaseTemplateBuilderLiveTest` passes perfectly. I squashed the commits again, with the latest change that include:
- [Removed](https://github.com/jclouds/jclouds-labs/pull/145/files#diff-d60e9024736ca1880e1eb4a045763065R69) `ProfitBricksTemplateBuilder` altogether - the default behavior was sufficient
- Used `DateService` instead of `DateCodecFactory` - [resolved](https://github.com/jclouds/jclouds-labs/pull/145/files#diff-11ea89320c87f906e05816c31c5773a1R87) the unparseable date issue mentioned yesterday.

P.S.: There's a compilation error in the *jdbc* module. I was wondering if I fix it in this PR, or create a separate one. I chose the latter [PR181](https://github.com/jclouds/jclouds-labs/pull/181).

devcsrj:
rebuild please

profitbricks/src/main/java/org/jclouds/profitbricks/compute/ProfitBricksComputeServiceAdapter.java
```diff
@@ -0,0 +1,480 @@
...
+            logger.trace(">> provisioning complete for storage. returned id='%s'", storageId);
+         } catch (Exception ex) {
+            if (i - 1 == 1) // if first storage (one with image) provisioning fails; stop method
+               throw Throwables.propagate(ex);
+            logger.warn(ex, ">> failed to provision storage. skipping..");
+         }
+
+      Double cores = 0d;
+      for (Processor processor : hardware.getProcessors())
+         cores += processor.getCores();
```
nacx: [minor] Use the [ComputeServiceUtils](https://github.com/jclouds/jclouds/blob/master/compute/src/main/java/org/jclouds/compute/util/ComputeServiceUtils.java#L116-L121) instead.

nacx:
Just one very minor comment, otherwse LGTM.

One important thing to remember: IIRC the default templates provided by ProfitBricks weren't accessible via SSH, right? People have to create an image in their account to make it usable (or at least to make it usable with all the features in the ComputeService) by jclouds. If that is the case, you should add a README file explaining a bit how the provider works, and the steps that need to be done in order to be able to use the provider.

Fantastic job, @devcsrj!

profitbricks/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+```
+
+- Go back to the DCD, and *make a snapshot* of the storage. Put a descriptive name.
+- Configure jclouds to use this *snapshot*.
+
+```java 
+Template template = compute.templateBuilder()
+	.imageNameMatches( "<ideally-unique-snapshot-name>" )
+	.options( new TemplateOptions()
```
nacx: [minor] chante to `compute.templateOptions()`

profitbricks/README.md
```diff
@@ -0,0 +1,64 @@
...
+	
+compute.createNodesInGroup( "cluster1", 1, template );
+```
+> If no `locationId` is specified in the template, jclouds will look for a `DataCenter` that is of same scope as the `Image`.
+
+
+## Limitations
+
+1. There's no direct way of specifying arbitrary number of cores, RAM size, and storage size via the compute interface, at least until after [JCLOUDS-482](https://issues.apache.org/jira/browse/JCLOUDS-482) is resolved. The adapter uses a predefined list hardware profiles instead.
+2. All provisioned nodes are connected at LAN 1 of a datacenter.
```
nacx: TBD in future PRs (not now). The template options allow to specify a set of "network ids". Perhaps that is the place to allow users to specify to which LANs the servers should be attached. The adapter should take into account that option.

profitbricks/README.md
devcsrj: You mean, a separate line like this?
```
compute.templateOptions()
    .overrideLoginUser( "root" )
    .overrideLoginPassword( "password" );

Template template = compute.templateBuilder().imageNameMatches("---").build();
```

profitbricks/README.md
devcsrj: Oh. Is there another provider I could follow for this? I think I saw something like `DockerTemplateOptions` in docker, is that it?

profitbricks/README.md
nacx: No, just to substitute the "new". Don't encourage users to build a new template options object; it should be better retrieved from the ComputeService, as it will provide the provider-specific implementation allowing users to use the `.as(ProfitBricksTempalteOptions.class)` when needed.

profitbricks/README.md
nacx: You can have a look at [openstack-nova](https://github.com/jclouds/jclouds/blob/master/apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/NovaComputeServiceAdapter.java#L119-L121).

The network ids will come as a String and it is up to the adapter to use those values. The id is provider-specific, so you can assume they are already the IDs of the LANs or whatever makes sense int he provider to attach the nodes to those networks.

devcsrj:
Re-squashed commits, with changes including:
- Use [ComputeServiceUtils](https://github.com/jclouds/jclouds-labs/pull/145/files#diff-1a27cebd5920714ae4c207026edd5255R153) .
- Updated README.md. [1](https://github.com/jclouds/jclouds-labs/pull/145/files#diff-60534573ad8a36041d9c4ad20d2927f2R48), [2](https://github.com/jclouds/jclouds-labs/pull/145/files#diff-60534573ad8a36041d9c4ad20d2927f2R62)
- [Use](https://github.com/jclouds/jclouds-labs/pull/145/files#diff-1a27cebd5920714ae4c207026edd5255R145) `#getNetworks` from TemplateOptions; Fallback to default value if invalid.

This PR doesn't have a `ProfitBricksTemplateOptions`, is this required? Or will this do for now (and maybe be updated in a future PR)?

nacx:
>Use #getNetworks from TemplateOptions; Fallback to default value if invalid.

Can a server be attached to more than one network?

nacx:
Pushed to [master](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/17a838b2) and [1.9.x](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/d07fcb17). Thanks for all the effort @devcsrj!

PR #147, opened 2015-03-04
[JCLOUDS-664] Fixing live test execution on Azure

ilgrosso:
This PR is quite large because it was started when PR #135 was not merged yet.

The main purpose is to allow the successful execution of live tests with real Azure subscription, and contains several fixes.


ilgrosso:
Closing in the attempt to provide a more manageable PR complying with JClouds policies.

PR #153, opened 2015-03-13
[JCLOUDS-838] Introducing InMemoryKeyManager for PEM-encoded certificate...

ilgrosso:
... and private key


azurecompute/pom.xml
```diff
@@ -56,6 +56,11 @@
       <version>${project.parent.version}</version>
     </dependency>
     <dependency>
+      <groupId>org.apache.jclouds.driver</groupId>
+      <artifactId>jclouds-sshj</artifactId>
+      <version>${project.parent.version}</version>
```
nacx: We shouldn't force users to a concrete SSH driver. Wouldn't be enough to just add the BouncyCastle driver dependency? Can we get rid of that dependency, anyway?

nacx:
The credentials supplier does many things. I'd refactor it in a "delegating credential supplier" and extract the two branches of the conditional into a "FileBasedSupplier" and an "InMemorySupplier". Inject both suppliers into the delegating one and pick the right one based on the type of the provided credentials.

azurecompute/pom.xml
ilgrosso: Without `jclouds-sshj` I don't have classes like as `org.bouncycastle.openssl.PEMKeyPair`, `org.bouncycastle.openssl.PEMParser` and `org.bouncycastle.openssl.jcajce.JcaPEMKeyConverter` even if I include `jclouds-bouncycastle`.

Does this mean that I need to stop using such classes and revert to manual parsing?

ilgrosso:
@nacx I've carried out the refactoring suggested, please have a look.
FYI at the end I've removed completely the dependency on BouncyCastle.

ilgrosso:
Rebuild, please.

nacx:
Thanks @ilgrosso! Looks like an unrelated build failure. let's trigger the build again: rebuild please!

ilgrosso:
@nacx Last rebuilt was successful :-)

nacx:
Pushed to master and 1.9.x. Thanks!

PR #156, opened 2015-03-17
[JCLOUDS-850] Providing AzureTemplateBuilderLiveTest

ilgrosso:


azurecompute/src/main/java/org/jclouds/azurecompute/compute/functions/OSImageToImage.java
```diff
@@ -81,34 +83,44 @@ private Location createLocation(final String input) {
...
       }
-      return new LocationBuilder().
-              id(input).
-              scope(LocationScope.REGION).
-              description(input).
-              parent(Iterables.getOnlyElement(provider.get())).
-              metadata(ImmutableMap.<String, Object>of("name", input)).
-              build();
+
+      final LocationBuilder builder = new LocationBuilder();
```
nacx: Instead of manually creating a location here, you'd better inject the location supplier into this class and pick one of the existing locations.

azurecompute/src/main/java/org/jclouds/azurecompute/options/AzureComputeTemplateOptions.java
```diff
@@ -89,6 +89,14 @@ public void copyTo(final TemplateOptions to) {
...
             eTo.storageAccountType(storageAccountType.get());
          }
+
+         final LoginCredentials credentials = new LoginCredentials.Builder().
+                 user(loginUser).
+                 password(loginPassword == null || !loginPassword.isPresent() ? null : loginPassword.get()).
+                 privateKey(loginPrivateKey == null || !loginPrivateKey.isPresent() ? null : loginPrivateKey.get()).
+                 authenticateSudo(authenticateSudo == null ? false : authenticateSudo).
+                 build();
+         eTo.overrideLoginCredentials(credentials);
```
nacx: Is this really needed? Doesn't the superclass already copy this?

azurecompute/src/test/java/org/jclouds/azurecompute/compute/AzureTemplateBuilderLiveTest.java
```diff
@@ -0,0 +1,51 @@
...
+
+   @Override
+   protected ProviderMetadata createProviderMetadata() {
+      synchronized (this) {
+         if (providerMeta == null) {
+            providerMeta = new AzureComputeProviderMetadata();
+         }
+      }
+      return providerMeta;
+   }
```
nacx: Is this really necessary? Isn't the base class method working?

ilgrosso: I could not find any way to fetch the provider metadata once created by base class method (and I needed it for `getIso3166Codes()`), so I thought the simpler way was to override and provide a local reference.

BTW This is what I've also seen from [AWSEC2](https://github.com/jclouds/jclouds/blob/acd06b30245c4df27e4e36176916c95df60e42ca/providers/aws-ec2/src/test/java/org/jclouds/aws/ec2/compute/AWSEC2TemplateBuilderLiveTest.java#L65)

azurecompute/src/main/java/org/jclouds/azurecompute/options/AzureComputeTemplateOptions.java
ilgrosso: Without this, I get failure in [BaseTemplateBuilderLiveTest#testFromTemplate](https://github.com/jclouds/jclouds/blob/master/compute/src/test/java/org/jclouds/compute/internal/BaseTemplateBuilderLiveTest.java#L65) because `loginPasswordPresent=true, loginPrivateKeyPresent=true` is not correctly copied.

nacx: Worth fixing the class that does not properly copy that then?

azurecompute/src/test/java/org/jclouds/azurecompute/compute/AzureTemplateBuilderLiveTest.java
nacx: Fair enough. Is the synchronization needed?

ilgrosso: Probably not, I'll re-run all tests without it for further check.

azurecompute/src/main/java/org/jclouds/azurecompute/options/AzureComputeTemplateOptions.java
ilgrosso: After some more investigation, it seems that without the snippet above, the original object shows
```
  loginPassword=Optional.absent()
  loginPrivateKey=Optional.absent()
```
while `eTo`
```
  loginPassword=<null>
  loginPrivateKey=<null>
```

I believe the problem is located in [TemplateOptions around lines 89-96](https://github.com/jclouds/jclouds/blob/master/compute/src/main/java/org/jclouds/compute/options/TemplateOptions.java#L89-L96) which should be better replaced by the snippet above. Alternatively, `loginPassword` and `loginPrivateKey` should be initialized to `Optional.absent()`.

WDYT? Shall I open an issue / PR about this and leave the fix temporarily here (in order to keep the test case above work)?

ilgrosso:
@nacx all comments should now be addressed; still waiting for confirmation on [this one](https://github.com/jclouds/jclouds-labs/pull/156#discussion_r27569236).

azurecompute/src/main/java/org/jclouds/azurecompute/options/AzureComputeTemplateOptions.java
nacx: >WDYT? Shall I open an issue / PR about this and leave the fix temporarily here (in order to keep the test case above working)?

If you can do that, that would be much appreciated. Let's fix jclouds as soon as we find some issues instead of silo-ing fixes in providers.

ilgrosso: As you can see, I have opened [JCLOUDS-876](https://issues.apache.org/jira/browse/JCLOUDS-876) and provided a comment there; for the moment I have also reduced the temporary fix for this issue with last commit 00058df98c7d2e1c9a60ee006885d7610cb7bd3f

nacx:
When going to merge this I had a conflict in the "default template" property I wasn't sure how to properly resolve. Mind rebasing to the latest version (that has #158 and #153 merged)? Meanwhile I'll try to merge JCLOUDS-876 to have a cleaner PR. Thanks!

ilgrosso:
@nacx Sure! I'll try to rebase ASAP, thx for all merges.

ilgrosso:
@nacx PR rebased: all live tests (80, took approximately 2,5 hours) are working.

nacx:
See https://github.com/jclouds/jclouds/pull/724 for a fix for the copy options issue. Can you try this on top of that commit, without the workaround?

ilgrosso:
@nacx I've applied the patch from jclouds/jclouds#724, rebuilt there then removed the workaround from here: everything is working, cool!
Shall I wait for that PR to be merged before removing the workaround here?

nacx:
Squashed and pushed to master and 1.9.x. Many thanks @ilgrosso!

PR #154, opened 2015-03-16
JCLOUDS-853: Improve Create VirtualMachine Deployment

hsbhathiya:
Jclouds-853: Improve Create Virtual Machine Deployment [1]

Create Virtual Machine Deployment [2] Implementation should support different combination that can be used to create a virtual machine deployment.
[1] https://issues.apache.org/jira/browse/JCLOUDS-853
[2] https://msdn.microsoft.com/en-us/library/azure/jj157194.aspx

PR #157, opened 2015-03-25
JCLOUDS-664: Alternate Compute Service Adapter 

hsbhathiya:
Current mapping of Deployment to Node is the assumption that deployment always consist of single roleinststace.

1 - Introduce VirtualMachine in domain 
2 - Map DeploymentToVirtualMachines where single deployment produces list of virtual machines
3 - Map VirtualMachineToNodeMetadata
4 - Change  AzureComputeServiceAdapter<VirtualMachine, RoleSize,OSImage, Location>


azurecompute/src/main/java/org/jclouds/azurecompute/binders/DeploymentParamsToXML.java
```diff
@@ -16,104 +16,188 @@
...
+                        .up(); // Windows ConfigurationSet
+               } else if (roleParam.OSVirtualHardDiskParam().OS() == OSImage.Type.LINUX) {
+                  LinuxConfigurationSetParams linuxParams = roleParam.linuxConfigurationSet();
+                  XMLBuilder configBuilder = configSetsBuilder.e("ConfigurationSet")
+                        .a("i:type", "LinuxProvisioningConfigurationSet"); // Linux
+                  configBuilder.e("ConfigurationSetType").t("LinuxProvisioningConfiguration").up();
+                  add(configBuilder, "HostName", linuxParams.hostName());
+                  add(configBuilder, "UserName", linuxParams.userName());
+                  add(configBuilder, "UserPassword", linuxParams.userPassword());
+                  configBuilder.e("DisableSshPasswordAuthentication").t("false").up()
```
imesh: Wouldn't it better to use constants for defining these strings?
Thanks

azurecompute/src/main/java/org/jclouds/azurecompute/binders/DeploymentParamsToXML.java
```diff
@@ -16,104 +16,188 @@
...
+
+            }
+            roleBuilder.up(); // role
+         }
+         roleListBuilder.up(); // RoleList
+         add(builder, "VirtualNetworkName", params.virtualNetworkName());
+         builder.up();
          // TODO: Undeprecate this method as forcing users to wrap a String in guava's ByteSource is not great.
          return (R) request.toBuilder().payload(builder.asString()).build();
       } catch (Exception e) {
```
imesh: It might be better to add a meaningful message inside the catch block when propagating an exception. Otherwise it would be difficult for users to understand which functionality did not work when an exception is raised.
Thanks

ilgrosso:
Hi @hsbhathiya, could you rebase this?

hsbhathiya:
@ilgrosso 
No had some issues with rebasing https://github.com/jclouds/jclouds-labs/pull/154. Shall we merge those it first and then come to this ?

ilgrosso:
@hsbhathiya fine, let's proceed this way, then. Looking forward to have all your changes onboard!

andrewgaul:
Should we close this given the work in #279?

nacx:
I think hre is little hope for this PR and the ones it depends on to progress, so given that the ARM provider is getting a lot of love and will be ready soon, I think we can close these ones.

PR #158, opened 2015-03-27
[JCLOUDS-849] All tests are green

ilgrosso:


hsbhathiya:
Hi great work @ilgrosso !
I tested against my azure account. Here is the [results](https://gist.github.com/hsbhathiya/aa94e79ae6a61a618c88#file-azure849live)

andreaturli:
Tested on my account, here the results:

Failed tests:
  AzureComputeServiceLiveTest.testOptionToNotBlock:50->BaseComputeServiceLiveTest.testOptionToNotBlock:827 » RunNodes
  AzureComputeServiceLiveTest>BaseComputeServiceLiveTest.testCreateAndRunAService:708->BaseComputeServiceLiveTest.createAndRunAServiceInGroup:725 » RunNodes
  AzureComputeServiceLiveTest>BaseComputeServiceLiveTest.testAScriptExecutionAfterBootWithBasicTemplate:224 » RunNodes
  AzureComputeServiceLiveTest>BaseComputeServiceLiveTest.testConcurrentUseOfComputeServiceToCreateNodes:476 » Execution

Tests run: 55, Failures: 4, Errors: 0, Skipped: 13

ilgrosso:
@andreaturli @hsbhathiya your failures should be due to to a different choice of default location (among our subscription and yours): as you can see from last commit, I have now explicitly set the default test image location to "West Europe": could you please launch the live tests again? Don't forget to clean up completely your subscription environment before.

In case of further error(s), could you please report the full error stacktrace? Thanks.

hsbhathiya:
@ilgrosso 
No test failiures now.
------------------------------------------------------------------------------
Test set: TestSuite
-------------------------------------------------------------------------------
Tests run: 55, Failures: 0, Errors: 0, Skipped: 2, Time elapsed: 5,516.931 sec - in TestSuite

ilgrosso:
@hsbhathiya Good news then! Thanks for feedback.

azurecompute/src/main/java/org/jclouds/azurecompute/compute/strategy/impl/AzureAdaptingComputeServiceStrategies.java
```diff
@@ -0,0 +1,70 @@
...
+      this.client = (AzureComputeServiceAdapter) client;
+      this.nodeMetadataAdapter = nodeMetadataAdapter;
+   }
+
+   @Override
+   public NodeMetadata destroyNode(final String id) {
+      final Deployment node = client.internalDestroyNode(id);
+      return node == null
+              ? null
+              : nodeMetadataAdapter.apply(node);
```
nacx: Out of curiosity, why does this method need to be overridden?

nacx:
Just a minor. lgtm! Thanks @ilgrosso!

azurecompute/src/main/java/org/jclouds/azurecompute/compute/strategy/impl/AzureAdaptingComputeServiceStrategies.java
ilgrosso: The [default implementation of destroyNode()](https://github.com/jclouds/jclouds/blob/master/compute/src/main/java/org/jclouds/compute/strategy/impl/AdaptingComputeServiceStrategies.java#L177) will simply rely on `getNode()` and, if such method returns NULL, not perform any delete.

The `getNode()` logic in Azure needs to be handled in a slightly different way (a "node" is returned only when the CloudService is available and all RoleInstance elements are in a non-transient state) so we implemented the logic for a sort of "getNode for destroy" in the `internalDestroyNode`.

nacx:
Fantastic work done here. Pushed to master and 1.9.x. Many thanks!

PR #162, opened 2015-04-06
Add template option to specify direct (unmapped) ports

grkvlt:
Allows container port to map directly to a specific port on the host for e.g. public web servers on 80 or 443.

```Java
DockerTemplateOptions options = ...
options.portBindings(ImmutableMap.of(80, 80, 443, 443));
```

### DIFF
--- a/docker/src/main/java/org/jclouds/docker/compute/options/DockerTemplateOptions.java
+++ b/docker/src/main/java/org/jclouds/docker/compute/options/DockerTemplateOptions.java
@@ -54,6 +54,7 @@ public class DockerTemplateOptions extends TemplateOptions implements Cloneable 
    protected Optional<List<String>> commands = Optional.absent();
    protected Optional<Map<String, String>> volumes = Optional.absent();
    protected Optional<List<String>> env = Optional.absent();
+   protected Optional<List<Integer>> directPorts = Optional.absent();
 
    @Override
    public DockerTemplateOptions clone() {
@@ -84,10 +85,13 @@ public class DockerTemplateOptions extends TemplateOptions implements Cloneable 
          }
          if (cpuShares.isPresent()) {
              eTo.cpuShares(cpuShares.get());
-          }
+         }
          if (env.isPresent()) {
              eTo.env(env.get());
-          }
+         }
+         if (directPorts.isPresent()) {
+             eTo.directPorts(directPorts.get());
+         }
       }
    }
 
@@ -104,12 +108,13 @@ public class DockerTemplateOptions extends TemplateOptions implements Cloneable 
               equal(this.memory, that.memory) &&
               equal(this.commands, that.commands) &&
               equal(this.cpuShares, that.cpuShares) &&
-              equal(this.env, that.env);
+              equal(this.env, that.env) &&
+              equal(this.directPorts, that.directPorts);
    }
 
    @Override
    public int hashCode() {
-      return Objects.hashCode(super.hashCode(), volumes, hostname, dns, memory, commands, cpuShares, env);
+      return Objects.hashCode(super.hashCode(), volumes, hostname, dns, memory, commands, cpuShares, env, directPorts);
    }
 
    @Override
@@ -122,6 +127,7 @@ public class DockerTemplateOptions extends TemplateOptions implements Cloneable 
               .add("commands", commands)
               .add("volumes", volumes)
               .add("env", env)
+              .add("directPorts", directPorts)
               .toString();
    }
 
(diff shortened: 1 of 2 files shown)


grkvlt:
@andreaturli @nacx coukld you please give this a quick review?

docker/src/main/java/org/jclouds/docker/compute/strategy/DockerComputeServiceAdapter.java
```diff
@@ -136,6 +140,14 @@ public DockerComputeServiceAdapter(DockerApi api) {
               .publishAllPorts(true)
               .privileged(true);
 
+      if (templateOptions.getDirectPorts().isPresent()) {
```
andreaturli: Maybe it is better to use the docker naming `portBindings` instead of `directPorts`

docker/src/main/java/org/jclouds/docker/compute/strategy/DockerComputeServiceAdapter.java
```diff
@@ -136,6 +140,14 @@ public DockerComputeServiceAdapter(DockerApi api) {
...
+         Map<String, List<Map<String, String>>> portBindings = Maps.newHashMap();
+         for (Integer port : templateOptions.getDirectPorts().get()) {
+            portBindings.put(port + "/tcp", Lists.<Map<String, String>>newArrayList(ImmutableMap.of("HostPort", Integer.toString(port))));
```
andreaturli: Shouldn't check if this `port` has been already specified as `exposedPort` so basically if it is part of `inboundPorts`?

grkvlt: It may not have been specified there, though. For example, a Dockerfile will have `EXPOSE` commands, but jclouds does not need to know about them, Docker will find out when it starts the container...

docker/src/main/java/org/jclouds/docker/compute/strategy/DockerComputeServiceAdapter.java
grkvlt: Agree. Could have a `Map<Integer, Integer>` type here, instead? This would be mappings of host port to container port (this is the order  Docker uses with `--publish`.)

grkvlt:
@andreaturli is this better?

docker/src/main/java/org/jclouds/docker/compute/options/DockerTemplateOptions.java
```diff
@@ -122,13 +127,18 @@ public String toString() {
...
    }
 
    public static final DockerTemplateOptions NONE = new DockerTemplateOptions();
 
    public DockerTemplateOptions volumes(Map<String, String> volumes) {
-      this.volumes = Optional.<Map<String, String>> of(ImmutableMap.copyOf(volumes));
+      for (Map.Entry<String, String> entry : checkNotNull(volumes, "volumes").entrySet()) {
+         checkNotNull(entry.getKey(), "host paths must not be null");
+         checkNotNull(entry.getValue(), "container paths must not be null");
+      }
```
nacx: In general for the whole PR, are these checks really needed? Guava's immutable collections don't allow null values so checking the null input in the map and list options here looks redundant?

andreaturli:
Lgtm, very minor comment.

docker/src/main/java/org/jclouds/docker/compute/options/DockerTemplateOptions.java
grkvlt: I couldn't remember. So, do you think we get a useful enogh error when we create an `ImmutableMap` from a `HashMap` with nulls in it? If so, we just need the `checkNotNull(input)` type validation, and I'll update the PR appropriately. _**Checked** and it does a `checkEntryNotNull` so we're all good_

grkvlt:
Cheers for the quick response @andreaturli @nacx - I have made the changes you suggested.

andreaturli:
@grkvlt could you fix the checkstyle violation at 

```DockerComputeServiceAdapter.java[22:8] (imports) UnusedImports: Unused import - java.util.Arrays.```

so to make the builder happier?

grkvlt:
@andreaturli ok build seems happy now :frog:

andreaturli:
thanks @grkvlt, could you please squash the commits so that I can merge it? thanks!

grkvlt:
@andreaturli there are now just two commits, the original and the squashed review changes, which I think preserves enough context without cluttering up the commit history. Or would you prefer squashed to a single commit?

andreaturli:
I think 2 commits is ok. @demobox @nacx, thoughts ?

andreaturli:
rebuild please

grkvlt:
@andreaturli what do you mean?

andreaturli:
@grkvlt the previous build failed because of an external error, so I asked for a rebuild in that way.
Merging the PR just now

grkvlt:
@andreaturli np, thanks!

andreaturli:
Merged at [8d1404038dfe952b7fc](https://git1-us-west.apache.org/repos/asf?p=jclouds-labs.git;a=commit;h=8d140403)

andreaturli:
Also backported to 1.9.x at http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/38b09ced

PR #164, opened 2015-04-10
DockerTemplateOptions: use the correct return type

andreaturli:
@grkvlt and @neykov can you give a quick review?

### DIFF
--- a/docker/src/main/java/org/jclouds/docker/compute/options/DockerTemplateOptions.java
+++ b/docker/src/main/java/org/jclouds/docker/compute/options/DockerTemplateOptions.java
@@ -16,20 +16,20 @@
  */
 package org.jclouds.docker.compute.options;
 
-import com.google.common.base.Objects;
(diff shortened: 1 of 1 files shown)


grkvlt:
Looks fine. Note that you could also at the same time remove the casting from the `Builder` static methods, as it isn't needed. So the static method for `env` becomes:

```Java
/**
 * @see DockerTemplateOptions#env(Iterable)
 */
public static DockerTemplateOptions env(Iterable<String> env) {
   DockerTemplateOptions options = new DockerTemplateOptions();
   return options.env(env);
}
```

And the same for all the other options that are defined in this class.

neykov:
+1

andreaturli:
Thanks @grkvlt for your tip ;)
Merging as soon as the builder is green

andreaturli:
Merged at http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/eae89101
and backported to 1.9.x

PR #165, opened 2015-04-11
JCLOUDS-882: Accounts API implementation

cdancy:
Initial accounts implementation. Mock tests and Live tests have both been run locally and verified working.


cdancy:
@nacx let's please merge PR 166 before this as the Accounts API has a minor dependency on the Roles API. Both are fairly simple and straight forward so I don't imagine too much churn.

shipyard/src/main/java/org/jclouds/shipyard/domain/accounts/CreateAccount.java
```diff
@@ -0,0 +1,45 @@
...
+   public abstract CreateRole role();
+   
+   CreateAccount() {
+   }
+
+   @SerializedNames({ "username", "password", "role"})
+   public static CreateAccount create(String username, String password, CreateRole role) {
+      checkArgument(username.length() > 0, "username must not be empty string");
+      checkArgument(password.length() > 0, "password must not be empty string");
+      if (role == null) role = CreateRole.create("user");
```
nacx: Instead of doing this I'dd add two factory methods, one with the role and one without it.

shipyard/src/test/java/org/jclouds/shipyard/features/AccountsApiLiveTest.java
```diff
@@ -0,0 +1,70 @@
...
+@Test(groups = "live", testName = "AccountsApiLiveTest", singleThreaded = true)
+public class AccountsApiLiveTest extends BaseShipyardApiLiveTest {
+
+   private final String username = UUID.randomUUID().toString().replaceAll("-", "");
+   private final String password = UUID.randomUUID().toString().replaceAll("-", "");
+
+   @AfterClass (alwaysRun = true)
+   protected void tearDown() {
+      boolean removed = api().deleteAccount(username);
+      assertTrue(removed);
```
nacx: This will fail if the test that creates the account fails.

shipyard/src/test/java/org/jclouds/shipyard/features/AccountsApiLiveTest.java
```diff
@@ -0,0 +1,70 @@
...
+   public void testListAccounts() throws Exception {
+      List<AccountInfo> possibleAccounts = api().listAccounts();
+      assertNotNull(possibleAccounts, "possibleAccounts was not set");
+      assertTrue(possibleAccounts.size() > 0, "Expected at least 1 Account but list was empty");
+      boolean accountFound = false;
+      for (AccountInfo possibleAccount : possibleAccounts) {
+         if (possibleAccount.username().equals(username)) {
+            accountFound = true;
+         }
+      }
```
nacx: Use a more idiomatic Iterables.find or similar approach.

nacx:
Thanks @cdancy!

cdancy:
@nacx comments have been addressed and things look good on my end. Now that JCLOUDS-886 has been merged we are good to go here. This is the last piece of work I wanted to get in before tackling the compute API.

shipyard/src/main/java/org/jclouds/shipyard/domain/accounts/CreateAccount.java
```diff
@@ -0,0 +1,50 @@
...
+
+   @SerializedNames({ "username", "password", "role"})
+   public static CreateAccount create(String username, String password, RoleInfo role) {
+      checkArgument(username.length() > 0, "username must not be empty string");
+      checkArgument(password.length() > 0, "password must not be empty string");
+      if (role == null) role = RoleInfo.create("user", null);
+      return new AutoValue_CreateAccount(username, password, role);
+   }
+   
+   @SerializedNames({ "username", "password", "role"})
```
nacx: Remove the `role` fro the annotation.

nacx:
Just one very minor. Mind addressing it and squash the commits to we can cleanly merge this? Thanks @cdancy!

shipyard/src/main/java/org/jclouds/shipyard/domain/accounts/CreateAccount.java
cdancy: That appears to be required which is why I left it in ;) Removing it causes jclouds to complain about it not being there

nacx: Oh, you're right. It might fail [here](https://github.com/jclouds/jclouds/blob/master/core/src/main/java/org/jclouds/json/internal/NamingStrategies.java#L177-193).

jclouds uses that annotation to get the serialized names for the fields, and matches the declared fields in the class with the annotation values, expecting all fields to have the corresponding value. As having the extra "role" parameter in the annotation looks confusing, can you try just removing the annotation from this method, and leave it only in the factory method that accepts all the parameters? This would be more close to the approach we follow with the constructors, where we only annotate the constructor that has all values.

I haven't tried it with the factory method approach, but looking at the code I think it should work.

shipyard/src/main/java/org/jclouds/shipyard/domain/accounts/CreateAccount.java
```diff
@@ -0,0 +1,50 @@
...
+   public static CreateAccount create(String username, String password, RoleInfo role) {
+      checkArgument(username.length() > 0, "username must not be empty string");
+      checkArgument(password.length() > 0, "password must not be empty string");
+      if (role == null) role = RoleInfo.create("user", null);
+      return new AutoValue_CreateAccount(username, password, role);
+   }
+   
+   @SerializedNames({ "username", "password" })
+   public static CreateAccount create(String username, String password) {
+      return CreateAccount.create(username, password, null);
```
cdancy: @nacx this fails going through cloudbees but succeeds locally for me. Not exactly sure why...

nacx: In my last comment I suggested to try removing the annotation from this method, leaving only one method with the `@SerializedNames` annotation in the class. Could you try that?

cdancy:
@nacx we can remove this PR. Project is more/less dead and I've just submitted a PR to remove it from jclouds-labs.

nacx:
Ok @cdancy. Thanks for all the good work you've put in this provider!

PR #170, opened 2015-05-04
Add option to remove a volume, when removing docker container

jbartece:
It would be great, if you could backport it to 1.9.x branch ;-)

### DIFF
--- a/docker/src/main/java/org/jclouds/docker/options/RemoveContainerOptions.java
+++ b/docker/src/main/java/org/jclouds/docker/options/RemoveContainerOptions.java
@@ -29,6 +29,17 @@ public class RemoveContainerOptions extends BaseHttpRequestOptions {
       this.queryParameters.put("force", force.toString());
       return this;
    }
+   
+   /**
+    * Remove the volumes associated to the container
+    * 
+    * @param volume If set to true the volume associated to the container will be removed. 
+    * Otherwise it will not be removed.
+    */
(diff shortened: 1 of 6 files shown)


docker/src/main/java/org/jclouds/docker/options/RemoveContainerOptions.java
```diff
@@ -29,6 +29,17 @@ public RemoveContainerOptions force(Boolean force) {
...
       return this;
    }
+   
+   /**
+    * Remove the volumes associated to the container
+    * 
+    * @param volume If set to true the volume associated to the container will be removed. 
+    * Otherwise it will not be removed.
+    */
+   public RemoveContainerOptions volume(Boolean volume) {
```
nacx: Change the parameter to be a primitive as it does not make sense to pass a `null` value here. It could throw a NPE in the next line.

Could you also change the existing methods to use a `boolean` primitive too?

nacx:
Just one minor comment. Thanks @jbartece!

docker/src/main/java/org/jclouds/docker/options/RemoveContainerOptions.java
jbartece: OK. I'll change it

nacx:
Pushed to master and 1.9.x. Thanks @jbartece!

PR #172, opened 2015-05-06
[JCLOUDS-898] better management of HTTP connections

andreaturli:


docker/src/main/java/org/jclouds/docker/suppliers/SSLContextWithKeysSupplier.java
```diff
@@ -68,21 +67,20 @@
...
-         throw propagate(e);
+      if (new File(currentCreds.identity).isFile() && new File(currentCreds.credential).isFile()) {
+         try {
+            SSLContext sslContext = SSLContext.getInstance("TLS");
+            X509Certificate certificate = getCertificate(loadFile(currentCreds.identity));
+            PrivateKey privateKey = getKey(loadFile(currentCreds.credential));
+            sslContext.init(new KeyManager[]{new InMemoryKeyManager(certificate, privateKey)}, trustManager, new SecureRandom());
+            return sslContext;
+         } catch (GeneralSecurityException e) {
+            throw new AssertionError(); // The system has no TLS. Just give up.
```
nacx: propagate a proper exception here.

docker/src/main/java/org/jclouds/docker/suppliers/SSLContextWithKeysSupplier.java
```diff
@@ -68,21 +67,20 @@
...
+         } catch (IOException e) {
+            throw propagate(e);
+         }
       }
+      return null; // identity and/or credentials are not files, can't setup sslContext
```
nacx: Don't return null. Suppliers should not return null. Instead, move the is file checks to the calling class, and only call the supplier if needed.

nacx:
Up to date PR: https://github.com/jclouds/jclouds-labs/pull/198

PR #173, opened 2015-05-07
JCLOUDS-826: Add jdbc api skeleton

rcoedo:
First commit for the new jdbc api. Contains the skeleton and the container operations.


jdbc/src/main/java/org/jclouds/jdbc/strategy/JdbcStorageStrategy.java
```diff
@@ -0,0 +1,191 @@
...
+   }
+
+   @Override
+   public ContainerAccess getContainerAccess(String name) {
+      return ContainerAccess.PRIVATE;
+   }
+
+   @Override
+   public void setContainerAccess(String s, ContainerAccess containerAccess) {
+      // Unsupported
```
zack-shoylev: ?

rcoedo: Since the database can't make public rows, I'm returning ContainerAccess.PRIVATE for all the containers and never changing it. Maybe I'm missing something here?

jdbc/.gitignore
```diff
@@ -0,0 +1 @@
+Makefile
```
nacx: Add this to the `.gitignore` file in the root of the repo and remove this one.

jdbc/pom.xml
```diff
@@ -0,0 +1,155 @@
...
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.apache.jclouds</groupId>
+    <artifactId>jclouds-project</artifactId>
```
nacx: If you reference the pom in the root of the repo, You have to use [these Maven coordinates](https://github.com/jclouds/jclouds-labs/blob/master/pom.xml#L27-L28).

jdbc/pom.xml
```diff
@@ -0,0 +1,155 @@
...
+    <version>2.0.0-SNAPSHOT</version>
+    <relativePath>../pom.xml</relativePath>
+  </parent>
+  <groupId>org.apache.jclouds.api</groupId>
```
nacx: Change to `org.apache.jclouds.labs`.

jdbc/pom.xml
```diff
@@ -0,0 +1,155 @@
...
+  <artifactId>jdbc</artifactId>
+  <name>jclouds jdbc core</name>
+  <description>jclouds components to access jdbc</description>
+  <packaging>bundle</packaging>
+
+  <properties>
+    <!-- This api has been written in a manner which requires Java language level 7. -->
+    <maven.compile.source>1.7</maven.compile.source>
+    <maven.compile.target>1.7</maven.compile.target>
```
nacx: jclouds should run on Java 6. Try ti stick to that version.

jdbc/pom.xml
```diff
@@ -0,0 +1,155 @@
...
+    </dependency>
+    <dependency>
+      <groupId>com.google.auto.service</groupId>
+      <artifactId>auto-service</artifactId>
+      <optional>true</optional>
+    </dependency>
+    <dependency>
+      <groupId>com.google.auto.value</groupId>
+      <artifactId>auto-value</artifactId>
+      <version>1.0</version>
```
nacx: You should be able to remove this version as it is already configured in the jclouds parent pom.xml

jdbc/pom.xml
```diff
@@ -0,0 +1,155 @@
...
+    <dependency>
+      <groupId>org.apache.jclouds.driver</groupId>
+      <artifactId>jclouds-log4j</artifactId>
+      <version>${project.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.google.auto.service</groupId>
+      <artifactId>auto-service</artifactId>
+      <optional>true</optional>
```
nacx: Remove this and configure in `provided` scope.

jdbc/pom.xml
nacx: Configure this dependency in `provided` scope.

jdbc/pom.xml
```diff
@@ -0,0 +1,155 @@
...
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>animal-sniffer-maven-plugin</artifactId>
+        <configuration>
+          <signature>
+            <groupId>org.codehaus.mojo.signature</groupId>
+            <artifactId>java17</artifactId>
+            <version>1.0</version>
+          </signature>
+        </configuration>
+      </plugin>
```
nacx: You should make it work with Java 6, so you should be able to remove this too.

jdbc/src/main/java/org/jclouds/jdbc/JdbcApiMetadata.java
```diff
@@ -0,0 +1,73 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jclouds.jdbc;
+
+import com.google.auto.service.AutoService;
+import com.google.common.collect.ImmutableSet;
+import com.google.inject.Module;
+import org.jclouds.apis.ApiMetadata;
```
nacx: This import is only used because the javadoc references it. Remove it and use the fully qualified name directly in the javadoc, to avoid unused import warnings.

jdbc/src/main/java/org/jclouds/jdbc/JdbcApiMetadata.java
```diff
@@ -0,0 +1,73 @@
...
+   protected JdbcApiMetadata(Builder builder) {
+      super(builder);
+   }
+
+   public static class Builder extends BaseApiMetadata.Builder<Builder> {
+
+      protected Builder() {
+         id("jdbc")
+         .name("Jdbc BlobStore")
+         .identityName("Unused")
```
nacx: Also add the `credentialName`. Will both be unused? Or will they be configured to the DB user and pass?

jdbc/src/main/java/org/jclouds/jdbc/JdbcApiMetadata.java
```diff
@@ -0,0 +1,73 @@
...
+         .defaultEndpoint("http://localhost/transient")
+         .defaultIdentity("unused")
+         .defaultCredential("unused")
```
nacx: Don't declare this. Let's better keep them `absent()`, which is the default instead of having an invalid value.

jdbc/src/main/java/org/jclouds/jdbc/config/JPAInitializer.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+package org.jclouds.jdbc.config;
+
+import com.google.inject.Inject;
+import com.google.inject.persist.PersistService;
+
+public class JPAInitializer {
+
+   @Inject
+   public JPAInitializer(PersistService service) {
```
nacx: Injection constructors should be package private. Remove the `public` modifier.

jdbc/src/main/java/org/jclouds/jdbc/repository/GenericRepository.java
```diff
@@ -0,0 +1,55 @@
...
+import javax.persistence.EntityManager;
+import java.io.Serializable;
+import java.lang.reflect.ParameterizedType;
+
+public abstract class GenericRepository<T, PK extends Serializable> {
+
+   protected Class<T> entityClass;
+
+   @Inject
+   protected Provider<EntityManager> entityManager;
```
nacx: Prefer constructor injection, to make the constructor package private and this field final.

jdbc/src/main/java/org/jclouds/jdbc/repository/GenericRepository.java
```diff
@@ -0,0 +1,55 @@
...
+import com.google.inject.Inject;
+import com.google.inject.Provider;
+
+import javax.persistence.EntityManager;
+import java.io.Serializable;
+import java.lang.reflect.ParameterizedType;
+
+public abstract class GenericRepository<T, PK extends Serializable> {
+
+   protected Class<T> entityClass;
```
nacx: Make this field final.

jdbc/src/main/java/org/jclouds/jdbc/strategy/JdbcStorageStrategy.java
```diff
@@ -0,0 +1,191 @@
...
+   @Resource
+   protected Logger logger = Logger.NULL;
+
+   protected final Provider<BlobBuilder> blobBuilders;
+   protected final ContainerRepository containerRepository;
+   protected final JdbcContainerNameValidator jdbcContainerNameValidator;
+   protected final JdbcBlobKeyValidator jdbcBlobKeyValidator;
+
+   @Inject
+   protected JdbcStorageStrategy(Provider<BlobBuilder> blobBuilders,
```
nacx: Remove the `protected` modifier, so you know the constructor is only called by the injector. Then you can safely remove all redundant `checkNotNull` checks.

jdbc/src/main/java/org/jclouds/jdbc/strategy/JdbcStorageStrategy.java
```diff
@@ -0,0 +1,191 @@
...
+   }
+
+   @Override
+   @Transactional
+   public boolean createContainerInLocation(String name, Location location,
+         CreateContainerOptions createContainerOptions) {
+      logger.debug("Creating container %s", name);
+      jdbcContainerNameValidator.validate(name);
+      containerRepository.create(Container.builder().name(name).build());
+      return true;
```
nacx: Does this method always return `true`? Will it throw an exception if it fails? Why returning a boolean then and not just `void`?

jdbc/src/main/java/org/jclouds/jdbc/strategy/JdbcStorageStrategy.java
nacx: Would it be better to fail? Or fail if the `containerAccess` parameter is other than private?

jdbc/src/main/java/org/jclouds/jdbc/util/JdbcBlobUtils.java
```diff
@@ -0,0 +1,72 @@
...
+ * Implements the {@link BlobUtils} interfaced and act as a bridge to
+ * {@link LocalStorageStrategy} when used inside {@link BlobStore}
+ */
+public class JdbcBlobUtils implements BlobUtils {
+
+   protected final JdbcStorageStrategy storageStrategy;
+   protected final Provider<BlobBuilder> blobBuilders;
+
+   @Inject
+   public JdbcBlobUtils(LocalStorageStrategy storageStrategy, Provider<BlobBuilder> blobBuilders) {
```
nacx: Same here, remove the public modifier and the null checks.

jdbc/src/test/java/org/jclouds/jdbc/JdbcBlobStoreTest.java
```diff
@@ -0,0 +1,72 @@
...
+import org.testng.annotations.AfterMethod;
+import org.testng.annotations.BeforeMethod;
+import org.testng.annotations.Test;
+
+import java.io.IOException;
+import java.util.Set;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@Test(groups = "unit", testName = "JdbcBlobStoreTest", singleThreaded = true)
```
nacx: Does this test really need to be single threaded?

jdbc/src/test/java/org/jclouds/jdbc/strategy/JdbcStorageStrategyTest.java
```diff
@@ -0,0 +1,75 @@
...
+import org.jclouds.blobstore.domain.StorageType;
+import org.jclouds.jdbc.module.TestContextModule;
+import org.testng.annotations.BeforeMethod;
+import org.testng.annotations.Test;
+
+import java.util.Date;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@Test(groups = "unit", testName = "JdbcStorageStrategyTest", singleThreaded = true)
```
nacx: Same here, does this really need to be single threaded?

jdbc/pom.xml
```diff
@@ -0,0 +1,155 @@
...
+    </dependency>
+    <dependency>
+      <groupId>com.google.inject.extensions</groupId>
+      <artifactId>guice-persist</artifactId>
+      <version>4.0</version>
+    </dependency>
+    <dependency>
+      <groupId>org.hibernate.javax.persistence</groupId>
+      <artifactId>hibernate-jpa-2.1-api</artifactId>
+      <version>1.0.0.Final</version>
```
nacx: An hibernate dependency should be only in test scope, right? We just need to import the standard JPA interface, but let users pick their preferred impl.

jdbc/pom.xml
nacx: Looking at the code, you are not using Google AutoValue, as it does not seem to be easy to use that model with JPA. Remove the dependency if it is the case.

jdbc/src/main/java/org/jclouds/jdbc/strategy/JdbcStorageStrategy.java
zack-shoylev: The permissions can be part of the database record. You don't even need posix level permissions, I think you mostly need owner and private/public.

rcoedo: What difference does it make if a container is marked on the database as private/public? I can store it in the database, but it seems pretty useless to me since the container can't be shared to other database users this way like, let's say, a S3 bucket. Again, probably I am missing something :)

jdbc/src/main/java/org/jclouds/jdbc/JdbcApiMetadata.java
rcoedo: This one is used in AutoService(ApiMetadata.class) too, not just for the javadoc.

jdbc/src/main/java/org/jclouds/jdbc/JdbcApiMetadata.java
rcoedo: The datasource config belongs in a persistence.xml file. The persistence unit must be passed as a JpaPersistModule to the blobstore.

jdbc/src/main/java/org/jclouds/jdbc/JdbcApiMetadata.java
rcoedo: I get an error if the identity is not present in properties.

jdbc/src/main/java/org/jclouds/jdbc/strategy/JdbcStorageStrategy.java
rcoedo: I can return false when a persistence exception raises, but I will have to make a service layer to handle the transactions then. I'll fix this.

jdbc/src/test/java/org/jclouds/jdbc/JdbcBlobStoreTest.java
rcoedo: It does, the methods are running against the same database.

jdbc/src/test/java/org/jclouds/jdbc/strategy/JdbcStorageStrategyTest.java
rcoedo: Same as above.

jdbc/src/main/java/org/jclouds/jdbc/strategy/JdbcStorageStrategy.java
zack-shoylev: It will allow us to run the public/private blobstore integration tests.

jdbc/pom.xml
rcoedo: The 2.1 javax.persistence api is part of Java EE. It is available as a standalone package but the last version is 1.0.2. Hibernate and other providers offer a 2.1 api and the implementation as two different packages. I added the hibernate jpa api as a dependency, The entity manager (the implementation) is test scoped.

Thoughts?

jdbc/src/main/java/org/jclouds/jdbc/strategy/JdbcStorageStrategy.java
rcoedo: :+1: I'm on it.

zack-shoylev:
merged

jdbc/pom.xml
nacx: +1 :) Thanks for explaining!

PR #176, opened 2015-05-26
JCLOUDS-826: Add blob operations

rcoedo:


zack-shoylev:
Rebuild please

zack-shoylev:
Actually it seems like a compilation failure

rcoedo:
Seems like I imported a java 8 package instead of the guava one. I fixed it and changed the IDE jdk now.

rcoedo:
Now it's failing due to this `Found duplicate classes/resources in test classpath.`, but I need the duplicated classes for testing. Any idea on how could I solve it?

rcoedo:
Rebuild please

jdbc/src/main/java/org/jclouds/jdbc/entity/PayloadEntity.java
```diff
@@ -0,0 +1,198 @@
...
+import javax.persistence.Entity;
+import javax.persistence.FetchType;
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+import java.util.Date;
+import java.util.List;
+
+@Entity
+public class PayloadEntity {
```
zack-shoylev: This might not work, but can we use AutoValue 1.1 for classes such as this one?

rcoedo: Is it possible? These are entities, not values. I think the ORM needs them to be mutable.

jdbc/src/test/resources/META-INF/persistence.xml
```diff
@@ -3,21 +3,47 @@
...
     <properties>
+      <property name="hibernate.connection.useUnicode" value="true" />
+      <property name="hibernate.connection.characterEncoding"
+                value="utf8" />
+      <property name="hibernate.connection.charSet" value="utf8" />
       <property name="hibernate.archive.autodetection" value="class" />
       <property name="hibernate.dialect" value="org.hibernate.dialect.HSQLDialect" />
       <property name="hibernate.connection.driver_class" value="org.hsqldb.jdbcDriver" />
-      <property name="hibernate.connection.url" value="jdbc:hsqldb:file:target/testdb;shutdown=true" />
+      <property name="hibernate.connection.url" value="jdbc:hsqldb:file:target/testdb;shutdown=true;sql.enforce_strict_size=true;hsqldb.tx=mvcc" />
```
zack-shoylev: Add some documentation about why some of these are set as they are here.

jdbc/src/test/java/org/jclouds/jdbc/BaseJdbcBlobStoreTest.java
```diff
@@ -0,0 +1,746 @@
...
+   @Test
+   public void testDeleteContainer() {
+      assertThat(blobStore.createContainerInLocation(null, CONTAINER_NAME)).isTrue();
+      assertThat(blobStore.containerExists(CONTAINER_NAME)).isTrue();
+      blobStore.deleteContainer(CONTAINER_NAME);
+      assertThat(blobStore.containerExists(CONTAINER_NAME)).isFalse();
+   }
+
+   @Test
+   public void testPutBlob() {
```
zack-shoylev: Is chunked data tested someplace else?

jdbc/src/main/java/org/jclouds/jdbc/strategy/JdbcStorageStrategy.java
```diff
@@ -123,63 +121,82 @@ public void setContainerAccess(String container, ContainerAccess containerAccess
...
+         metadata.setLocation(null);
+         metadata.setCreationDate(containerEntity.getCreationDate());
+      }
       return metadata;
    }
 
    @Override
-   public boolean blobExists(String s, String s1) {
-      return false;
+   public boolean blobExists(String container, String key) {
```
zack-shoylev: At least the parameters in the strategy methods should probably have javadocs, even if the methods are mostly self-documented. Other stuff that might be useful is null-safety or invalid characters, if any.
The strategy itself probably needs some javadocs on how stuff is stored.

jdbc/src/main/java/org/jclouds/jdbc/entity/PayloadEntity.java
zack-shoylev: It might make sense to doc this, if so.

jdbc/src/test/java/org/jclouds/jdbc/BaseJdbcBlobStoreTest.java
rcoedo: Yes, there are many putBlob/getBlob tests in the BaseJdbcStorageStrategyTest and BaseJdbcBlobStoreTest classes.

jdbc/src/main/java/org/jclouds/jdbc/entity/ContainerEntity.java
```diff
@@ -96,6 +96,30 @@ public static Builder builder() {
...
+      ContainerEntity containerEntity = (ContainerEntity) o;
+
+      return containerAccess == containerEntity.containerAccess && !(creationDate != null ?
+            !creationDate.equals(containerEntity.creationDate) :
+            containerEntity.creationDate != null) && !(id != null ? !id.equals(containerEntity.id) : containerEntity.id != null) && !(
+            name != null ? !name.equals(containerEntity.name) : containerEntity.name != null);
+   }
+
+   @Override
+   public int hashCode() {
```
zack-shoylev: Use guava helpers for hashcode/equals/tostring, to reduce code:
https://code.google.com/p/guava-libraries/wiki/CommonObjectUtilitiesExplained

jdbc/src/main/java/org/jclouds/jdbc/service/JdbcService.java
```diff
@@ -16,54 +16,243 @@
...
+      }
+      BlobEntity blobEntity = blobToBlobEntity.apply(blob);
+      blobEntity.getPayload().setChunks(storeData(blob.getPayload().openStream()));
+      blobEntity.setContainerEntity(containerRepository.findContainerByName(containerName));
+      blobEntity.setKey(key);
+      blobEntity.setBlobAccess(blobAccess);
+      blobEntity.setCreationDate(creationDate);
+
+
+      HashCode hash = ByteStreams2.hashAndClose(blob.getPayload().openStream(), Hashing.md5());
```
zack-shoylev: Somewhat of a small problem with this.
In a lot of cases, the user has already checked the hash of the stream they are trying to upload to a cloud/jdbc/etc. Which means that by having us to it again, performance is reduced overall. That is really the point of returning the hash.
I am mostly OK with keeping it this way, but maybe explicitly document that the check happens here too? Either that or remove the integrity check. Removing it might be better. WDYT?

zack-shoylev:
Ok, this is a gigantic PR but seems like it might be ready to merge. @nacx do you want to give this a parse?
Otherwise, let's have a look at those last few issues I mentioned, then squash, and I will merge it.

jdbc/src/main/java/org/jclouds/jdbc/repository/ContainerRepository.java
```diff
@@ -19,32 +19,40 @@
...
-      return entityManager.get().createQuery("SELECT c FROM " + entityClass.getName() + " c WHERE c.name = :name", entityClass)
-            .setParameter("name", name)
-            .getSingleResult();
+   public ContainerEntity findContainerByName(String name) {
+      try {
+         return entityManager.get().createQuery("SELECT c FROM " + entityClass.getName() + " c WHERE c.name = :name", entityClass)
+               .setParameter("name", name)
+               .getSingleResult();
+      } catch (PersistenceException e) {
+         return null;
```
nacx: You'd better just catch the `NoResultException` to return null and propagate the exception otherwise.

jdbc/src/main/java/org/jclouds/jdbc/service/JdbcService.java
```diff
@@ -16,54 +16,243 @@
...
+   @Transactional
+   public void deleteBlobsByDirectory(String containerName, String directoryName, boolean recursive) {
+      List<BlobEntity> blobs = findBlobsByDirectory(containerName, directoryName, false);
+      for (BlobEntity blob : blobs) {
+         if (!blob.isDirectory()) {
+            deleteBlob(containerName, blob.getKey());
+         }
+         else {
+            if (recursive) {
+               deleteBlob(containerName, blob.getKey());
```
nacx: Shouldn't this be deleted *after* all the blobs in the directory are deleted, after the recursive call?

jdbc/src/main/java/org/jclouds/jdbc/service/JdbcService.java
```diff
@@ -16,54 +16,243 @@
...
+      return chunks.build();
+   }
+
+   private void checkIntegrity(Blob blob, BlobEntity entity) throws IOException {
+      HashCode hash = blob.getMetadata().getContentMetadata().getContentMD5AsHashCode();
+      if (hash != null) {
+         HashingInputStream his = new HashingInputStream(Hashing.md5(), new JdbcInputStream(this, entity.getPayload().getChunks()));
+         HashCode actualHash = his.hash();
+         if (hash.equals(actualHash)) {
+            throw new IOException("MD5 hash code mismatch");
```
nacx: Would be more appropriate to throw an IllegalStateException here?

jdbc/src/main/java/org/jclouds/jdbc/util/JdbcInputStream.java
```diff
@@ -0,0 +1,67 @@
...
+      return b;
+   }
+
+   private boolean hasFinished() {
+      return chunks.size() == 0 && bytesRead >= currentChunk.getSize();
+   }
+
+   private void readNextChunk() {
+      if (chunks.size() > 0) {
+         this.currentChunk = jdbcService.findChunkById(chunks.get(0));
```
nacx: It this method returns `null`, the `hasFinished` will throw a NPE, and a NPE thrown from inside the `read` method of an InputStream is bad.

Fix the methods in this class so the `read` method only throws appropriate `IOExceptions`. Also add a unit test for this class that exercises the edge cases such as an empty initial chunk list, a null list (which is now allowed) an initial list with invalid IDs, etc.

rcoedo: To properly implement a unit test for this class I'll probably need some library to mock the jdbcService (e.g. mockito). Is this ok?

rcoedo:
Thank you both for the review, I'll check these issues and fix them.

jdbc/src/main/java/org/jclouds/jdbc/util/JdbcInputStream.java
nacx: Absolutely, but EasyMock is the one we use for this kind of tests. You'll have it already available as a transitive dependency, so there's no need to add it to the pom.xml

jdbc/src/main/java/org/jclouds/jdbc/service/JdbcService.java
rcoedo: Some blob tests seem to expect IOException in this case

jdbc/src/main/java/org/jclouds/jdbc/entity/ContainerEntity.java
rcoedo: We didn't need these (I generated them by mistake) so I just deleted them.

jdbc/src/main/java/org/jclouds/jdbc/service/JdbcService.java
rcoedo: A test fails if I don't do the integrity check. What's a good place to document this behavior?

zack-shoylev:
Seems like the license check fails, so one of the files doesn't have the right apache licence.

rcoedo:
I missed the headers on the new test class! It should be fixed now.

zack-shoylev:
Looking pretty good. Let's squash and I will do some tests and merge.

jdbc/src/main/java/org/jclouds/jdbc/service/JdbcService.java
```diff
@@ -16,54 +16,242 @@
...
+   public BlobEntity createOrModifyBlob(String containerName, Blob blob, BlobAccess blobAccess) throws IOException {
+      String key = blob.getMetadata().getName();
+      Date creationDate = null;
+      BlobEntity oldBlobEntity = findBlobById(containerName, key);
+      if (oldBlobEntity != null) {
+         creationDate = oldBlobEntity.getCreationDate();
+         deleteBlob(containerName, key);
+      }
+      BlobEntity blobEntity = blobToBlobEntity.apply(blob);
+      blobEntity.getPayload().setChunks(storeData(blob.getPayload().openStream()));
```
andrewgaul: Do you need to close the `InputStream`?  Shouldn't you also calculate the hash while copying instead of re-reading from the  database below?

jdbc/src/main/java/org/jclouds/jdbc/service/JdbcService.java
```diff
@@ -16,54 +16,242 @@
...
+         buffer = new byte[JdbcConstants.DEFAULT_CHUNK_SIZE];
+      }
+      data.close();
+      return chunks.build();
+   }
+
+   private void checkIntegrity(Blob blob, BlobEntity entity) throws IOException {
+      HashCode hash = blob.getMetadata().getContentMetadata().getContentMD5AsHashCode();
+      if (hash != null) {
+         HashingInputStream his = new HashingInputStream(Hashing.md5(), new JdbcInputStream(this, entity.getPayload().getChunks()));
```
andrewgaul: Do you need to close the input stream?

jdbc/src/main/java/org/jclouds/jdbc/service/JdbcService.java
andrewgaul: `IOException` matches other providers.

jdbc/src/test/java/org/jclouds/jdbc/BaseJdbcBlobStoreTest.java
```diff
@@ -0,0 +1,746 @@
+/*
```
andrewgaul: What is this test?  Why don't you get the same functionality from running `mvn integration-test jdbcblobstore`?

zack-shoylev:
@andrewgaul : thanks for the help
@rcoedo : ping

jdbc/src/main/java/org/jclouds/jdbc/service/JdbcService.java
rcoedo: The InputStream is closed in the storeData method :)

I think we need to check the integrity after the data is copied to the database. I think we need to either set things up in the database to make it return the hash of the copied data, or re-read the data from it and calculate it. Any other ideas?

jdbc/src/main/java/org/jclouds/jdbc/service/JdbcService.java
rcoedo: I could make this explicit, but I think it's safe to leave it opened (JdbcInputStream close method does nothing).

jdbc/src/test/java/org/jclouds/jdbc/BaseJdbcBlobStoreTest.java
rcoedo: This test is the equivalent of FilesystemBlobStoreTest. Some of the test methods are different from the ones included in the integration tests and since I wanted to be sure everything was working I took these too. Should I remove these?

rcoedo:
Thanks for the review @andrewgaul ! :)

zack-shoylev:
@rcoedo You will have to pull rebase master, fix conflicts, and push the branch before I can merge. Thanks!

rcoedo:
Sorry, I rebased my own fork (doh!)

zack-shoylev:
merged

PR #177, opened 2015-05-27
JCLOUDS-913 CloudSigma2 Adding password to returned login credentials

stuarthendren:
Password is not currently added to the login credentials.
This is done in other providers adapters.


nacx:
Thanks @stuarthendren!

The password in the LoginCredentials should be the SSH password used to connect to the node. In other words, the returned password should be the one that can be used to run scripts, and automated stuff on the node, and the VNC one does not meet that goal.

In CloudSigma, however, the images don't allow a password login, because an interactive prompt is shown on the first login to set the password. The only images that allow an automated interaction are the cloud-init ones, which allow to configure an SSH public key. jclouds configures that ssh key and the username, and intentionally sets the password to null.

Does this make sense?

stuarthendren:
Yes, that makes sense, and the ssh key gets added later, so we don't need to do anything here.
Sorry.

nacx:
No need for apologies! Every contribution is very welcome :)
We need to do much better documenting how each provider works.

PR #178, opened 2015-06-22
Fixes for Clocker

grkvlt:
Some minor fixes to the Docker provider discovered running Clocker.

- The `DockerTemplateOptions` code needed tidying and incosistencies fixed
- The OkHttp connection should support plaintext as well as TLS, to allow Powerstrip proxying
- The `HostConfig` domain object should support a list of `Dns` strings

### DIFF
--- a/docker/src/main/java/org/jclouds/docker/compute/options/DockerTemplateOptions.java
+++ b/docker/src/main/java/org/jclouds/docker/compute/options/DockerTemplateOptions.java
@@ -18,43 +18,46 @@ package org.jclouds.docker.compute.options;
 
 import static com.google.common.base.Objects.equal;
 import static com.google.common.base.Preconditions.checkNotNull;
+
 import java.util.List;
 import java.util.Map;
 
+import org.jclouds.compute.ComputeService;
 import org.jclouds.compute.options.TemplateOptions;
 import org.jclouds.domain.LoginCredentials;
 import org.jclouds.javax.annotation.Nullable;
 import org.jclouds.scriptbuilder.domain.Statement;
 
 import com.google.common.base.Objects;
-import com.google.common.base.Optional;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 
 /**
- * Contains options supported in the {@code ComputeService#runNode} operation on the
- * "docker" provider. <h2>Usage</h2> The recommended way to instantiate a
- * DockerTemplateOptions object is to statically import DockerTemplateOptions.* and invoke a static
- * creation method followed by an instance mutator (if needed):
- * <p/>
- * <code>
- * import static org.jclouds.docker.compute.options.DockerTemplateOptions.Builder.*;
- * <p/>
+ * Contains options supported by the {@link ComputeService#createNodesInGroup(String, int, TemplateOptions) createNodes}
+ * operation on the <em>docker</em> provider.
+ *
+ * <h2>Usage</h2>
+ *
+ * The recommended way to instantiate a
+ * DockerTemplateOptions object is to statically import {@code DockerTemplateOptions.Builder.*}
+ * and invoke one of the static creation methods, followed by an instance mutator if needed.
+ *
+ * <pre>{@code import static org.jclouds.docker.compute.options.DockerTemplateOptions.Builder.*;
+ *
  * ComputeService api = // get connection
  * templateBuilder.options(inboundPorts(22, 80, 8080, 443));
- * Set<? extends NodeMetadata> set = api.createNodesInGroup(tag, 2, templateBuilder.build());
- * <code>
+ * Set<? extends NodeMetadata> set = api.createNodesInGroup(tag, 2, templateBuilder.build());}</pre>
  */
 public class DockerTemplateOptions extends TemplateOptions implements Cloneable {
 
-   protected Optional<String> dns = Optional.absent();
-   protected Optional<String> hostname = Optional.absent();
-   protected Optional<Integer> memory = Optional.absent();
-   protected Optional<Integer> cpuShares = Optional.absent();
-   protected Optional<List<String>> commands = Optional.absent();
-   protected Optional<Map<String, String>> volumes = Optional.absent();
-   protected Optional<List<String>> env = Optional.absent();
-   protected Optional<Map<Integer, Integer>> portBindings = Optional.absent();
+   protected List<String> dns = ImmutableList.of();
+   protected String hostname;
+   protected Integer memory;
+   protected Integer cpuShares;
+   protected List<String> commands = ImmutableList.of();
+   protected Map<String, String> volumes = ImmutableMap.of();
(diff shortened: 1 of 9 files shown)


grkvlt:
@andreaturli Does this look OK? What about backporting to the 1.9.x branch?

andreaturli:
Lgtm

docker/src/main/java/org/jclouds/docker/compute/options/DockerTemplateOptions.java
```diff
@@ -384,7 +400,7 @@ public static DockerTemplateOptions overrideLoginCredentials(LoginCredentials cr
       }
 
       /**
-       * @see TemplateOptions#blockUntilRunning
+       * @see TemplateOptions#blockUntilRunning(boolean)
        */
       public static DockerTemplateOptions blockUntilRunning(boolean blockUntilRunning) {
          DockerTemplateOptions options = new DockerTemplateOptions();
```
nacx: What is the purpose of having the options builder methods accepting a `null` value? Is there a need to "clear" any previously set value?

BTW, having a look at this class, it makes no sense to have the variables being an Optional. They're only used internally so they could be perfectly plain nullable objects. That would allow to cleanup the compute service adapter code and make it more readable. WDYT?

docker/src/main/java/org/jclouds/docker/compute/options/DockerTemplateOptions.java
```diff
@@ -32,22 +34,24 @@
...
+ *
+ * <pre>{@code import static org.jclouds.docker.compute.options.DockerTemplateOptions.Builder.*;
+ *
  * ComputeService api = // get connection
  * templateBuilder.options(inboundPorts(22, 80, 8080, 443));
- * Set<? extends NodeMetadata> set = api.createNodesInGroup(tag, 2, templateBuilder.build());
- * <code>
+ * Set<? extends NodeMetadata> set = api.createNodesInGroup(tag, 2, templateBuilder.build());}</pre>
  */
 public class DockerTemplateOptions extends TemplateOptions implements Cloneable {
```
zack-shoylev: It might be better to refactor this to autovalue and autovalue builders?

nacx: Given that this extends the `TemplateOptions` class I'd say to leave this as-is, to avoid having a mixture of styles in one single class.
I agree this is something we can (and should) to in future PRs, but changing the TemplateOptions class in compute and all providers, but that is out of the scope of this PR.

zack-shoylev: This is reasonable enough. Thanks @nacx

docker/src/main/java/org/jclouds/docker/compute/options/DockerTemplateOptions.java
grkvlt: The code is using the pattern found in other similar classes, see `SoftLayerTemplateOptions` and so on. If we're going to change this would it not make more sense to change _all_ of these in a separate pull request?

nacx: A quick look to all existing template options subclasses show that the only ones that follow this pattern are the ones for Docker, SoftLayer and Azure Compute.
Given that this is not the common pattern and that this PR already cleans up the options in Docker, I think it's better to use this PR to remove the optionals from Docker, and fix SoftLayer and Azure in different PRs.

grkvlt: OK, thanks @nacx I'll fix that here, and SoftLayer and Azure can be done elsewhere. I hadn't realised my choice of template option classes was so unusual!

nacx: I wouldn't say unusual but "recent providers". I didn't notice this pattern was being used until I've had a deeper look at this PR. I'll take care of fixing SoftLayer and Azure. Thanks for spotting!

grkvlt:
@nacx @andreaturli I have changed the classes to remove the use of `Optional` and also updated the tests to reflect the change to `HostConfig`, is there anything else I need to do to make this available for merging with 1.9.1-SNAPSHOT as well as 2.0.0-SNAPSHOT?

andreaturli:
@grkvlt lgtm

@nacx, is it ok for you as well?

docker/src/main/java/org/jclouds/docker/compute/options/DockerTemplateOptions.java
```diff
@@ -68,29 +71,29 @@ public void copyTo(TemplateOptions to) {
...
       if (to instanceof DockerTemplateOptions) {
          DockerTemplateOptions eTo = DockerTemplateOptions.class.cast(to);
-         if (volumes.isPresent()) {
-            eTo.volumes(getVolumes().get());
+         if (!volumes.isEmpty()) {
+            eTo.volumes(volumes);
          }
-         if (hostname.isPresent()) {
-            eTo.hostname(hostname.get());
+         if (hostname != null) {
```
nacx: The null conditionals can be removed as these properties are `nullable` and there's no null check.

docker/src/main/java/org/jclouds/docker/compute/strategy/DockerComputeServiceAdapter.java
```diff
@@ -97,38 +97,34 @@ public DockerComputeServiceAdapter(DockerApi api) {
...
 
-      if (templateOptions.getCommands().isPresent()) {
-         containerConfigBuilder.cmd(templateOptions.getCommands().get());
+      if (!templateOptions.getCommands().isEmpty()) {
+         containerConfigBuilder.cmd(templateOptions.getCommands());
       }
 
-      if (templateOptions.getMemory().isPresent()) {
-         containerConfigBuilder.memory(templateOptions.getMemory().get());
+      if (templateOptions.getMemory() != null) {
```
nacx: Simplify this by removing the null check conditionals? These methods accept null values, so let's call them directly.

nacx:
Thanks @grkvlt! Just two minors, but lgtm. Mind addressing them while squashing the commits to prepare the merge?

grkvlt:
@nacx Thanks for the review, have squashed the commits and made the changes you suggested.

nacx:
Mind remove the conditionals [here](https://github.com/grkvlt/jclouds-labs/blob/fix/clocker/docker/src/main/java/org/jclouds/docker/compute/strategy/DockerComputeServiceAdapter.java#L104-L114) too?

grkvlt:
@nacx Good catch; done.

nacx:
Pushed to [master](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/7c5ded68) and [1.9.x](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/dcd1dc15). Thanks @grkvlt!

PR #181, opened 2015-06-26
Fixed compilation error due to change in jdbc LocalStorageStrategy

devcsrj:
There seems to be a compilation error in the *jdbc* module, due to a change in the `org.jclouds.blobstore.LocalStorageStrategy`.


andrewgaul:
Removed formatting changes, reworded commit message, and pushed to master as dc96c99a86d7899ab721feea6f3fcccc6eeed062.

PR #183, opened 2015-06-29
JCLOUDS-946: Properly scope images to the locations where they are available

nacx:
https://issues.apache.org/jira/browse/JCLOUDS-946

This changes how images are returned by the compute service. Since there are images that are only available in certain regions, and the mapping between an image and a location in jclouds is a one to one, we have to return one image for each location where it is available.

This PR does exactly this, and encodes the region in the image id, as done in other providers such as AWS or OpenStack.

/cc @ccustine


ccustine:
:+1: Tests run fine for me in several regions and the change looks good.

nacx:
Yay! Note that v1 is also affected but I won't apply this change there as the change in the image id can break existing code (for example the Jenkins plugin). This is something that can be expected when migrating from v1 to v2.

Thanks for the feedback @ccustine!

nacx:
Squashed, rebased, and pushed to master as [c001f260](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/c001f260).

PR #187, opened 2015-07-13
Add docker cacert to keystore with maven

csabapalfi:
Added a small script for unattended/automated setup (no need to type password, etc).

Assumes a default OSX and boot2docker setup.

Also I think we totally should just do this programatically instead of using the keytool. I mean simply passing in the cacert path via property to jclouds and it should do the rest.


csabapalfi:
Removed the shell script and now using the [keytool maven plugin](http://www.mojohaus.org/keytool/keytool-maven-plugin/usage.html) to import/delete docker ca certs.

You can now just run `sudo -E  mvn -Pimport-ca-cert generate-resources` and it'll import the boot2docker cacert to your Java cacerts keystore. Can also be fully fully customized via system properties.

Subsequent imports (e.g. after boot2docker regenerates certificates) requires deleting the imported certificate first. There's another profile for that:
`sudo -E  mvn -Pdelete-ca-cert -Pimport-ca-cert generate-resources`

This requires sudo because of the default file permissions of the Java cacerts keystore. I'm using the -E option to preserve environment variables when sudoing. Probably need to update setup docs, too.

csabapalfi:
This one is no longer relevant if #190 gets accepted.

PR #188, opened 2015-07-13
Upgrade to docker 1.7

csabapalfi:
Upgrading to docker 1.7 and also making sure all live tests pass.


csabapalfi:
Two live tests are still failing and trying fix those now.

csabapalfi:
One live test is still failing :(

csabapalfi:
That last live test is passing now.

grkvlt:
@nacx @andreaturli Can you have a look at this for 1.9.1 and 2.0.0? Also #189 and #190 please. Together they are for 1.7.0 integration, basically. We have been testing with https://github.com/brooklyncentral/clocker/pull/136

grkvlt:
@csabapalfi You mentioned reinstating the `Devices` map, is this still required?

csabapalfi:
I actually removed that as adding it back would require a new custom type (rather than a map) and not sure if it's worth it as most people are just happy with using `--priviliged`.

grkvlt:
From the 1.19 API docs I see that the `Devices` field in the create cointainer JSON is _A list of devices to add to the container specified as a JSON object_ and looks like the following array:

    [ {
        "PathOnHost": "/dev/deviceName",
        "PathInContainer": "/dev/deviceName",
        "CgroupPermissions": "mrw"
    }, { ... } ]

So, to support this we would need to make another DTO with these device fields. Looking at the next few lines of these docs there are several other configuration properties not supported by our driver, such as `Ulimits`nand `LogConfig`, so it is not alone. We should leave the `Devices` field as `null` in this PR, but if people feel strongly about supporting it then another PR should be created to do this, and we can explore the other gaps in the API coverage at the same time.

I agree with @csabapalfi that there is an existing workaround using the `--privileged` flag that people are using, which we can support via setting the `privileged` field in the DTO, so I'm happy with the way things are here.

grkvlt:
@csabapalfi I take it back, you're better squashing the commits from #189 and #190 into this PR

csabapalfi:
Done, closing #189 and #190

docker/src/main/java/org/jclouds/docker/config/DockerHttpApiModule.java
```diff
@@ -48,8 +65,45 @@ protected void bindErrorHandlers() {
...
+         ConnectionSpec cleartextSpec = new ConnectionSpec.Builder(ConnectionSpec.CLEARTEXT)
+                 .build();
+         client.setConnectionSpecs(ImmutableList.of(tlsSpec, cleartextSpec));
+         client.setSslSocketFactory(sslContextWithKeysSupplier.get().getSocketFactory());
+
+         if (utils.relaxHostname()) {
+            client.setHostnameVerifier(hostnameVerifier);
+         }
+         return client;
+      }
```
nacx: Why do you need to do this? I mean, this seems to be a copy of the OkHttp config module but overriding some defaults.
The idea behind the OkHttp client supplier is to allow users to provide their custom and initialised client (configuring there all needed SSL stuff, protocols, etc). Copy/pasting the OkHttp module here does not seem like the best approach.

Why isn't the OkHttp client suppler enough for Docker?

csabapalfi: Providing a custom supplier can work great unless you want to provide a custom `SSLSocketFactory` as the provider will override it if trustAllCerts is set. (See [OkHttpCommandExecutorServiceModule.java#L83](https://github.com/jclouds/jclouds/blob/master/drivers/okhttp/src/main/java/org/jclouds/http/okhttp/config/OkHttpCommandExecutorServiceModule.java#L83)). 
I think in case of docker even if trustAllCerts is set to true by any reason (e.g. for a different provider) we always want to have our custom SSLContext with the docker keys.

Maybe we can just use the default provider and a custom supplier and document that docker doesn't work when trustAllCerts is set or there're some other solutions. I agree that having duplicate code is a pretty bad code smell.

nacx: In that case I'd suggest that the docker provider uses a linked binding to bind the [UntrustedSSLContextSupplier](https://github.com/jclouds/jclouds/blob/master/core/src/main/java/org/jclouds/http/config/SSLModule.java#L76) to one that both implements the trustallcerts but has the Docker keys in it, but keep the OkHttp client supplier approach and avoid copy/pasting the OkHttp module config. WDYT?

csabapalfi: Great idea, makes a lot of sense :)

One area which is a bit unclear is what should happen when someone sets trustAllCerts=true but also sets a docker ca cert path. You can either setup a trustmanager with the docker ca cert and ignore trustAllCerts or ignore the docker ca cert and trust all certs.

I have a feeling if someone sets trustAllCerts they know what they're doing and that's a 'stronger statement' so in that case all else should be ignored (e.g docker cacert, etc).

nacx: Agree. The trustallcerts thing is something like "Oh, ok, I don't want to deal (and I don't care) with SSL at all but I have an SSL endpoint", so I'd say the Docker certs should also be ignored. If that is the case, the default implementation of the untrusted supplier should already work and no changes would be required.

csabapalfi: Ok, great, will update this PR accordingly. Might still be worth logging the fact the the docker ca cert was explicitly set but ignored because of `trustAllCerts` to prevent people going crazy while debugging and not realising. What do you think?

nacx: Feedback is always welcome, so +1 to log what is going on. There should no ned to provide a custom unstrusted ssl context for Docker. The logging can happen in the SSLContextWithKeysSupplier get() method, saying that since the trust all certs property has been enabled its configuration will be ignored. At that point you should have everything you need to build the log (and the injector will already have been created so you can properly use an injected logger).

csabapalfi: The reason we need the custom untrusted SSL context is to add the docker client key and certificate for client authentication (as docker uses mutual SSL). So even if we trust all certificates via a trustmanager, we still need to have the keys in the keymanager when initializing the SSL context.

csabapalfi:
@nacx pushed my refactor based on your comments. If you want I can squash commits otherwise please let me know if this is now good to merge.

/cc @andreaturli @grkvlt

nacx:
This looks great now @csabapalfi. Thanks!
Pushed to [master](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/73632907) and [1.9.x](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/5c42238a).

grkvlt:
@nacx @csabapalfi Thanks for this, it works great in Clocker now

PR #189, opened 2015-07-14
Use boot2docker env vars as default docker properties

csabapalfi:
We can get working credentials to access docker if boot2docker is used as it generates certificates and sets up environment variables.

This way you can simply type `mvn -Plive integration-test` to run the live test if you have boot2docker already set up (and have the output of `boot2docker shellinit` in your bash/zsh profiile)

(One caveat is that I had to use the [regex-property plugin](http://www.mojohaus.org/build-helper-maven-plugin/regex-property-mojo.html) as the `DOCKER_HOST` env var set by boot2docker starts with `tcp://...` instead of `https://...`)


csabapalfi:
Just had another idea: we could totally use the [keytool plugin](http://www.mojohaus.org/keytool/keytool-maven-plugin/usage.html) to import the ca cert generated by boot2docker by default. That way you can just install boot2docker, clone the project and voilà: live tests work.

What do you think?

docker/pom.xml
```diff
@@ -34,10 +34,10 @@
   <packaging>bundle</packaging>
 
   <properties>
-    <test.docker.endpoint>https://localhost:4243</test.docker.endpoint>
+    <!--<test.docker.endpoint>${env.DOCKER_HOST}</test.docker.endpoint>-->
```
grkvlt: Why commented out"?

csabapalfi: It's now set by the property-regex maven plugin based on env.DOCKER_HOST. Can be removed so I did just that. :)

csabapalfi:
Now part of #188

PR #190, opened 2015-07-15
Use docker cacert in SSLContext

csabapalfi:
You can manage the local cacert keystore with `keytool` on each host but when you deploy to production-like environments it's much simpler use a custom SSLContext which knows about the docker cacert.


docker/src/main/java/org/jclouds/docker/suppliers/SSLContextWithKeysSupplier.java
```diff
@@ -35,36 +52,20 @@
...
    private final TrustManager[] trustManager;
    private final Supplier<Credentials> creds;
 
+
    @Inject
-   SSLContextWithKeysSupplier(@Provider Supplier<Credentials> creds, HttpUtils utils, TrustAllCerts trustAllCerts) {
-      this.trustManager = utils.trustAllCerts() ? new TrustManager[]{trustAllCerts} : null;
+   SSLContextWithKeysSupplier(@Provider Supplier<Credentials> creds, @Named(DockerApiMetadata.DOCKER_CA_CERT_PATH) String caCertPath) {
       this.creds = creds;
+      this.trustManager = !isNullOrEmpty(caCertPath) ? getTrustManagerWithCaCert(caCertPath) : null;
```
grkvlt: Should we keep the original logic (to optionally use `TrustAllCerts`) if no CA certificate is supplied?

docker/src/main/java/org/jclouds/docker/suppliers/SSLContextWithKeysSupplier.java
```diff
@@ -113,8 +128,8 @@ private static PrivateKey getKey(String privateKey) {
       }
    }
 
-   private static String loadFile(final String filePath) throws IOException {
-      return Files.toString(new File(filePath), Charsets.UTF_8);
+   private static String loadFile(final String filePath) throws IOException{
+         return Files.toString(new File(filePath), Charsets.UTF_8);
```
grkvlt: Spacing is off here, I think

csabapalfi:
Now part of #188

docker/src/main/java/org/jclouds/docker/suppliers/SSLContextWithKeysSupplier.java
csabapalfi: Fixing on `feature/docker-1.7`

docker/src/main/java/org/jclouds/docker/suppliers/SSLContextWithKeysSupplier.java
csabapalfi: Fixing on `feature/docker-1.7`

PR #193, opened 2015-07-27
JCLOUDS-826: Add H2 provider

rcoedo:
Simple jpa provider with a preconfigured embedded database.


h2-jdbc/pom.xml
```diff
@@ -0,0 +1,158 @@
...
+    <version>2.0.0-SNAPSHOT</version>
+  </parent>
+  <artifactId>h2-jdbc</artifactId>
+  <name>jclouds h2 jdbc provider</name>
+  <description>jclouds implementation to target h2 databases</description>
+  <packaging>bundle</packaging>
+
+  <properties>
+    <maven.compile.source>1.6</maven.compile.source>
+    <maven.compile.target>1.6</maven.compile.target>
```
nacx: Is this really needed? Remove it if possible to inherit the global config.

h2-jdbc/pom.xml
```diff
@@ -0,0 +1,158 @@
...
+        </configuration>
+      </plugin>
+      <plugin>
+        <groupId>org.basepom.maven</groupId>
+        <artifactId>duplicate-finder-maven-plugin</artifactId>
+        <configuration>
+          <checkTestClasspath>false</checkTestClasspath>
+        </configuration>
+      </plugin>
+    </plugins>
```
nacx: Same for these plugins. Is this configuration really needed? Is there something to fix in the parent pom instead?

h2-jdbc/src/main/java/org/jclouds/h2/jdbc/H2JdbcApiMetadata.java
```diff
@@ -0,0 +1,60 @@
...
+import com.google.inject.Module;
+import org.jclouds.apis.ApiMetadata;
+import org.jclouds.blobstore.BlobStoreContext;
+import org.jclouds.h2.jdbc.config.H2JdbcBlobStoreContextModule;
+import org.jclouds.jdbc.JdbcApiMetadata;
+
+/**
+ * Implementation of {@link H2JdbcApiMetadata} for the H2 database
+ */
+@AutoService(ApiMetadata.class)
```
nacx: Remove this annotation. The provider metadata has the same ID than the API metadata and we want the ServiceLoader to pick just one of them.

nacx: In fact, this class should be removed and let the provider metadata use the `JdbcApiMetadata` directly. That's how generic apis and providers are supposed to be used. (A provider is an API + defaults + location).

h2-jdbc/src/main/java/org/jclouds/h2/jdbc/config/H2JdbcBlobStoreContextModule.java
```diff
@@ -0,0 +1,38 @@
...
+
+public class H2JdbcBlobStoreContextModule extends JdbcBlobStoreContextModule {
+
+   private static final String DEFAULT_FILE = "./jclouds-db";
+
+   protected void configure() {
+      super.configure();
+
+      Properties properties = new Properties();
+      properties.setProperty("hibernate.connection.url", "jdbc:h2:" + DEFAULT_FILE);
```
nacx: Move this properties to the provider metadata `defaultProperties` method.

nacx: IT would be good to have the file configurable via context properties in some way, if possible.

rcoedo: If I move it to the metadata, how can I inject it to the module? I tried so many things to make this work but I just couldn't find a way to do it! :(

nacx: Agree. Thinking about it I reworded the comment (and it seems that was *during* your reply). I don't know if it is possible at the Guice config time, so it is not a strong requirement.

rcoedo:
About the failed tests, it's a problem with the jdbc api. Here is a link to the JIRA issue https://issues.apache.org/jira/browse/JCLOUDS-979

rcoedo:
Can we rebuild? It should be good now.

nacx:
Sure, you just have to summon jenkins by saying: rebuild please! :)

rcoedo:
I knew there were some magic words but I couldn't remember them!

nacx:
LGTM now. Mind squashing the commits? Also, do you have an output of the live tests you could share in a gist or pastie? Thanks @rcoedo!

rcoedo:
I may be using an old snapshot or maybe a new bug was introduced, but one test in the jdbc api is still failing (BaseBlobIntegrationTest.testGetTwoRanges). http://pastebin.com/JR8DcJZQ

I'll take a look.

zack-shoylev:
You might have to rebuild the other projects, as it might be using your local snapshot first.

rcoedo:
Rebuild please

rcoedo:
Seems like it's passing now!

zack-shoylev:
merged

PR #194, opened 2015-07-27
JCLOUDS-971: init for etcd provider.

cdancy:
ADDED: version and statistics API as well as mock tests for each. Both APIS are very simple as they are all GET requests and do not require any parameters. 5 methods total between the 2. This should be a good starting point for us to break apart the API into smaller chunks for others/reviewers to digest.


cdancy:
Not sure how you want to go about breaking this into chunks but the API as a whole can be broken down into these features (ordered from smallest to biggest): VersionApi, StatisticsApi, MembersApi, and KeysApi. This PR covers the first 2 (Version and Statistics) as both are very small and not complicated. Once this gets in I will push integration tests and figured I'd move forward in much the same way with the other API's, that being: !.) submit initial API + mock tests 2.) when merged submit integration tests. The big piece here is the init which I've sent in so all other PR's will be much smaller. Hope it's not too big and let me know what you think of my plan. I can of course change if you want me to further granularize things.

etcd/pom.xml
```diff
@@ -0,0 +1,150 @@
...
+    <packaging>bundle</packaging>
+
+    <properties>
+        <test.etcd.endpoint>http://127.0.0.1:2379</test.etcd.endpoint>
+        <test.etcd.identity>FIXME</test.etcd.identity>
+        <test.etcd.credential>FIXME</test.etcd.credential>
+        <test.etcd.api-version>v2</test.etcd.api-version>
+        <jclouds.osgi.export>org.jclouds.etcd*;version="${project.version}"</jclouds.osgi.export>
+        <jclouds.osgi.import>
+            org.jclouds.labs*;version="${project.version}",
```
nacx: Are there any specific classes in `org.jclouds.labs` that need to be imported?

etcd/pom.xml
```diff
@@ -0,0 +1,150 @@
...
+            <type>test-jar</type>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.jclouds</groupId>
+            <artifactId>jclouds-compute</artifactId>
+            <version>${jclouds.version}</version>
+            <type>test-jar</type>
+            <scope>test</scope>
+        </dependency>
```
nacx: This does not seem a compute API that will require the compute service abstraction or an SSH driver. Remove the dependencies (even if they are just in test scope).

etcd/pom.xml
```diff
@@ -0,0 +1,150 @@
...
+        <dependency>
+            <groupId>org.apache.jclouds</groupId>
+            <artifactId>jclouds-core</artifactId>
+            <version>${jclouds.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.jclouds.driver</groupId>
+            <artifactId>jclouds-okhttp</artifactId>
+            <version>${jclouds.version}</version>
+        </dependency>
```
nacx: Is there any need to force a concrete HTTP driver? Remove this dependency if possible.

etcd/src/main/java/org/jclouds/etcd/config/EtcdHttpApiModule.java
```diff
@@ -0,0 +1,46 @@
...
+import org.jclouds.http.annotation.ClientError;
+import org.jclouds.http.annotation.Redirection;
+import org.jclouds.http.annotation.ServerError;
+import org.jclouds.http.config.ConfiguresHttpCommandExecutorService;
+import org.jclouds.http.okhttp.config.OkHttpCommandExecutorServiceModule;
+import org.jclouds.rest.ConfiguresHttpApi;
+import org.jclouds.rest.config.HttpApiModule;
+
+@ConfiguresHttpApi
+@ConfiguresHttpCommandExecutorService
```
nacx: Remove this annotation once the OkHttp module is not used.

etcd/src/main/java/org/jclouds/etcd/config/EtcdHttpApiModule.java
```diff
@@ -0,0 +1,46 @@
...
+	protected void bindErrorHandlers() {
+		bind(HttpErrorHandler.class).annotatedWith(Redirection.class).to(EtcdErrorHandler.class);
+		bind(HttpErrorHandler.class).annotatedWith(ClientError.class).to(EtcdErrorHandler.class);
+		bind(HttpErrorHandler.class).annotatedWith(ServerError.class).to(EtcdErrorHandler.class);
+	}
+	
+	protected void configure() {
+		super.configure();
+		install(new OkHttpCommandExecutorServiceModule());
+	}
```
nacx: Remove this. We don't want to force users to a concrete HTTP driver unless strictly required.

etcd/src/main/java/org/jclouds/etcd/domain/statistics/Leader.java
```diff
@@ -0,0 +1,40 @@
...
+	public abstract String leader();
+	
+	public abstract Map<String, Follower> followers();
+	
+	Leader() {
+	}
+	
+	@SerializedNames({"leader", "followers"})
+	public static Leader create(String leader, Map<String, Follower> followers) {
+		return new AutoValue_Leader(leader, followers);
```
nacx: Make the map immutable with `ImmutableMap.copyOf`.

etcd/src/main/java/org/jclouds/etcd/handlers/EtcdErrorHandler.java
```diff
@@ -0,0 +1,71 @@
...
+      Exception exception = null;
+      try {
+         
+         message = message != null ? message : String.format("%s -> %s", command.getCurrentRequest().getRequestLine(), response.getStatusLine());
+         exception = new HttpResponseException(message, command, response);
+
+      } catch (Exception e) {
+         exception = new HttpResponseException(command, response, e);
+      } finally {
+         if (exception == null) {
```
nacx: At this point the exception will never be null.

etcd/src/main/java/org/jclouds/etcd/handlers/EtcdErrorHandler.java
```diff
@@ -0,0 +1,71 @@
...
+   public void handleError(HttpCommand command, HttpResponse response) {
+
+      String message = parseMessage(response);
+      Exception exception = null;
+      try {
+         
+         message = message != null ? message : String.format("%s -> %s", command.getCurrentRequest().getRequestLine(), response.getStatusLine());
+         exception = new HttpResponseException(message, command, response);
+
+      } catch (Exception e) {
```
nacx: This catch block seems not to be reachable?

nacx:
Thanks @cdancy!

Apart from the comments, the code has to be formatted according to the jclouds style guide.

>Not sure how you want to go about breaking this into chunks.

I'd say submit one *complete* API at a time, being complete: API + mock tests + live tests.

etcd/pom.xml
cdancy: Required because the API makes heavy use of the http DELETE method and java 6 has known issues with this. The OkHttp driver solves this for java6.

etcd/src/main/java/org/jclouds/etcd/handlers/EtcdErrorHandler.java
cdancy: It's just a stub for now. Once we get going this class will flesh out quite a bit more.

cdancy:
@nacx that sounds fair enough and will do.

etcd/pom.xml
cdancy: It will need the compute api for testing once I've added the relevant testers. I can remove it but I'm going to need to add it back in once I've begun to add more tests.

etcd/src/main/java/org/jclouds/etcd/handlers/EtcdErrorHandler.java
cdancy: That is true but it's just a place holder for now. I can remove it but the code will get put back in a couple commits from now so I thought to just leave this bare minimum for now.

cdancy:
@nacx I've removed the ErrorHandler as it's not used with this PR and will make things less confusing and easier to digest. I've addressed the style issues (forgot to use the jclouds eclipse formatter) and added live tests. Live tests are not doing anything special as they are just GET requests with no parameters. I believe that addresses all your comments. Let me know how things look. Thanks!

cdancy:
Hmmm ... not exactly sure why the whitespace changes are not getting propagated to github when I commit

cdancy:
Failure not related to my changes. Code formatting issues are now solved.

cdancy:
Aaahhhh ... issued a force push and I believe it closed this PR. @nacx created another PR which references this but is more direct and to the point: https://github.com/jclouds/jclouds-labs/pull/195

PR #195, opened 2015-07-28
JCLOUDS-971: add statistics and miscellaneous API

cdancy:
continuing from https://github.com/jclouds/jclouds-labs/pull/194#event-367539973

statistics and miscellaneous API have both been fleshed out. Both are very small which is why they are being included in this PR. Mock and integration tests have been written and pass against latest etcd version: 2.1.1.


cdancy:
Failures due to jdbc provider not etcd.

nacx:
Looks pretty good! Just a few comments:

* Only add dependencies when needed: remove the OkHttp driver and the jclouds-compute one. Regarding the JRE and DELETE requests, IIRC they are supported if they don't have a body. Having a quick look at the etcd api docs I don't see any DELETE request with a body (I can be wrong), so the default http driver should work fine and we can let users pick their preferred driver.
* Live tests right now do not seem to have to be single-threaded. Remove that from the annotation and also the `threadCount` in the pom.xml.
* Seeing the nature of the MiscApi methods, consider removing the class and moving its methods to the EtcdApi class. Also move common method annotations to class level.

cdancy:
@nacx on your comments:

1.) All comments should be addressed. And you're right about the DELETE method, no request I could find uses a body when issuing a delete. All tests proved successful so it looks like we are good on that end.

2.) Comments have been addressed.

3.) I've added one more request to the MiscellaneousApi: metrics. Documentation on it can be found here: https://github.com/coreos/etcd/blob/master/Documentation/metrics.md. It's experimental at this point, so far as features/structure, but works as expected and returns the expected output. It's the last endpoint I want to get in before closing this PR and I think is a good reason to keep this as a separate class rather than merging with EtcdApi class. I'm not sure if the name 'Miscellaneous' fits but it's what I came up with to put all requests coming from the root URL in one place. Moving these into the EtcdApi seems it may clutter things IMO and I know there are commits for the next version to add at least 1 more endpoint to the root URL. If we can I'd like to leave them separated if only to make things easier to digest. The name is not great but I'm open to suggestions.

etcd/src/main/java/org/jclouds/etcd/domain/miscellaneous/Version.java
```diff
@@ -0,0 +1,38 @@
...
+import org.jclouds.json.SerializedNames;
+
+import com.google.auto.value.AutoValue;
+
+@AutoValue
+public abstract class Version {
+
+   public abstract String etcdserver();
+
+   public abstract String etcdcluster();
```
nacx: Use camel case for these methods.

etcd/src/main/java/org/jclouds/etcd/EtcdApiMetadata.java
```diff
@@ -0,0 +1,81 @@
...
+   public static Properties defaultProperties() {
+      Properties properties = BaseHttpApiMetadata.defaultProperties();
+      return properties;
+   }
+
+   public static class Builder extends BaseHttpApiMetadata.Builder<EtcdApi, Builder> {
+
+      protected Builder() {
+         super(EtcdApi.class);
+         id("etcd").name("Etcd API").identityName("Optional Username").defaultIdentity("Optional Username and Password")
```
nacx: It does not seem to be authentication in this API. Should the api metadata define the `defaultCredential` too then? Also add the `credentialName`.

cdancy: There is an auth API that was just introduced that I will tackle in a later PR. It's documentaiton can be found here:

https://github.com/coreos/etcd/blob/master/Documentation/auth_api.md

But yes by default I put these in if only to keep jclouds from complaining. Is there a better to way go about doing this?

etcd/src/main/java/org/jclouds/etcd/EtcdApiMetadata.java
```diff
@@ -0,0 +1,87 @@
...
+
+   public static class Builder extends BaseHttpApiMetadata.Builder<EtcdApi, Builder> {
+
+      protected Builder() {
+         super(EtcdApi.class);
+         id("etcd").
+         name("Etcd API").
+         identityName("username:password").
+         defaultIdentity("Optional Username").
+         defaultCredential("Optional Password").
```
nacx: I'm confused. Why not have the identity for the username and the credential for the password?

cdancy: Perhaps I'm confused here. For optional username/password should I then ONLY set 'defaultIdentity' and 'defaultCredential' or do something else?

nacx: The defaults are ok. The `identityName` property describes the identity (a username, an email, a token), and the `credentialName` describes the credential (a password, a private key, etc), so users know what to put there.
Once there is auth in this provider, the description of the identity would me misleading as it would encourage users to use the form "username:password" in the identity field. What about the credential field?
The defaults values are OK; it's just a matter about properly describing how the authentication is expected to be provided.

cdancy: @nacx Ok. So to address I've set 'identityName' and 'credentialName' to optional username/password respectively. To make jclouds happy, and get tests to run, I've set PROPERTY_IDENTITY and PROPERTY_CREDENTIAL to the same aforementioned values within the defaultProperties method call.

nacx: That's OK. One last thing is to think if this API is intended to be used without authentication, or authentication will be mandatory.
If the common case is to use it without auth, then the api metadata should provide the default values for the identity and credential (even dummy ones) so users can create a context without having to explicitly pass dummy values. Otherwise, this PR is good to go.

cdancy: @nacx it's intended to be used without as OOTB etcd does not default to requiring credentials. Authentiction, as an option, was only just added in the recent stable release and is what I'm targeting for this API.

So with that said ... I think we're good to go? I've nothing else to sneak in from my end WRT to this PR.

nacx: In that case you have to restore the `defaultIdentity` and `defaultCredential` methods (now we have only the ones that "describe" the authentication) so users don't have to explicitly pass dummy credentials when creating the context. You should then be able to remove the default credentials from the tests too.

cdancy: If I do that, and use the default* methods, then tests fail with an NPE on 'identityName'.

nacx: I'm not saying you have to use ones or the others. All four methods should be present:

* `identityName` and `credentialName` are mandatory for all ApiMetadata classes. They are just labels that define the format of the identity and credential so users know what to put there when creating the context. They're just labels to give a hint and allow, for example, tooling to prompt for the right thing.
* `defaultIdentity` and `defaultCredential` are default values to set for the identity and credential fields, if that makes sense, in case the user does not supply one of them. These are optional as in most apis/providers users are required both the identity and the credential.

In this case, as the credentials are not used, you have to set all four methods in your api metadata, in order to configure a default dummy credentials so the user does not have to provide them (which would be useless and confusing) when creating the context.

cdancy:
@nacx identityName and credentialName are now set to "Optional ..." with defaultIdentity and defaultCredential set to 'N/A'. Will this suffice?

nacx:
It looks great now. Thanks @cdancy!
Mind squashing the commits into a single one?

cdancy:
@nacx done!

cdancy:
Fixed last minute typo

nacx:
Pushed to master as [9cd3cfa5](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/9cd3cfa5). Thanks @cdancy!

PR #196, opened 2015-08-07
JCLOUDS-985: Add members API

cdancy:
Members API has been added with mock and integration tests of which all pass successfully. I refactored/formatted (eclipse format as well as feature/API comments) some code so a few more files than I would have liked were touched and made this PR just a little bit bigger than I would have liked. 

Please pick it apart and let me know how things look!


etcd/src/main/java/org/jclouds/etcd/domain/members/Member.java
```diff
@@ -0,0 +1,51 @@
...
+   public abstract List<String> peerURLs();
+
+   @Nullable
+   public abstract List<String> clientURLs();
+
+   Member() {
+   }
+
+   @SerializedNames({ "id", "name", "peerURLs", "clientURLs" })
+   public static Member create(String id, String name, List<String> peerURLs, List<String> clientURLs) {
```
zack-shoylev: Make this create method private, it shouldn't be visible to users.

Make another copy of this class, call it CreateMember, make sure you use AutoValue's builders to make sure a user can instantiate properly and then pass the object to:

Member add(@BinderParam(BindToJsonPayload.class) CreateMember member);

etcd/src/main/java/org/jclouds/etcd/features/MembersApi.java
```diff
@@ -0,0 +1,72 @@
...
+
+   /**
+    * @param member
+    *           non-existing member to add to cluster
+    * @return newly created member or null if member previously existed
+    */
+   @Named("members:add")
+   @Fallback(NullOnAddExistingMemberAnd409.class)
+   @POST
+   Member add(@BinderParam(BindToJsonPayload.class) Member member);
```
zack-shoylev: Some notes above. It makes more sense to have to different classes to handle what you send to the service and what you get back from the service, usually, at least for future-proofing.

cdancy: @zack-shoylev thanks. As a result of your comments I've now made this PR just a little bit bigger by ensuring all non-user-facing value classes have private creation methods. Also separated out the CreateMember from the Member class. All mock/integration tests pass locally.

etcd/src/main/java/org/jclouds/etcd/domain/members/CreateMember.java
```diff
@@ -0,0 +1,48 @@
...
+@AutoValue
+public abstract class CreateMember {
+
+   @Nullable
+   public abstract String name();
+
+   public abstract List<String> peerURLs();
+
+   @Nullable
+   public abstract List<String> clientURLs();
```
nacx: The factory method makes sure this is never null, so better remove the `@Nullable` annotation.

etcd/src/main/java/org/jclouds/etcd/domain/members/Member.java
```diff
@@ -0,0 +1,50 @@
...
+public abstract class Member {
+
+   public abstract String id();
+
+   @Nullable
+   public abstract String name();
+
+   public abstract List<String> peerURLs();
+
+   @Nullable
```
nacx: Same as above.

etcd/src/main/java/org/jclouds/etcd/features/MembersApi.java
```diff
@@ -0,0 +1,76 @@
...
+   /**
+    * @param member
+    *           non-existing member to add to cluster
+    * @return newly created member or null if member previously existed. Because
+    *         this call returns before registration is fully complete it is not
+    *         guaranteed that the Member class will have any variable, other than
+    *         peerURLs, from the CreateMember class present. 
+    */
+   @Named("members:add")
+   @Fallback(NullOnAddExistingMemberAnd409.class)
```
nacx: Don't use 409 fallbacks in POST/PUT methods. What does a `null` value mean to users? Although documented it is not a good practice: does it mean the member was not created (and does not exist) because the operation failed? Does it mean there was one that already existed? Returning `null` is too ambiguous and we don't use fallbacks in write operations (POST/PUT).

Let the method throw an exception. jclouds already has the `ResourceAlreadyExistsException` exception, so configure the error handler to throw it in this case instead of using a fallback.

etcd/src/main/java/org/jclouds/etcd/features/MembersApi.java
```diff
@@ -0,0 +1,76 @@
...
+
+   /**
+    * @param memberID
+    *           id of previously existing member
+    * @return true if member was deleted or false if id did not match an
+    *         existing member
+    */
+   @Named("members:delete")
+   @Path("/{id}")
+   @Fallback(BooleanOnDeleteNonExistentMemberAnd404.class)
```
nacx: Wouldn't it be enough to use the `FalseOnNotFoundOr404` fallback? If not, and you want to catch only responses with a concrete message... it does not look OK from a user perspective to expect a false (on certain circumstances), and an HttpResponseException on other 404.

etcd/src/main/java/org/jclouds/etcd/handlers/EtcdErrorHandler.java
```diff
@@ -0,0 +1,94 @@
...
+
+         message = message != null ? message
+               : String.format("%s -> %s", command.getCurrentRequest().getRequestLine(), response.getStatusLine());
+
+         switch (response.getStatusCode()) {
+            case 404:
+               if (command.getCurrentRequest().getMethod().equals("DELETE")) {
+                  if (command.getCurrentRequest().getEndpoint().getPath().startsWith(
+                        "/" + EtcdApiMetadata.API_VERSION + "/members") && message.contains("No such member")) {
+                     exception = new ResourceNotFoundException(message,
```
nacx: Can't we safely throw a RNFE on every 404, without checking the method and path?

etcd/src/main/java/org/jclouds/etcd/handlers/EtcdErrorHandler.java
```diff
@@ -0,0 +1,94 @@
...
+                           new HttpResponseException(command, response, message));
+                  }
+               }
+               break;
+            case 409:
+               if (command.getCurrentRequest().getMethod().equals("POST")) {
+                  if (command.getCurrentRequest().getEndpoint().getPath().endsWith(
+                        EtcdApiMetadata.API_VERSION + "/members") && message.contains("etcdserver: ID exists"))
+                     exception = new HttpResponseException(message, command, response);
```
nacx: I'd remove the method and path checks, and throw in general a `ResourceAlreadyExistsException ` exception if the ID exists, and the `HttpResponseException` otherwise.

etcd/src/test/java/org/jclouds/etcd/features/MembersApiLiveTest.java
```diff
@@ -0,0 +1,104 @@
...
+
+   private String selfID;
+   private Member nonSelfMember;
+   private Member addedMember;
+
+   @BeforeClass
+   protected void init() {
+      String id = api.statisticsApi().self().id();
+      assertNotNull(id);
+      selfID = id;
```
nacx: Avoid the intermediate variable and directly assign `selfID`.

etcd/src/test/java/org/jclouds/etcd/features/MembersApiLiveTest.java
```diff
@@ -0,0 +1,104 @@
...
+         if (!member.id().equals(selfID)) {
+            this.nonSelfMember = member;
+            break;
+         }
+      }
+   }
+
+   @Test(dependsOnMethods = "testListMembers")
+   public void testDeleteMember() {
+      assertNotNull(nonSelfMember);
```
nacx: Fail the test or skip it?

etcd/src/main/java/org/jclouds/etcd/handlers/EtcdErrorHandler.java
```diff
@@ -0,0 +1,94 @@
...
+   public void handleError(HttpCommand command, HttpResponse response) {
+
+      String message = parseMessage(response);
+      Exception exception = null;
+      try {
+
+         message = message != null ? message
+               : String.format("%s -> %s", command.getCurrentRequest().getRequestLine(), response.getStatusLine());
+
+         switch (response.getStatusCode()) {
```
nacx: I've seen that requests with malformed urls return a 400. Consider catching it here and propagating an `IllegalArgumentException`, as this is the jclouds convention.

etcd/src/test/java/org/jclouds/etcd/features/MembersApiLiveTest.java
cdancy: Done

etcd/src/main/java/org/jclouds/etcd/handlers/EtcdErrorHandler.java
cdancy: Makes sense now as the API is smaller. Once I flesh things out we may need to return and do it this way but for now, if only for simplicities sake, I'm on board.

etcd/src/main/java/org/jclouds/etcd/handlers/EtcdErrorHandler.java
cdancy: We can and will do so now. Once the API gets bigger we will have more use-cases, but as I've noted elsewhere, we can do this now for simplicities sake.

etcd/src/main/java/org/jclouds/etcd/handlers/EtcdErrorHandler.java
cdancy: Agreed and done.

etcd/src/main/java/org/jclouds/etcd/features/MembersApi.java
cdancy: Removed and instead will throw a ResourceNotFoundException. This seems to make the most sense IMO.

etcd/src/main/java/org/jclouds/etcd/features/MembersApi.java
cdancy: Agreed and done.

cdancy:
@nacx all comments should be addressed but it looks like the jdbc provider is having issues.

etcd/src/main/java/org/jclouds/etcd/domain/members/Member.java
```diff
@@ -0,0 +1,49 @@
...
+
+   public abstract List<String> clientURLs();
+
+   Member() {
+   }
+
+   @SerializedNames({ "id", "name", "peerURLs", "clientURLs" })
+   private static Member create(String id, String name, List<String> peerURLs, List<String> clientURLs) {
+      if (clientURLs == null)
+         clientURLs = ImmutableList.of();
```
nacx: In general, treat input parameters as final. The calling code could expect the clientURLs to still be `null` after having called this method.

etcd/src/test/java/org/jclouds/etcd/features/MembersApiLiveTest.java
```diff
@@ -0,0 +1,104 @@
...
+      List<Member> members = api().list();
+      assertNotNull(members);
+      assertTrue(members.size() > 0);
+      for (Member member : members) {
+         if (!member.id().equals(selfID)) {
+            this.nonSelfMember = member;
+            return;
+         }
+      }
+      throw new RuntimeException("Could not find another member in cluster with different id");
```
nacx: Better use `Assert.fail`

nacx:
Thanks @cdancy! Nice cleanup. I've just added a couple final comments about treating method parameters as final and a minor style thing about making tests fail.

Could you squash the commits into a single one and rebase to the latest master so I can cleanly merge this? There is no need to another fix commit; let's fix the remaining bits while squashing.

Thanks!

cdancy:
@nacx should be good to go. Thanks for the reviews and helpful comments getting the members API fleshed out.

etcd/src/main/java/org/jclouds/etcd/domain/members/CreateMember.java
```diff
@@ -0,0 +1,47 @@
...
+
+   public abstract List<String> peerURLs();
+
+   public abstract List<String> clientURLs();
+
+   CreateMember() {
+   }
+
+   @SerializedNames({ "name", "peerURLs", "clientURLs" })
+   public static CreateMember create(final String name, final List<String> peerURLs, List<String> clientURLs) {
```
zack-shoylev: You should probably make this method private.

Then, to be able to instantiate CreateMember, provide a builder using auto-value similar to:

https://github.com/jclouds/jclouds-labs-openstack/blob/master/openstack-heat/src/main/java/org/jclouds/openstack/heat/v1/options/CreateStack.java

etcd/src/main/java/org/jclouds/etcd/domain/members/Member.java
```diff
@@ -0,0 +1,49 @@
...
+   public abstract String id();
+
+   @Nullable
+   public abstract String name();
+
+   public abstract List<String> peerURLs();
+
+   public abstract List<String> clientURLs();
+
+   Member() {
```
zack-shoylev: I don't think this is needed?

etcd/src/main/java/org/jclouds/etcd/domain/members/CreateMember.java
cdancy: Aaahhhh yes you noted this before and I missed it amongst other changes being made. Jumping on this now

etcd/src/main/java/org/jclouds/etcd/domain/members/Member.java
cdancy: @zack-shoylev just following conventions I've seen elsewhere within jclouds-labs by providing an empty constructor for AutoValue classes. If the standard going forward is to remove these, as they are not really needed anyway, then I'm more than happy to remove it ;)

zack-shoylev: I see. =\ I am not aware of any issues if these are removed. Unless there is a code doc explaining why these are needed, I think remove.

nacx: It is a best practice. See: https://github.com/google/auto/blob/master/value/README.md#best-practices

etcd/src/main/java/org/jclouds/etcd/domain/members/CreateMember.java
nacx: I'd say don't use builders by default. We are talking about value objects that should be simple by nature. Using builders everywhere by default sounds like unnecessary over-engineering, but that's only my opinion :)

In this case, it has only three fields and perhaps makes more sense to provide several factory methods (if needed).

This said, I think consistency is important, so if the general case in this API is that value objects *users will build* are complex or justify the use of a builder, then use them.

zack-shoylev: @nacx I just think it's convenient and consistent to provide the Builder, especially because auto-value takes care of most of it.

etcd/src/main/java/org/jclouds/etcd/domain/members/Member.java
zack-shoylev: A-ha! Thanks @nacx !

etcd/src/main/java/org/jclouds/etcd/domain/members/CreateMember.java
cdancy: @zack-shoylev I think you made the point previously, maybe here or on IRC, that I agree with in that builders, at least so far as AutoValue classes are concerned, should be used when those classes are client facing. In this case the CreateMember class would be just that and so I think it makes sense to Builderize (?!?) it. Other classes which are not client facing, and are simply handed back from a function call, need not have this functionality. Maybe this is obvious but I thought it a point worth making. I'm actually writing up the builder class now so if we are in agreement I can send in another squashed commit with the builder functionality added on top. Going forward I will follow the same model with this provider, that being: client facing value classes will get builders if only to keep things consistent.

zack-shoylev: Thanks @cdancy ! That sounds perfect.

cdancy:
@nacx can we merge this? I'm going to backtrack on statements I made above about requiring builders. If it's not necessary at this point I'd rather leave it out. I was not having great success in implementing it and would like to continue moving forward and circle back at some later date to code the builders.

nacx:
Works for me @cdancy. @zack-shoylev?

zack-shoylev:
@cdancy @nacx Alright, then. We can revisit later if needed.

nacx:
Cherry-picked to master as [8b11430a](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/8b11430a). Thx!

PR #203, opened 2015-09-17
JCLOUDS-997 Allow nullable docker configuration (cmd, entrypoint, env)

kwart:
https://issues.apache.org/jira/browse/JCLOUDS-997

This commit allows null values for List Config entries.

The `null` has another meaning than en empty `List`.  E.g. when using `CMD` from Docker remote API then  `null` value means "use the image default", on the other hand the empty List has meaning "don't provide any command".

The `entrypoint` configuration is now available in `DockerTemplateOptions`.

Other 2 nullable field fixes were probably a typos in the `Config` class:
* `hostConfig`
* `dns`


nacx:
Pushed to [master](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/3440a0b5) and [1.9.x](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/e3dede23). Thanks @kwart!

PR #204, opened 2015-09-28
[JCLOUDS-1006] Fix DockerUntrustedSSLContextSupplier to allow test runs without TLS

kwart:
JIRA: https://issues.apache.org/jira/browse/JCLOUDS-1006

This PR allows to use Docker endpoints without TLS configured for running live tests.

The functionality change is located in the first commit. The second one is only source code reformat to fit the jclouds coding standards.


nacx:
Pushed to master. Thanks @kwart!

PR #205, opened 2015-09-28
[JCLOUDS-1007] Implemented Docker Exec support in MiscApi

kwart:
JIRA: https://issues.apache.org/jira/browse/JCLOUDS-1007

New methods were added to `MiscApi`:
* `execCreate` -  Sets up an exec instance in a running container with given Id;
* `execStart` - Starts a previously set up exec instance id.
* `execInspect` - Returns low-level information about the exec command id.

Additional support for Docker multiplexed standard stream (IN, OUT, ERR) was added in a form of `DockerInputStream`.

Usage:
```java
ExecCreateParams execCreateParams = ExecCreateParams.builder()
   .cmd(ImmutableList.<String> of("/bin/sh", "-c", "echo -n Standard >&1 && echo -n Error >&2"))
   .attachStderr(true).attachStdout(true).build();
Exec exec = miscApi.execCreate(container.id(), execCreateParams);
ExecStartParams execStartParams = ExecStartParams.builder().detach(false).build();
try (DockerInputStream inputStream = new DockerInputStream(miscApi.execStart(exec.id(), startParams))) {
   StdStreamData data = null;
   while (null != (data = inputStream.readStdStreamData())) {
      byte[] payload = data.getPayload();
      switch (data.getType()) {
         case OUT:            
            // process stadard output
            break;
         case ERR:
            // process error output
            break;
         default:
            throw new IllegalArgumentException();
            break;
      }
   }
}
ExecInspect execInspect = miscApi.execInspect(exec.id());
int exitCode = execInspect.exitCode();
```


kwart:
I've updated the PR and added also method `execInspect` - it's needed to retrieve exit code from the exec instance.

docker/src/main/java/org/jclouds/docker/domain/ExecCreateParams.java
```diff
@@ -0,0 +1,65 @@
...
+   }
+
+   @AutoValue.Builder
+   public abstract static class Builder {
+
+      public abstract Builder attachStdout(boolean b);
+
+      public abstract Builder attachStderr(boolean b);
+
+      public abstract Builder cmd(List<String> cmds);
```
nacx: We want lists to be immutable, but we don't want to force an implementation in the signature. Consider using [this builder pattern](https://github.com/jclouds/jclouds-labs-openstack/blob/master/openstack-heat/src/main/java/org/jclouds/openstack/heat/v1/options/CreateStack.java#L95-L115) to make the builder produce immutable lists. The trick is to add getters that are only visible to the builder itself but provide access to the actual list, so we can generate a proper immutable one when building, regardless of the list the used passed when calling the public builder methods.

docker/src/main/java/org/jclouds/docker/domain/ExecStartParams.java
```diff
@@ -0,0 +1,47 @@
...
+      return new AutoValue_ExecStartParams.Builder().detach(false);
+   }
+
+   @AutoValue.Builder
+   public abstract static class Builder {
+
+      public abstract Builder detach(boolean b);
+
+      public abstract ExecStartParams build();
+   }
```
nacx: Some domain classes have builders, some not. Should we remove all the builders, given that these classes are pretty simple, for consistency?

nacx:
This is really nice @kwart. There just a couple minor comments about the builders, but apart from that lgtm. Loving the stream test :)

kwart:
@nacx, thank you for the review.

I'll rewrite the builder with the immutable `List` implemented according to sample you've linked.

I would keep the builders in the new domain classes. IMO they improve code readability. Even it's more talkative, it's clearer to use 
```java
ExecStartParams.builder().detach(false).build();
```
than
```java
ExecStartParams.create(false);
```

It's also simpler to keep API backward compatible with the builders. E.g. if somebody will add [tty attribute](https://docs.docker.com/reference/api/docker_remote_api_v1.16/#exec-create) implementation to `ExecCreateParams` class, then instead of introducing new argument of the `create()` method, one new method to Builder's flow API will be added.

The reason why the builders were not included in all domain classes was the fact that some of them are not intended to be used by users directly. They are used only as a return values from remote API calls.

I'll add the fixes as a new commit and if it's OK, I can squash it into one after the review.

nacx:
Thanks for the explanation. Changes will be minors and the PR is not big, so feel free to squash the commits directly :)
Thanks!

grkvlt:
@kwart Looks good, this will be useful for Clocker too, thanks.

kwart:
Rebased to latest master.

nacx:
Pushed to master as [e2a4ca0d](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/e2a4ca0d). Thanks @kwart!

PR #206, opened 2015-10-02
[JCLOUDS-1011] force closing HTTP client connection after using Docker.build API

kwart:
JIRA: https://issues.apache.org/jira/browse/JCLOUDS-1011

Reusing HTTP connection after calling Docker `build` REST API causes troubles in subsequent calls. We need to close the connection after the `build` call.


nacx:
lgtm. The MiskApiMockTest should be updated too, to make sure we verify the appropriate headers are being sent.

kwart:
The assertions were added.

nacx:
Pushed to master as [f1a53f0f](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/f1a53f0f). Thanks @kwart!

PR #208, opened 2015-10-08
[JCLOUDS-1010] use a lightweight image as a base for Docker build test

kwart:
JIRAs:
* https://issues.apache.org/jira/browse/JCLOUDS-1010
* https://issues.apache.org/jira/browse/JCLOUDS-1017
* https://issues.apache.org/jira/browse/JCLOUDS-1020

**JCLOUDS-1010**
To limit bandwidth during running Docker provider live tests a lightweight image is used. The originally used Ubuntu image with OpenSSH server was replaced by Alpine Linux with dropbear SSH server.

**JCLOUDS-1017**
Implement `tagImage` in `ImageApi` of Docker provider.

**JCLOUDS-1020**
Fix query parameter for adding image tag in `BuildOptions` class (Docker provider).


docker/src/test/resources/Dockerfile
```diff
@@ -14,19 +14,21 @@
...
 
-RUN apt-get update && apt-get install -y openssh-server
-RUN mkdir /var/run/sshd
-RUN echo 'root:screencast' | chpasswd
-RUN sed -i 's/PermitRootLogin without-password/PermitRootLogin yes/' /etc/ssh/sshd_config
 
-# SSH login fix. Otherwise user is kicked off after login
-RUN sed 's@session\s*required\s*pam_loginuid.so@session optional pam_loginuid.so@g' -i /etc/pam.d/sshd
+FROM alpine:3.2
+MAINTAINER JClouds Dev <jclouds-dev@apache.org>
```
nacx: [minor] Change to  `dev@jclouds.apache.org`

nacx:
lgtm. Thanks @kwart!

docker/src/test/resources/Dockerfile
kwart: fixed

nacx:
Looks like a transient failure: rebuild please

docker/src/main/java/org/jclouds/docker/features/ImageApi.java
```diff
@@ -102,4 +104,23 @@
...
+    * @param tag
+    *           the new tag name
+    * @param force
+    *           force create if tag already exists
+    * @return true if tag is created, false when image is not found.
+    */
+   @Named("image:tag")
+   @POST
+   @Path("/images/{name}/tag")
+   @Fallback(FalseOnNotFoundOr404.class)
```
nacx: In general we try to avoid using 404 fallbacks in POST/PUT methods, as often they don't provide the right feedback to the user. Operations done on objects that are supposed to exist will already [throw a ResourceNotFoundException](https://github.com/jclouds/jclouds-labs/blob/master/docker/src/main/java/org/jclouds/docker/handlers/DockerErrorHandler.java#L71-L75). I'd change this method to return `void` and remove the fallback.

docker/src/test/java/org/jclouds/docker/compute/SshToCustomPortLiveTest.java
```diff
@@ -0,0 +1,223 @@
...
+      super.setupContext();
+
+      final String tag = toTag(IMAGE_REPOSITORY, IMAGE_TAG_1);
+      BuildOptions options = BuildOptions.Builder.tag(tag).verbose(false).nocache(false);
+      InputStream buildImageStream;
+      try {
+         buildImageStream = api().getMiscApi().build(BaseDockerApiLiveTest.tarredDockerfile(), options);
+         consumeStreamSilently(buildImageStream);
+      } catch (IOException e) {
+         e.printStackTrace();
```
nacx: Should the exception be propagated instead, to avoid executing the tests?

docker/src/test/java/org/jclouds/docker/compute/SshToCustomPortLiveTest.java
```diff
@@ -0,0 +1,223 @@
...
+      image = api().getImageApi().inspectImage(tag);
+      api().getImageApi().tagImage(image.id(), IMAGE_REPOSITORY, IMAGE_TAG_2, true);
+   }
+
+   /**
+    * After the test remove created image (with all tags).
+    * 
+    * @see #setupContext()
+    */
+   @AfterClass
```
nacx: Use `@AfterClass(alwaysRun = true)` to make sure the cleanup is executed even if some tests fail.

docker/src/test/java/org/jclouds/docker/compute/SshToCustomPortLiveTest.java
```diff
@@ -0,0 +1,223 @@
...
+      try {
+         try {
+            while (isr.read(tmpBuff) > -1) {
+               // empty
+            }
+         } finally {
+            isr.close();
+         }
+      } catch (IOException e) {
+         e.printStackTrace();
```
nacx: Propagate the exception or use the `java.util.Logger.getAnonymousLogger()` to log test errors.

docker/src/test/java/org/jclouds/docker/features/ImageApiLiveTest.java
```diff
@@ -38,18 +49,40 @@ public void testCreateImage() {
...
-      assertNotNull(api().listImages());
+      List<ImageSummary> listImages = api().listImages();
+      assertNotNull(listImages);
+
+      ImageSummary taggedImage = Iterables.find(listImages, new Predicate<ImageSummary>() {
+         @Override
+         public boolean apply(ImageSummary input) {
+            return input.repoTags().contains("jclouds:testTag");
+         }
+      });
```
nacx: The `find` method will throw an exception if not found, so the assertion below is redundant. Use `tryFind` if you want to be explicit or just remove the assertion.

docker/src/test/java/org/jclouds/docker/features/ImageApiLiveTest.java
```diff
@@ -38,18 +49,40 @@ public void testCreateImage() {
...
+      assertNotNull(taggedImage);
    }
 
    @Test(dependsOnMethods = "testListImages")
```
nacx: Use `alwaysRun = true` to remove the images even if the "listImages" test fails?

nacx:
lgtm. Just some small comments in the tests.

kwart:
Commits edited according to the comments.

nacx:
Pushed to msater and 1.9.x. Thanks!

PR #209, opened 2015-10-09
JCLOUDS-1014: Make the login port lookup function configurable

nacx:
https://issues.apache.org/jira/browse/JCLOUDS-1014

Some Docker images can have the SSH server running in a port different than 22, and that configuration might not be populated in the container's networking configuration section. This pull request introduces a pluggable mechanism to let users provide their own port lookup function.

One could provide a custom port lookup function as follows:

```java
// Define custom functions that given a container, returns the login port
LoginPortForContainer CustomPortLookup = new LoginPortForContainer() {
   @Override public Integer apply(Container input) {
      return 2222;  // Look for the port in the container object
   }
};

// Create a Guice module that configures binds the function
Module customLookupModule = new AbstractModule() {
   @Override protected void configure() {
      bind(LoginPortForContainer.class).to(CustomPortLookup.class).in(Scopes.SINGLETON);
};

// Provide the lookup module in the module list when creating the context
ComputeServiceContext context = ContextBuilder.newBuilder("docker")
   .credentials("identity", "credential")
   .modules(ImmutableSet.<Module> of(customLookupModule, new SshjSshClientModule()))
   .buildView(ComputeServiceContext.class);
```

Alternatively, this pull request also introduces a convenient clean way to bind custom port lookup functions to concrete Docker images. This can be done by creating the custom Guice module like shown in the example below, and creating the context including that module as shown in the previous example:

```java
Module customLookupModule = new AbstractModule() {
   @Override protected void configure() {
      MapBinder<String, LoginPortForContainer> lookup = loginPortLookupBinder(binder());
      lookup.addBinding(".*alpine-ext.*").toInstance(AlpinePortLookup);
      lookup.addBinding(".*ubuntu.*").toInstance(GenericUbuntuPortLookup);
      lookup.addBinding(".*ubuntu:12\\.04.*").toInstance(Ubuntu1204PortLookup);
};
```

/cc @kwart

### DIFF
--- a/docker/pom.xml
+++ b/docker/pom.xml
@@ -85,6 +85,10 @@
       <version>${project.version}</version>
     </dependency>
     <dependency>
+      <groupId>com.google.inject.extensions</groupId>
+      <artifactId>guice-multibindings</artifactId>
+    </dependency>
+    <dependency>
       <groupId>org.apache.jclouds</groupId>
       <artifactId>jclouds-core</artifactId>
       <version>${project.version}</version>
--- a/docker/src/main/java/org/jclouds/docker/compute/config/DockerComputeServiceContextModule.java
+++ b/docker/src/main/java/org/jclouds/docker/compute/config/DockerComputeServiceContextModule.java
@@ -16,8 +16,6 @@
  */
 package org.jclouds.docker.compute.config;
 
-import com.google.common.base.Function;
-import com.google.inject.TypeLiteral;
 import org.jclouds.compute.ComputeServiceAdapter;
 import org.jclouds.compute.config.ComputeServiceAdapterContextModule;
 import org.jclouds.compute.domain.Hardware;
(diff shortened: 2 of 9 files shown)


nacx:
The build got stuck and I had to abort it. Let's try again: rebuild please!

andreaturli:
@nacx not sure I get the purpose of this feature: the login port like any other ports depends on the docker engine, specifically the network type, and it is not a feature of an image. Am I missing something?

nacx:
As said in the conversation copied in the jira issue, there are cases where there is no network info in the container details. If the network mode is host, that's the only info you get, but in that case the port 22 might be taken by the host itself, so it won't be available to the container.

Some images address this by reading the port froman env var, orhers might hardcode another port, etc. This PR allows users to configure a *fallback* method to be invoked if the container does not have the port info populated.

nacx:
@andreaturli @kwart Thinking about this a bit more, I refactored the port lookup code to its own function. This allows users to configure per-image lookup functions, but also to easily override the default implementation. Have a look at the updated PR description for the details.

nacx:
I've added a small change based on @kwart's feedback. Images are better mapped using patterns, as this allows a more precise mapping between concrete images and their lookup functions.

kwart:
I like the new version. 
I'm just playing with replacing the default lookup implementation.

I'll play with the imageName-to-Function mapping later.

Thanks, @nacx .

PR #212, opened 2015-10-21
JCLOUDS-1022: Automatically handle DigitalOcean rate limit

nacx:
https://issues.apache.org/jira/browse/JCLOUDS-1022

This pull request introduces a retry handler to deal with the [DigitalOcean rate limit](https://developers.digitalocean.com/documentation/v2/#rate-limit). The retry handler will read the rate limit information provided in the response to determine the amount of time the request has to be delayed. It will block the execution thread that amount of time, to finally return `true` so the command executor executing the request will retry it.

It is basically the same retry mechanism that is used to follow redirects, but introducing an active wait *before* returning if the request has to be retried or not. This allows us to apply this directly in the http layer, making it transparent to all the code and request execution.

The rate limit is disabled by default, and rate limit failures will be propagated as `RateLimitExceededExceptions` (note this PR depends on https://github.com/jclouds/jclouds/pull/867 and the build may fail until the latest snapshots are published). To use the rate limit features users will just have to include the `DigitalOcean2rateLimitModule` when creating the context.

The `jclouds.max-retries` and the `jclouds.jclouds.max-ratelimit-wait` can be used to customize the number of retries and the maximum acceptable time to wait for a rate limited request,

/cc @ccustine Feedback is very welcome!


nacx:
Snapshots should be in place now. Rebuild please!

nacx:
Snapshots should be in place now: rebuild please!

ccustine:
@nacx This is a great addition and this is quality code. I haven't had time to test but the code looks great and this should be highly useful for DO users.  Thanks for doing this.

nacx:
I've just added a `DigitalOcean2RateLimitExceededException` that extends the default one and provides detailed information about the rate limits. This will allow users to be able to deal with the rate limit manually, even if the rate limit module is not installed.

nacx:
Squashed and merged to [master](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/70dba8eb) and [1.9.x](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/917c06ee). Thanks for the feedback @ccustine!

PR #219, opened 2015-10-30
Remove DigitalOcean v1

nacx:
The version 1 of the DigitalOcean api will be turned off in a few days. The provider must be removed and users should migrate to digitalocean2.


nacx:
Another labs build that hangs in jdbc... https://jclouds.ci.cloudbees.com/job/jclouds-labs-pull-requests/922/console /cc @rcoedo 
rebuild please

rcoedo:
If this only happens with hibernate maybe we should disable the hibernate tests and keep the eclipselink tests, at least until we fix it. It is causing too much trouble.

nacx:
Disabling the tests will mostly lead us to forget about them, so I'd rather keep them enabled. This will also force us to investigate what's going on! :)

PR #220, opened 2015-11-02
JCLOUDS-1033: Don't fail if no private key has been provided in DigitalOcean

nacx:
https://issues.apache.org/jira/browse/JCLOUDS-1033

We shouldn't fail if no ssh private key has been provided in order to let users use the ssh-agent to authenticate.


nacx:
rebuild please

nacx:
Pushed to master as [d91ea5f4](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/d91ea5f4).

PR #224, opened 2015-12-10
JCLOUDS-1047: Fix ProfitBricks compute service live tests

devcsrj:
Changes:
1. Implemented `ProfitBricksTemplateBuilderImpl`, so that `Location`s used by built template will always be of scope `ZONE`. Looks for the [first found](https://github.com/jclouds/jclouds-labs/pull/224/files#diff-1500f3a9004863afdd8a7869407b390aR68) `ZONE` if `REGION` is provided. Also fixes [errors](https://gist.github.com/devcsrj/9675b0c3bb04a133fb50/86898b12fc9b494a6de31c08028fa82daa210a8e#file-jclouds-profitbricks-live-tests-log-L498) when only providing `imageId` in template.
2. Deleted unused interfaces left from previous PR. (`org.jclouds.profitbricks.domain.internal.*`)
3. [Added](https://github.com/jclouds/jclouds-labs/pull/224/files#diff-1a27cebd5920714ae4c207026edd5255R129) *provisionable type* as metadata, because API complains if password has [invalid characters](https://github.com/jclouds/jclouds-labs/blob/master/profitbricks/src/main/java/org/jclouds/profitbricks/util/Preconditions.java#L96).
4. Incorrect [content-length](https://github.com/jclouds/jclouds-labs/pull/224/files#diff-e9ba6240d723094ed0863af75d9416a5R48).

Ran with:
```
mvn clean install -Plive -Dtest.profitbricks.identity=<identity> -Dtest.profitbricks.credential='<credential>' -Dtest.profitbricks.template="imageId=<snapshotId>,loginUser=<user>:<password>"
```
[Live tests](https://gist.github.com/devcsrj/9675b0c3bb04a133fb50/5d1d3f6d5bc291c291679e56231d1af645041983) 
`Tests run: 92, Failures: 2, Errors: 0, Skipped: 0`

(edit: Added description)


devcsrj:
[Live tests](https://gist.github.com/devcsrj/9675b0c3bb04a133fb50/1211b7546f9b17b5d4edaab578092b2b45fb44ac)
`
Tests run: 92, Failures: 1, Errors: 0, Skipped: 12`

devcsrj:
Hi @nacx !

I'm stuck with one last live test, failing in the [`weCanCancelTasks`](https://github.com/jclouds/jclouds/blob/master/compute/src/test/java/org/jclouds/compute/internal/BaseComputeServiceLiveTest.java#L305) method from [`testAScriptExecutionAfterBootWithBasicTemplate:262`](https://gist.github.com/devcsrj/9675b0c3bb04a133fb50/1211b7546f9b17b5d4edaab578092b2b45fb44ac#file-jclouds-profitbricks-live-tests-log-L332). Any insight/similar issue/ideas or clue where to start investigating this?

nacx:
Excelent job @devcsrj! And thanks for the very detailed PR. It really makes the difference when it comes to review it!

Regarding the Jetty test... Is it possible that the IP addresses that are being added to the node are not reachable? Or is it possible that the image you are using has a firewall (iptables/SELinux) enabled by default and is filtering traffic to port 8080?

The second one is more interesting... and I haven't any ideas yet. It looks like it can authenticate to submit the script, but then it fails to authenticate when it tries to abort the execution of the script on the node. I'm not sure if it is related, but given how you [generate the password if mising](https://github.com/devcsrj/jclouds-labs/blob/develop/profitbricks/src/main/java/org/jclouds/profitbricks/compute/ProfitBricksComputeServiceAdapter.java#L114) and how you [return the node credentials](https://github.com/devcsrj/jclouds-labs/blob/develop/profitbricks/src/main/java/org/jclouds/profitbricks/compute/ProfitBricksComputeServiceAdapter.java#L227-L230), if the password is not provided and jclouds auto-generates one, that could not be the right password configured in the image (given that with your latest changes you are not explicitly setting the password if the image is a snapshot). Could this be introducing inconsistencies in the noe credentials that may cause authentication failures (perhaps related to these issues)?

devcsrj:
> Regarding the Jetty test... Is it possible that the IP addresses that are being added to the node are not reachable?

You were right. It turns out, it was our own network's firewall that blocks my machine's request to the remote node's public IP. (This explains as well as to why this test passes when I run this at home. *sigh* -.-)

> Failed test: weCanCancelTasks method from testAScriptExecutionAfterBootWithBasicTemplate:262 

To debug this: I logged in to the node before the tests [executes scripts](https://github.com/jclouds/jclouds/blob/master/compute/src/test/java/org/jclouds/compute/internal/BaseComputeServiceLiveTest.java#L229). After the test [adds a user](https://github.com/jclouds/jclouds/blob/master/compute/src/test/java/org/jclouds/compute/internal/BaseComputeServiceLiveTest.java#L253), I have confirmed that a user `foo` was added, along with it's home directory `/over/ridden/foo`. The test afterwards asserts that the `LoginCredentials` should've been [changed](https://github.com/jclouds/jclouds/blob/master/compute/src/test/java/org/jclouds/compute/internal/BaseComputeServiceLiveTest.java#L259) - which it does; from previously 
> `identity=root,password=<template password from args>`

 to 
> `identity=foo,privateKey=<my machine's private key>`. 

I verified `foo`'s [`authorized_keys`](https://github.com/jclouds/jclouds/blob/master/compute/src/test/resources/runscript_adminUpdate.sh#L99) in the remote node, and my private key is there. Finally, it fails when the `ListenableFuture` is [cancelled](https://github.com/jclouds/jclouds/blob/master/compute/src/test/java/org/jclouds/compute/internal/BaseComputeServiceLiveTest.java#L305).

I stopped the test before it [destroyed](https://github.com/jclouds/jclouds/blob/master/compute/src/test/java/org/jclouds/compute/internal/BaseComputeServiceLiveTest.java#L271) the node, and tried to login to the node with `ssh foo@ip`. It worked.

Now, the weirdest part of this last test, is that, sometimes, it *passes*. **It passes**. This usually, but not consistently, happens when I walk through line-by-line with the debugger.

Stumped. Thoughts?

nacx:
Unfortunately no. I'll try to debug the tests and see what I find. I'd like to keep this PR open to investigate this, in case you find what is going on, but I don't want to block it forever. Since the authentication works (otherwise other tests would be failing), I think we can safely merge this PR, say, early next week, as it brings a lot of improvements.

Could you try creating a minimal `main` that runs a script and cancels it, to help you verify if this is actually broken or it is just something in the test workflow that is causing some kind of conflict?

devcsrj:
So I've created two minimal `main` files to run the offending test:
1. [RunScriptUsingBasicLogin](https://github.com/devcsrj/bug-reports/blob/master/jclouds-1047/src/main/java/io/devcsrj/report/jclouds/RunScriptUsingBasicLogin.java#L97) - attempts to cancel using the default identity+credentials.
2. [RunScriptUsingPrivateKey](https://github.com/devcsrj/bug-reports/blob/master/jclouds-1047/src/main/java/io/devcsrj/report/jclouds/RunScriptUsingPrivateKey.java#L97) - attempts to cancel after overriding with user `foo`, then using it plus private key.

Both files reproduce the issue. BUT
> ..the weirdest part of this last test, is that, sometimes, it passes.

I can now reproduce the *weirdest part*. When debugging any of the two, adding a breakpoint to [`ssh.connect()`]( https://github.com/jclouds/jclouds/blob/master/compute/src/main/java/org/jclouds/compute/callables/SudoAwareInitManager.java#L74), then stepping to `return runAction(action);` will succeed, in contrast to the [original exception](https://gist.github.com/devcsrj/9675b0c3bb04a133fb50/1211b7546f9b17b5d4edaab578092b2b45fb44ac#file-jclouds-profitbricks-live-tests-log-L321). Without doing this, the connection would fail. 

I wonder if this has something to do with the delay I present before stepping through that, or maybe synchronization issues? Continuing investigation.

nacx:
Do we have any progress here? To put together the simplest test possible, could you try running something like the following:

```java
NodeMetadata node = Iterables.getOnlyElement(compute.createNodesInGroup("tasks", 1, template));
ListenableFuture<ExecResponse> future = compute.submitScriptOnNode(node.getId(), "sleep 300", nameTask("sleeper").runAsRoot(false));

checkState(!future.isDone(), "future.isDone()");
future.cancel(true);
checkState(future.isCancelled(), "future.isCancelled()");

```

devcsrj:
Ran the test using your provided [snippet](https://github.com/devcsrj/bug-reports/blob/master/jclouds-1047/src/main/java/io/devcsrj/report/jclouds/RunScriptUsingBasicLogin.java#L107), and it produced the same exception:

```
Creating test datacenter
Building template
Bootstrapping node
Testing future.cancel()
Deleting test datacenter
Exception in thread "main" org.jclouds.rest.AuthorizationException: (root:pw[97db1846570837fce6ff62a408f1d26a]@00.00.00.00:22) (root:pw[97db1846570837fce6ff62a408f1d26a]@00.00.00.00:22) error acquiring {hostAndPort=00.00.00.00:22, loginUser=root, ssh=1533657934, connectTimeout=60000, sessionTimeout=60000} (out of retries - max 7): Exhausted available authentication methods
	at org.jclouds.sshj.SshjSshClient.propagate(SshjSshClient.java:395)
	at org.jclouds.sshj.SshjSshClient.acquire(SshjSshClient.java:205)
	at org.jclouds.sshj.SshjSshClient.connect(SshjSshClient.java:225)
	at org.jclouds.compute.callables.SudoAwareInitManager.refreshAndRunAction(SudoAwareInitManager.java:74)
	at org.jclouds.compute.callables.BlockUntilInitScriptStatusIsZeroThenReturnOutput.interruptTask(BlockUntilInitScriptStatusIsZeroThenReturnOutput.java:159)
	at com.google.common.util.concurrent.AbstractFuture.cancel(AbstractFuture.java:136)
	at io.devcsrj.report.jclouds.RunScriptUsingBasicLogin.testWeCanCancelTask(RunScriptUsingBasicLogin.java:107)
	at io.devcsrj.report.jclouds.RunScriptUsingBasicLogin.main(RunScriptUsingBasicLogin.java:92)
Caused by: net.schmizz.sshj.userauth.UserAuthException: Exhausted available authentication methods
	at net.schmizz.sshj.SSHClient.auth(SSHClient.java:217)
	at net.schmizz.sshj.SSHClient.auth(SSHClient.java:193)
	at net.schmizz.sshj.SSHClient.authPassword(SSHClient.java:278)
	at net.schmizz.sshj.SSHClient.authPassword(SSHClient.java:248)
	at net.schmizz.sshj.SSHClient.authPassword(SSHClient.java:232)
	at org.jclouds.sshj.SSHClientConnection.create(SSHClientConnection.java:165)
	at org.jclouds.sshj.SSHClientConnection.create(SSHClientConnection.java:49)
	at org.jclouds.sshj.SshjSshClient.acquire(SshjSshClient.java:195)
	... 6 more
Caused by: net.schmizz.sshj.userauth.UserAuthException: Timeout expired
	at net.schmizz.sshj.userauth.UserAuthException$1.chain(UserAuthException.java:33)
	at net.schmizz.sshj.userauth.UserAuthException$1.chain(UserAuthException.java:26)
	at net.schmizz.concurrent.Promise.retrieve(Promise.java:139)
	at net.schmizz.sshj.userauth.UserAuthImpl.authenticate(UserAuthImpl.java:69)
	at net.schmizz.sshj.SSHClient.auth(SSHClient.java:211)
	... 13 more
Caused by: java.util.concurrent.TimeoutException: Timeout expired
	... 16 more
```

nacx:
Thanks! At this point, I think we can merge this PR, as it fixes all the tests but this one (which I'd say it is not a blocker one). I've [JCLOUDS-1058](https://issues.apache.org/jira/browse/JCLOUDS-1058) to track this.

devcsrj:
Updated to fix compilation errors in [`ProfitBricksTemplateBuilderImpl`](https://github.com/jclouds/jclouds-labs/pull/224/files#diff-1500f3a9004863afdd8a7869407b390aR49) due to change in core

nacx:
Pushed to master as [fef0d4bd](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/fef0d4bd). Great work here! Thanks!

PR #225, opened 2015-12-11
Profitbricks REST -  Datacenter 

jasminSPC:
Thanks @mirza-spc


nacx:
rebuild please

mirza-spc:
rebuild please

profitbricks-rest/.gitattributes
```diff
@@ -0,0 +1,2 @@
+text eol=crlf
+*.java text
```
nacx: Move this file to the root of the repo

profitbricks-rest/pom.xml
```diff
@@ -0,0 +1,157 @@
...
+            <groupId>com.squareup.okhttp</groupId>
+            <artifactId>mockwebserver</artifactId>
+            <exclusions>
+                <!-- Already provided by jclouds-sshj -->
+                <exclusion>
+                    <groupId>org.bouncycastle</groupId>
+                    <artifactId>bcprov-jdk15on</artifactId>
+                </exclusion>
+            </exclusions>
+            <version>2.2.0</version>
```
nacx: The version should already be configured in the jclouds-project. Is it possible to remove it?

profitbricks-rest/pom.xml
```diff
@@ -0,0 +1,157 @@
...
+            <type>jar</type>
+        </dependency>
+        <dependency>
+           <groupId>org.apache.jclouds.driver</groupId>
+           <artifactId>jclouds-okhttp</artifactId>
+           <version>${project.parent.version}</version>
```
nacx: Use `jclouds.version` too.

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/ProfitBricksApi.java
```diff
@@ -0,0 +1,30 @@
...
+
+package org.apache.jclouds.profitbricks.rest;
+
+import com.google.common.annotations.Beta;
+import java.io.Closeable;
+import org.apache.jclouds.profitbricks.rest.features.DataCenterApi;
+import org.jclouds.rest.annotations.Delegate;
+
+@Beta
+public interface ProfitBricksApi extends Closeable{
```
nacx: Style minor: Add a space before `{`.

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/ProfitBricksApiMetadata.java
```diff
@@ -0,0 +1,92 @@
...
+   }
+
+   public static Properties defaultProperties() {
+      Properties properties = BaseHttpApiMetadata.defaultProperties();
+      /*properties.put("oauth.endpoint", "https://cloud.digitalocean.com/v1/oauth/token");
+       properties.put(JWS_ALG, "RS256");
+       properties.put(AUDIENCE, "https://cloud.digitalocean.com/v1/oauth/token");
+       properties.put(CREDENTIAL_TYPE, BEARER_TOKEN_CREDENTIALS.toString());
+       properties.put(PROPERTY_SESSION_INTERVAL, 3600);
+       properties.put(TEMPLATE, "osFamily=UBUNTU,os64Bit=true");*/
```
nacx: Remove the commented copy/pasted properties.

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/ProfitBricksApiMetadata.java
```diff
@@ -0,0 +1,92 @@
...
+      properties.put(POLL_INITIAL_PERIOD, 5000);
+      properties.put(POLL_MAX_PERIOD, 20000);
+      // Node operations in DigitalOcean can be quite slow. Use a 5 minutes
+      // timeout by default
+      properties.put(TIMEOUT_NODE_RUNNING, 300000);
+      properties.put(TIMEOUT_NODE_SUSPENDED, 300000);
+      properties.put(TIMEOUT_NODE_TERMINATED, 300000);
```
nacx: Also remove these defaults if they don't apply to this provider.

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/ProfitBricksApiMetadata.java
```diff
@@ -0,0 +1,92 @@
...
+      return properties;
+   }
+
+   public static class Builder extends BaseHttpApiMetadata.Builder<ProfitBricksApi, Builder> {
+
+      protected Builder() {
+         id("profitbricks-rest")
+            .name("Digital Ocean v2 API")
+            .identityName("Not used for OAuth")
+            .credentialName("Must be oauth2 Bearer Token")
```
nacx: Change these literals accordingly.

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/ProfitBricksProviderMetadata.java
```diff
@@ -0,0 +1,73 @@
...
+   public static Properties defaultProperties() {
+      Properties properties = ProfitBricksApiMetadata.defaultProperties();
+      return properties;
+   }
+
+   public static class Builder extends BaseProviderMetadata.Builder {
+
+      protected Builder() {
+         id("profitbricks-rest")
+            .name("DigitalOcean Compute Services")
```
nacx: Fix this literal.

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/config/ProfitBricksComputeProperties.java
```diff
@@ -0,0 +1,32 @@
...
+
+   public static final String POLL_TIMEOUT = "jclouds.profitbricks.poll.timeout";
+   public static final String POLL_PERIOD = "jclouds.profitbricks.operation.poll.initial-period";
+   public static final String POLL_MAX_PERIOD = "jclouds.profitbricks.operation.poll.max-period";
+
+   private ProfitBricksComputeProperties() {
+      throw new AssertionError("Intentionally unimplemented");
+   }
+
+}
```
nacx: Remove this class until these properties are actually needed.

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/domain/DataCenters.java
```diff
@@ -0,0 +1,40 @@
...
+
+   public abstract String href();
+
+   public abstract List<DataCenter> items();
+
+   @SerializedNames({"id", "type", "href", "items"})
+   public static DataCenters create(String id, String type, String href, List<DataCenter> items) {
+      return new AutoValue_DataCenters(id, type, href, items == null ? ImmutableList.<DataCenter>of() : copyOf(items));
+   }
+}
```
nacx: Do the id, type and href fields provide useful information? Can we get rid of this object and let the API just "select" the items field and directly return lists of DataCenter objects? It looks unnecessary to have this entity if the fields it adds provide no useful info.

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/domain/FirewallRule.java
```diff
@@ -0,0 +1,79 @@
...
+    public enum Protocol {
+
+        TCP, UDP, ICMP, ANY, UNRECOGNIZED;
+
+        public static Protocol fromValue(String value) {
+            try {
+                return valueOf(value);
+            } catch (IllegalArgumentException e) {
+                return UNRECOGNIZED;
+            }
```
nacx: Use Enums.getIfPresent like in the previous enum or change that one, but make the code consistent using the same pattern when addressing the same thing.

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/domain/FirewallRule.java
```diff
@@ -0,0 +1,79 @@
...
+
+        public abstract String sourceMac();
+
+        public abstract String sourceIp();
+
+        public abstract String targetIp();
+
+        public abstract String icmpCode();
+
+        public abstract String icmpType();
```
nacx: Are all these properties mandatory? Will they *always* have a value in the responses?

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/domain/FirewallRules.java
```diff
@@ -0,0 +1,41 @@
...
+   public abstract String href();
+
+   public abstract List<FirewallRule> items();
+
+   @SerializedNames({"id", "type", "href", "items"})
+   public static FirewallRules create(String id, String type, String href, List<FirewallRule> items) {
+      return new AutoValue_FirewallRules(id, type, href, items == null ? ImmutableList.<FirewallRule>of() : copyOf(items));
+   }
+
+}
```
nacx: Same here. Consider removing this wrapper class.

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/domain/Lan.java
```diff
@@ -0,0 +1,70 @@
...
+   public abstract String href();
+
+   @Nullable
+   public abstract Metadata metadata();
+
+   @Nullable
+   public abstract Properties properties();
+
+   @Nullable
+   public abstract Entities entities();
```
nacx: These fields are not nullable in the previous classes but nullable here. It it the real behaviour of the API? Or should the corresponding `@Nullable` annotations be added to the other objects?

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/domain/Lans.java
```diff
@@ -0,0 +1,40 @@
...
+
+   public abstract String href();
+
+   public abstract List<Lan> items();
+
+   @SerializedNames({"id", "type", "href", "items"})
+   public static Lans create(String id, String type, String href, List<Lan> items) {
+      return new AutoValue_Lans(id, type, href, items == null ? ImmutableList.<Lan>of() : copyOf(items));
+   }
+}
```
nacx: Same about wrapper objects.

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/domain/Loadbalancer.java
```diff
@@ -0,0 +1,61 @@
...
+      @SerializedNames({"name", "ip", "dhcp"})
+      public static Properties create(String name, String ip, boolean dhcp) {
+	 return new AutoValue_Loadbalancer_Properties(name, ip, dhcp);
+      }
+   }
+
+   @AutoValue
+   public abstract static class Entities {
+
+   }
```
nacx: Remove this class if empty.

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/domain/Loadbalancers.java
```diff
@@ -0,0 +1,40 @@
...
+
+   public abstract String href();
+
+   public abstract List<Loadbalancer> items();
+
+   @SerializedNames({"id", "type", "href", "items"})
+   public static Loadbalancers create(String id, String type, String href, List<Loadbalancer> items) {
+      return new AutoValue_Loadbalancers(id, type, href, items == null ? ImmutableList.<Loadbalancer>of() : copyOf(items));
+   }
+}
```
nacx: Remove wrapper.

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/domain/Location.java
```diff
@@ -0,0 +1,51 @@
...
+ */
+package org.apache.jclouds.profitbricks.rest.domain;
+
+public enum Location {
+
+   DE_FKB("de/fkb"),
+   DE_FRA("de/fra"),
+   US_LAS("us/las"),
+   US_LAS_DEV("us/lasdev"),
+   UNRECOGNIZED("unknown");
```
nacx: If the available locations are known in advance, they should be added to the provider metadata default properties. You can take the [Rackspace ProviderMetadata](https://github.com/jclouds/jclouds/blob/master/providers/rackspace-cloudservers-us/src/main/java/org/jclouds/rackspace/cloudservers/us/CloudServersUSProviderMetadata.java#L66-L77) as an example. Also configure the [iso codes](https://github.com/jclouds/jclouds/blob/master/providers/rackspace-cloudservers-us/src/main/java/org/jclouds/rackspace/cloudservers/us/CloudServersUSProviderMetadata.java#L103).

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/domain/Metadata.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+package org.apache.jclouds.profitbricks.rest.domain;
+
+import com.google.auto.value.AutoValue;
+import org.jclouds.json.SerializedNames;
+
+@AutoValue
+public abstract class Metadata {
+
+    public abstract String createdDate();
```
nacx: Change type to Date.

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/domain/Metadata.java
```diff
@@ -0,0 +1,41 @@
...
+
+    public abstract String createdBy();
+
+    public abstract String etag();
+
+    public abstract String lastModifiedDate();
```
nacx: Change type to Date.

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/domain/Nics.java
```diff
@@ -0,0 +1,40 @@
...
+
+   public abstract String href();
+
+   public abstract List<Nic> items();
+
+   @SerializedNames({"id", "type", "href", "items"})
+   public static Nics create(String id, String type, String href, List<Nic> items) {
+      return new AutoValue_Nics(id, type, href, items == null ? ImmutableList.<Nic>of() : copyOf(items));
+   }
+}
```
nacx: Remove wrapper.

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/domain/DataCenters.java
nacx: I've just noticed wrappers might be needed to deserialize the "embedded entities". Is that its purpose?

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/features/DataCenterApi.java
```diff
@@ -0,0 +1,156 @@
...
+   
+   @Named("datacenter:list")
+   @GET
+   @Produces("application/vnd.profitbricks.resource+json")
+   @SelectJson("items")
+   @Fallback(EmptyListOnNotFoundOr404.class)
+   List<DataCenter> list(DepthOptions options);
+   
+   /**
+    * @param identifier Data Center identifier
```
nacx: `@param id`.

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/features/DataCenterApi.java
```diff
@@ -0,0 +1,156 @@
...
+   @Produces("application/vnd.profitbricks.resource+json")
+   @Fallback(NullOnNotFoundOr404.class)
+   DataCenter getDataCenter(@PathParam("id") String id, DepthOptions options);
+   
+   @Named("datacenter:create")
+   @POST
+   @ResponseParser(DataCenterParser.class)
+   @Produces("application/vnd.profitbricks.resource+json")
+   @MapBinder(DataCenterCreateMapBinder.class)
+   @Fallback(NullOnNotFoundOr404.class)
```
nacx: Do not use 404 fallbacks on POST/PUT. Just remove the fallback and let the method throw an exception if a 404 occurs.

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/features/DataCenterApi.java
```diff
@@ -0,0 +1,156 @@
...
+      @PayloadParam("location") String location
+   );
+
+   @Named("datacenter:update")
+   @PATCH
+   @Path("/{id}")
+   @ResponseParser(DataCenterParser.class)
+   @Produces("application/vnd.profitbricks.partial-properties+json")
+   @MapBinder(BindToJsonPayload.class)
+   @Fallback(NullOnNotFoundOr404.class)
```
nacx: Remove the fallback here too.

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/DataCenterApiLiveTest.java
```diff
@@ -0,0 +1,112 @@
...
+   }
+   
+   @Test
+   public void testDepth() {
+      
+      DataCenter dataCenter = createDataCenter();
+      
+      for (int i = 1; i <= 5; ++i) {
+         dataCenter = getDataCenter(dataCenter.id(), new DepthOptions().depth(i));
+         assertNotNull(dataCenter);
```
nacx: Verify that the depth actually changes something in the returned datanceter? Verify the entities?

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/DataCenterApiLiveTest.java
```diff
@@ -0,0 +1,112 @@
...
+   
+   private void deleteDataCenter(String id) {
+      dataCenterApi().delete(id);
+   }
+   
+   private DataCenterApi dataCenterApi() {
+      return api.dataCenterApi();
+   }
+   
+}
```
nacx: The test looks OK. Take into account that tests will take a while when the provider is complete, so you should consider chaining the tests and avoid creating a datacenter for each test if it is an expensive operation.

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/DataCenterApiMockTest.java
```diff
@@ -0,0 +1,115 @@
...
+   public void testGetList() {
+      server.enqueue(
+         new MockResponse().setBody(stringFromResource("/datacenter/list.json"))
+      );
+      
+      List<DataCenter> list = dataCenterApi().list(new DepthOptions().depth(1));
+      
+      assertNotNull(list);
+      assertEquals(list.size(), 3);
+      assertEquals(list.get(0).properties().name(), "vea");
```
nacx: You need to verify the recorded request. MockWebServer will return each enqueued request upon *any* request. You need to verify that the generated request is the expected one: expected request uri, method, headers and body.

The same must be applied in all methods in this class.

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/DataCenterApiMockTest.java
```diff
@@ -0,0 +1,115 @@
...
+      response.setBody(stringFromResource("/datacenter/get.json"));
+      response.setHeader("Content-Type", "application/vnd.profitbricks.resource+json");
+      
+      server.enqueue(response);
+      
+      DataCenter dataCenter = dataCenterApi().getDataCenter("some-id");
+      
+      assertNotNull(dataCenter);
+      assertEquals(dataCenter.properties().name(), "docker");
+   }
```
nacx: Add a mock test for the get + 404, as you define a fallback for that case.

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/DataCenterApiMockTest.java
```diff
@@ -0,0 +1,115 @@
...
+      server.enqueue(
+         new MockResponse().setBody(stringFromResource("/datacenter/list.json"))
+      );
+      
+      List<DataCenter> list = dataCenterApi().list(new DepthOptions().depth(1));
+      
+      assertNotNull(list);
+      assertEquals(list.size(), 3);
+      assertEquals(list.get(0).properties().name(), "vea");
+   }
```
nacx: Also test the list method that does not get the options parameter.

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/DataCenterApiMockTest.java
```diff
@@ -0,0 +1,115 @@
...
+   
+   @Test
+   public void testDelete() {
+      server.enqueue(
+         new MockResponse().setBody("")
+      );
+      
+      dataCenterApi().delete("some-id");
+   }
```
nacx: Add a test for the delete fallback.

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/internal/BaseProfitBricksApiMockTest.java
```diff
@@ -0,0 +1,139 @@
...
+import org.jclouds.rest.ApiContext;
+import static org.testng.Assert.assertEquals;
+import org.testng.annotations.AfterMethod;
+import org.testng.annotations.BeforeMethod;
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.collect.Iterables.getOnlyElement;
+
+public class BaseProfitBricksApiMockTest {
+
+   private static final String MOCK_BEARER_TOKEN = "c5401990f0c24135e8d6b5d260603fc71696d4738da9aa04a720229a01a2521d";
```
nacx: This does not apply here.

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/internal/BaseProfitBricksApiMockTest.java
```diff
@@ -0,0 +1,139 @@
...
+
+   // So that we can ignore formatting.
+   private final JsonParser parser = new JsonParser();
+
+   @BeforeMethod
+   public void start() throws IOException {
+      server = new MockWebServer();
+      server.play();
+      ApiContext<ProfitBricksApi> ctx = ContextBuilder.newBuilder("profitbricks-rest")
+	      .credentials("", MOCK_BEARER_TOKEN)
```
nacx: The provider uses basic auth. Use appropriate mock values; not an oauth token.

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/internal/BaseProfitBricksApiMockTest.java
```diff
@@ -0,0 +1,139 @@
...
+      String text = stringFromResource(resourceName);
+      return json.fromJson(text, type);
+   }
+
+   protected RecordedRequest assertSent(MockWebServer server, String method, String path) throws InterruptedException {
+      RecordedRequest request = server.takeRequest();
+      assertEquals(request.getMethod(), method);
+      assertEquals(request.getPath(), path);
+      assertEquals(request.getHeader("Accept"), "application/json");
+      assertEquals(request.getHeader("Authorization"), "Bearer " + MOCK_BEARER_TOKEN);
```
nacx: You will probably need to change this auth and the accept type when completing teh request verification in the mock tests.

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/internal/BaseProfitBricksLiveTest.java
```diff
@@ -0,0 +1,42 @@
...
+      provider = "profitbricks-rest";
+   }
+  
+   @Override
+   protected ProfitBricksApi create(Properties props, Iterable<Module> modules) {
+      
+      Injector injector = newBuilder().modules(modules).overrides(props).buildInjector();
+      
+      return injector.getInstance(ProfitBricksApi.class);
+   }
```
nacx: You shouldn't have to override this method just to get the api.

nacx:
Thanks @jasminSPC! This is a good start. Apart from the inline comments, there are two missing pieces for the initial structure of the provider:

* The creation of the ProfitBricksHttpApiModule (declare it too in the api metadata default module list).
* The creating of the ProfitBricksErrorHandler (you will configure it in the http api module) that transforms each failed response in the appropriate Java exception.

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/domain/Lan.java
mirza-spc: My knowledge is that, whether some field is present or not, is usually controlled by the depth param. So, as you can imagine, some fields may be existent, but if depth is too shallow they may be 'null'.

So what we've done in the depth testing, is to pull a response for each of the depth levels and simply try to instantiate a DataCenter object (which contains almost all the other domain objects). If we see the test failing (because of null entry), we put nullable on the field.

Now one may argue, that the field can still be null (in the server database), and it would be correct thing to say. To check this, we can go and try to instantiate each object with null on these fields or simply check in the UI if it can be empty.

nacx: Ok, Then is would make sense to mark the `Entities` properties as `@Nullable` in all objects. Otherwise deserialization will fail if the annotation is missing and the field does not come in the response.

What about the metadata and properties fields? Shouldn't be always present?

mirza-spc: The tree of hierarchy, with DataCenter as its root is 'filtered' by the depth level. So I think it will be always there for DataCenter, since it is the root. If however, at some point in future, there would be a grouping above datacenters (perhaps Region?), and if the depth is too low, then it could become null.

It is in a sense relative, if one is requesting an object from the api. Then the object's meta will not be null, but its childrens' meta may be, depending on the depth.

I think it would be sensible to have them nullable as well.

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/domain/Metadata.java
mirza-spc: These were date before, but test keep failing because of a parse error:

```
java.lang.IllegalArgumentException: Error parsing data at 0
	at java.text.DateFormat.parse(DateFormat.java:366)
	at org.jclouds.date.internal.SimpleDateFormatDateService.cDateParse(SimpleDateFormatDateService.java:80)
```

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/features/DataCenterApi.java
mirza-spc: I'm guessing this goes also for PATCH and DELETE

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/domain/Metadata.java
nacx: In which format are dates returned? This should be fixed by registering a proper date adapter in the http api module. Take a look at how [the Chef api](https://github.com/jclouds/jclouds/blob/master/apis/chef/src/main/java/org/jclouds/chef/config/ChefParserModule.java#L259) does that.

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/features/DataCenterApi.java
nacx: Just POST/PUT/PATCH, because the failure is ambiguous. What does a null value mean in that case? The resource does not exist? The operation completed and the modified Object is now `null` (the response did not return a body)? These semantics with null are particularly ambiguous in POST/PUT/PATCH, so we avoid the fallbacks and let the method fail.

In DELETE we can be less restrictive, as it could make sense to return null. The object being deleted is no longer there, so there are no conceptual conflicts.

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/DataCenterApiLiveTest.java
mirza-spc: Sure, but I wouldn't go to far in, as we would need to have all the sub-items instantiated such as servers, lans.. and so on. (although we can come back to it, when these other parts of API are implemented)

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/DataCenterApiLiveTest.java
mirza-spc: I agree. As far as I can see, a DataCenter is purely a logical construct and does not consume any real resources. But definitely for servers, volumes and so on.

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/DataCenterApiLiveTest.java
nacx: Agree. There's no need for a depth 5. I'd just call depth(1) and depth(2) (or 0 and 1, whatever is cheaper) and verify the different content in the returned objects.

mirza-spc:
rebuild please

mirza-spc:
I have tried to apply all the comments so far. When you are able to, you can review the changes.

mirza-spc:
rebuild please

nacx:
Thanks! I'll have a look later today/tomorrow.

For the next comments in this PR (and for upcoming new PRs), please don't squash the commits when addressing comments. It makes it a lot easier to review only the bits that have changed; otherwise we have to go through the entire PR again. We usually follow this procedure and once the PR is ready to merge, we ask to squash teh commits or squash them ourselves before merging.

NP for now :) I'm just suggesting this for the next review cycles.

mirza-spc:
rebuild please

mirza-spc:
rebuild please

mirza-spc:
rebuild please

profitbricks-rest/pom.xml
```diff
@@ -0,0 +1,156 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.apache.jclouds.driver</groupId>
+            <artifactId>jclouds-sshj</artifactId>
+            <version>${jclouds.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.squareup.okhttp</groupId>
+            <artifactId>mockwebserver</artifactId>
```
nacx: Declare this dependency in test scope.

profitbricks-rest/pom.xml
```diff
@@ -0,0 +1,156 @@
...
+                <exclusion>
+                    <groupId>org.bouncycastle</groupId>
+                    <artifactId>bcprov-jdk15on</artifactId>
+                </exclusion>
+            </exclusions>
+            <type>jar</type>
+        </dependency>
+        <dependency>
+           <groupId>org.apache.jclouds.driver</groupId>
+           <artifactId>jclouds-okhttp</artifactId>
```
nacx: Move this dependency up, "outside" the test dependencies block.

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/ProfitBricksProviderMetadata.java
```diff
@@ -0,0 +1,82 @@
...
+      super(builder);
+   }
+
+   public static Properties defaultProperties() {
+      Properties properties = ProfitBricksApiMetadata.defaultProperties();
+      properties.setProperty(PROPERTY_REGIONS, "DE,US");
+      properties.setProperty(PROPERTY_REGION + ".DE.FKB." + ISO3166_CODES, "DE-FKB");
+      properties.setProperty(PROPERTY_REGION + ".DE.FRA." + ISO3166_CODES, "DE-FRA");
+      properties.setProperty(PROPERTY_REGION + ".US.LAS." + ISO3166_CODES, "US-LAS");
+      properties.setProperty(PROPERTY_REGION + ".US.LASDEV." + ISO3166_CODES, "US-LASDEV");
```
nacx: It looks like the provider has zones and regions. I assume the region IDs are "DE" and "US", and the zone IDs are "FKB, "FRA", "LAS". The right way to configure this in jclouds is as follows:

```java
properties.setProperty(PROPERTY_ISO3166_CODES, "US-NV,DE-HE,DE-BW");

properties.setProperty(PROPERTY_REGIONS, "DE,US");
properties.setProperty(PROPERTY_REGION + ".DE.zones", "FKB,FRA")
properties.setProperty(PROPERTY_REGION + ".US.zones", "LAS");

properties.setProperty(PROPERTY_ZONES, "FKB,FRA,LAS");
properties.setProperty(PROPERTY_ZONE + ".FKB." + ISO3166_CODES, "DE-BW");
properties.setProperty(PROPERTY_ZONE + ".FRA." + ISO3166_CODES, "DE-HE");
properties.setProperty(PROPERTY_ZONE + ".LAS." + ISO3166_CODES, "US-NV");
```

Note that I've already fixed the ISO codes to use the right ones. Also, the zone IDs and regions IDs used here are up to you. Feel free to change them to something more user-friendly, as those "FKB/FRA/LAS" is what users will use as the "locationId" when choosing where they want to deploy the VMs.

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/ProfitBricksProviderMetadata.java
```diff
@@ -0,0 +1,82 @@
...
+
+   public static class Builder extends BaseProviderMetadata.Builder {
+
+      protected Builder() {
+         id("profitbricks-rest")
+            .name("ProfitBricks REST Compute")
+            .apiMetadata(new ProfitBricksApiMetadata())
+            .homepage(URI.create("https://www.profitbricks.com/"))
+            .console(URI.create("https://my.profitbricks.com/dashboard/dcdr2"))
+            .iso3166Codes("DE-FKB", "DE-FRA", "US-LAS", "US-LASDEV")
```
nacx: Fix this according to the previous comment.

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/domain/LicenceType.java
```diff
@@ -0,0 +1,31 @@
...
+public enum LicenceType {
+
+   WINDOWS, LINUX, OTHER, UNRECOGNIZED;
+
+   public static LicenceType fromValue(String v) {
+      try {
+         return valueOf(v);
+      } catch (IllegalArgumentException ex) {
+         return UNRECOGNIZED;
+      }
```
nacx: Use the same approach that in the other enums.

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/domain/Location.java
```diff
@@ -0,0 +1,51 @@
...
+   public String value() {
+      return id;
+   }
+
+   public static Location fromValue(String v) {
+      try {
+	 return valueOf(v);
+      } catch (IllegalArgumentException ex) {
+	 return UNRECOGNIZED;
+      }
```
nacx: Same here. Use a consistent approach in all enums.

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/DataCenterApiMockTest.java
```diff
@@ -0,0 +1,155 @@
...
+      assertEquals(server.getRequestCount(), 1);
+      assertSent(server, "GET", "/datacenters/some-id");
+   }
+   
+   public void testGetDataCenterWith404() throws InterruptedException {
+      server.enqueue(response404());
+
+      DataCenter dataCenter = dataCenterApi().getDataCenter("some-id");
+      
+      assertEquals(dataCenter, null);
```
nacx: [minor] Use assertNull

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/features/DataCenterApi.java
```diff
@@ -0,0 +1,154 @@
...
+import org.jclouds.rest.annotations.SelectJson;
+import org.jclouds.rest.binders.BindToJsonPayload;
+
+@Path("/datacenters")
+@RequestFilters(BasicAuthentication.class)
+public interface DataCenterApi extends Closeable {
+
+   @Named("datacenter:list")
+   @GET
+   @Produces("application/vnd.profitbricks.resource+json")
```
nacx: Remove this annotation if this method is not sending this content-type, and apply the same for all methods that do not generate a body.

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/internal/BaseProfitBricksApiMockTest.java
```diff
@@ -0,0 +1,147 @@
...
+      Map<String, T> object = json.fromJson(text, type.getType());
+      checkArgument(!object.isEmpty(), "The given json does not contain any object: %s", text);
+      checkArgument(object.keySet().size() == 1, "The given json does not contain more than one object: %s", text);
+      return object.get(getOnlyElement(object.keySet()));
+   }
+
+   protected <T> T objectFromResource(String resourceName, Class<T> type) {
+      String text = stringFromResource(resourceName);
+      return json.fromJson(text, type);
+   }
```
nacx: Remove all protected methods that are not used.

nacx:
Nice update! It looks pretty good now. There are only a few minors.
Apart from that, looking at how the ProfitBricks API returns the locations (lowercase region/zone), we'll have to configure some extra bits when it comes to implement the ComputeService abstraction (as the default jclouds implementation will just use the zone id without the "region/" prefix), but that is an easy change we'll do when we get there.

Nice job!

mirza-spc:
rebuild please

nacx:
Thanks @mirza-spc! Just three small comments left!

nacx:
Amended the commit message and pushed to master as [d63af3c2](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/d63af3c2).
Thanks @jasminSPC and @mirza-spc!

PR #230, opened 2016-01-19
Improved DigitalOcean image extension

nacx:
Do not block when waiting for the images to become available. The image extension returns a Future when creating images. Move all the logic that takes time to the Future to avoid blocking.


nacx:
Pushed to master as [1c5df25c](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/1c5df25c).

PR #232, opened 2016-01-21
[JCLOUDS-1063] make Role.OSVirtualHardDisk.sourceImageName nullable

Ladicek:


nacx:
Add a test to make sure this is fixed, or modify [the existing one](https://github.com/jclouds/jclouds-labs/blob/master/azurecompute/src/test/java/org/jclouds/azurecompute/xml/RoleHandlerTest.java)?

Ladicek:
Done. I added the exact XML received from Azure (actually the _Role_ part of the XML) that caused the problem. Some fields redacted, of course.

nacx:
Pushed to master as [6d3363fd](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/6d3363fd). Thanks!

PR #233, opened 2016-02-10
Profitbricks REST - Volume API

mirza-spc:
@jasminSPC


profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/binder/volume/CreateVolumeRequestBinder.java
```diff
@@ -0,0 +1,73 @@
...
+         properties.put("imagePassword", payload.imagePassword());
+      
+      if (payload.image() != null)
+         properties.put("image", payload.image());
+      else if (payload.licenceType() != null)
+         properties.put("licenceType", payload.licenceType());
+      
+      requestBuilder.put("properties", properties);
+      
+      return (new Gson()).toJson(requestBuilder);
```
devcsrj: Use the `Json` object, as you did with [here](https://github.com/jclouds/jclouds-labs/blob/master/profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/binder/server/AttachCdromRequestBinder.java#L52).

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/binder/volume/UpdateVolumeRequestBinder.java
```diff
@@ -0,0 +1,61 @@
...
+      if (payload.name() != null)
+         requestBuilder.put("name",  payload.name());
+      
+      if (payload.size() != null)
+         requestBuilder.put("size",  payload.size());
+      
+      if (payload.bus() != null)
+         requestBuilder.put("bus",  payload.bus());
+      
+      return (new Gson()).toJson(requestBuilder);
```
devcsrj: Use `Json` here too.

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/features/VolumeApi.java
```diff
@@ -19,20 +19,118 @@
...
+   @Named("volume:snapshot:create")
+   @POST
+   @MapBinder(CreateSnapshotRequestBinder.class)
+   @ResponseParser(SnapshotApi.SnapshotParser.class)
+   Snapshot createSnapshot(@PayloadParam("snapshot") Volume.Request.CreateSnapshotPayload payload);
+   
+   @Named("volume:snapshot:restore")
+   @POST
+   @MapBinder(RestoreSnapshotRequestBinder.class)
+   @Fallback(Fallbacks.VoidOnNotFoundOr404.class)
```
devcsrj: Remove fallback. Let's let the user know why the *restore* op failed

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/ServerApiMockTest.java
```diff
@@ -65,7 +65,7 @@ public void testGetListWithDepth() throws InterruptedException {
    
    @Test
    public void testGetListWith404() throws InterruptedException {
```
devcsrj: You may remove the test for 404 fallback, as this is already tested in the core.

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/VolumeApiMockTest.java
```diff
@@ -0,0 +1,186 @@
...
+import org.apache.jclouds.profitbricks.rest.domain.LicenceType;
+import org.apache.jclouds.profitbricks.rest.domain.Volume;
+import org.apache.jclouds.profitbricks.rest.domain.options.DepthOptions;
+import org.apache.jclouds.profitbricks.rest.internal.BaseProfitBricksApiMockTest;
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertNotNull;
+import static org.testng.Assert.assertTrue;
+import org.testng.annotations.Test;
+
+@Test(groups = "unit", testName = "VolumeApiMockTest", singleThreaded = true)
```
devcsrj: Hmm. Do you need this test to be `singleThreaded`?

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/VolumeApiMockTest.java
```diff
@@ -0,0 +1,186 @@
...
+      assertNotNull(list);
+      assertEquals(list.size(), 6);
+      assertEquals(list.get(0).properties().name(), "Docker Registry Volume");
+      
+      assertEquals(server.getRequestCount(), 1);
+      assertSent(server, "GET", "/datacenters/datacenter-id/volumes");
+   }
+   
+   @Test
+   public void testGetListWith404() throws InterruptedException {
```
devcsrj: No need to test this (already expected to return an empty list, because of the declared, already-tested fallback)

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/VolumeApiMockTest.java
```diff
@@ -0,0 +1,186 @@
...
+      Volume volume = volumeApi().getVolume("datacenter-id", "some-id");
+      
+      assertNotNull(volume);
+      assertEquals(volume.properties().name(), "Docker Registry Volume");
+      
+      assertEquals(server.getRequestCount(), 1);
+      assertSent(server, "GET", "/datacenters/datacenter-id/volumes/some-id");
+   }
+   
+   public void testGetVolumeWith404() throws InterruptedException {
```
devcsrj: No need for this too.

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/VolumeApiMockTest.java
```diff
@@ -0,0 +1,186 @@
...
+         new MockResponse().setBody("")
+      );
+      
+      volumeApi().deleteVolume("datacenter-id", "some-id");
+      assertEquals(server.getRequestCount(), 1);
+      assertSent(server, "DELETE", "/datacenters/datacenter-id/volumes/some-id");
+   }
+   
+   @Test
+   public void testDeleteWith404() throws InterruptedException {
```
devcsrj: Same as above

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/internal/BaseProfitBricksApiMockTest.java
```diff
@@ -89,13 +89,13 @@ protected MockResponse response204() {
    protected MockResponse response404() {
```
devcsrj: No need for this then. :) Also, remove 404 tests (or Unauthorized, if exists) from other tests, if possible.

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/util/MacAddresses.java
```diff
@@ -0,0 +1,29 @@
...
+package org.apache.jclouds.profitbricks.rest.util;
+
+import java.util.regex.Pattern;
+
+public class MacAddresses {
+
+   private static final String MAC_ADDR_FORMAT = "^([0-9a-f]{2}[:]){5}([0-9a-f]{2})$";
+   private static final Pattern MAC_ADDR_PATTERN = Pattern.compile(MAC_ADDR_FORMAT);
+
+   public static boolean isMacAddress(String in) {
```
devcsrj: Add missing test

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/util/Passwords.java
```diff
@@ -0,0 +1,64 @@
...
+ * limitations under the License.
+ */
+package org.apache.jclouds.profitbricks.rest.util;
+
+import java.util.Random;
+import java.util.regex.Pattern;
+
+import com.google.common.collect.ImmutableSet;
+
+public class Passwords {
```
devcsrj: Add test for this too, so we can validate it's [usage](https://github.com/jclouds/jclouds-labs/pull/233/files#diff-6d0cd6b1064677eca969d404ba5c0a91R175).

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/features/VolumeApi.java
```diff
@@ -19,20 +19,118 @@
...
+   @DELETE
+   @Path("/{volumeId}")
+   @Fallback(Fallbacks.VoidOnNotFoundOr404.class)
+   void deleteVolume(@PathParam("dataCenterId") String dataCenterId, @PathParam("volumeId") String volumeId);
+   
+   @Named("volume:snapshot:create")
+   @POST
+   @MapBinder(CreateSnapshotRequestBinder.class)
+   @ResponseParser(SnapshotApi.SnapshotParser.class)
+   Snapshot createSnapshot(@PayloadParam("snapshot") Volume.Request.CreateSnapshotPayload payload);
```
devcsrj: Hmm. Should this (and `restoreSnapshot`) be in `SnapshotApi.java` instead? If so, don't forget to move both of their [tests](https://github.com/jclouds/jclouds-labs/pull/233/files#diff-81aadea0059ec0e66cf6304f6234762bR148) too!

devcsrj:
Hi @mirza-spc !

Did initial review for the PR, while waiting for others to jump in. :) In addition to the comments above, could you add a test for [`ParseId`](https://github.com/StackPointCloud/jclouds-labs/commit/60869d038e8022165b6b1c32cb80233cee213033#diff-640a66e9fc832b6c994d936877e933caR29)? This way, we're confident that the `#parseId` method actually returns what we expect. And, can you move the [`Pattern.compile(..)`](https://github.com/StackPointCloud/jclouds-labs/commit/60869d038e8022165b6b1c32cb80233cee213033#diff-640a66e9fc832b6c994d936877e933caR43) as a class level [property](https://github.com/StackPointCloud/jclouds-labs/commit/60869d038e8022165b6b1c32cb80233cee213033#diff-640a66e9fc832b6c994d936877e933caR32), so that the pattern is only compiled once (and not everytime the method is called).

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/VolumeApiMockTest.java
mirza-spc: yes, because of server.enqueue

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/features/VolumeApi.java
mirza-spc: I can see your point, but then again a Volume is needed as well. Also it's under volumes at https://devops.profitbricks.com/api/rest/#volumes

mirza-spc:
`Pattern.compile(String.format("..%s..", prefix))` uses a method param "prefix", so I don't think that I could move it to a static variable. This is unless there is some way to do this, of which I'm not aware.

mirza-spc:
Also since ParseId test would require injected ParseId (because of Json), I'm wondering how I can achieve that.

nacx:
> Also since ParseId test would require injected ParseId (because of Json), I'm wondering how I can achieve that.

You can create a small Guice context, with the json configuration, for that test:

```java
Injector injector = Guice.createInjector(new GsonModule());
ParseId parseId = injector.getInstance(ParseId.class);
```

nacx:
Thanks for the fixes @mirza-spc!
@devcsrj Wanna review the fixes before squashing the commits?

devcsrj:
Thanks for the `injector`-approach @nacx !

@mirza-spc I noticed that there are missing test in the [`-Binder`](https://github.com/StackPointCloud/jclouds-labs/tree/profitbricks-volume-api/profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/binder) classes. Could you add tests for these so that we can verify as to whether those binders actually build the correct payload? (I for one usually *trip* on these types, like typos in property names, etc. :) ) After that, this PR looks good to me!

mirza-spc:
Seems like the tests are failing because of the different ordering of keys in jsonBinder.toJson method. Any thoughts on this?

nacx:
You can start from an object for the expected payload and serialize it to get the expected payload string. This way you should be able to avoid ordering issues caused by different JVM defaults.

devcsrj:
Binder tests, :heavy_check_mark: ! Unless @nacx has further comments, this LGTM!

nacx:
@mirza-spc mind squashing the commits?

mirza-spc:
So I merged the upstream. However if I squash it as well, the conflict with the upstream will arise again. So I've left it be for now.

devcsrj:
Hmm. Unfortunately, this branch needs to be rebased and squashed, such that it contains one commit with changes only related to this PR. You'll notice that old files, such as [this](https://github.com/jclouds/jclouds-labs/pull/233/files#diff-5f712604f84e8ddc5e1fb9ca7f759697R51) no longer exists in `jclouds/jclouds-labs`, but is existing here. 

I'll try to rebase later so that this gets merged, and we can proceed to #241 .

devcsrj:
Pushed to master as [9a8b346](https://github.com/jclouds/jclouds-labs/commit/9a8b346cafd9e332b0e532f1341e739d0388a347). Thanks for the effort, @mirza-spc !

nacx:
Just noticed this while reviewing the last commit. Is the [commons-lang](https://github.com/jclouds/jclouds-labs/blob/master/profitbricks-rest/pom.xml#L68-L74) dependency really needed?

nacx:
And even if it is needed we can't rely on a snapshot dependency, to it must be removed.

devcsrj:
Woops, missed that. Pushed a [commit](https://github.com/jclouds/jclouds-labs/commit/5bd8447b28fe8b0b7a2b85b03270de643a6a62c0) to remove the dep.

jasminSPC:
Is this completed? Can we rebase #241?

devcsrj:
Yes! This has been pushed to master. You may now rebase the other PR.

PR #234, opened 2016-02-13
JCLOUDS-1076: Initial implementation of Keys API

cdancy:
Initial implementation surrounding the etcd keys API. This initial commit is purposely leaving out tests so as to give reviewers less to chew on. Not too much special is going on here more just laying the groudwork the API itself and future commits to come.

@nacx you were working with me here before, maybe you could look things over? The idea is to get this base commit, with tests once things look good, added and then send in additional PR's for fleshing out this API. 

And my apologies AOT for not finishing this work months ago. I bought a new house along with a new baby so things have been a bit busy around these parts. Looking to get back into the swing of things.


etcd/src/main/java/org/jclouds/etcd/domain/keys/Node.java
```diff
@@ -0,0 +1,53 @@
...
+import com.google.auto.value.AutoValue;
+
+@AutoValue
+public abstract class Node {
+
+	public abstract int createdIndex();
+	
+	public abstract boolean dir();
+	
+	@Nullable public abstract List<Node> nodes();
```
nacx: Can we make this non-nullable? We should use empty lists instead of null ones where possible.

etcd/src/main/java/org/jclouds/etcd/domain/keys/Node.java
```diff
@@ -0,0 +1,53 @@
...
+   @Nullable public abstract String ttl();
+	
+   @Nullable public abstract String value();
+		
+	Node() {
+	}
+	
+	@SerializedNames({"createdIndex", "dir", "nodes", "expiration", "key", "modifiedIndex", "ttl", "value"})
+	public static Node create(int createdIndex, boolean dir, List<Node> nodes, String expiration, String key, int modifiedIndex, String ttl, String value) {
+		return new AutoValue_Node(createdIndex, dir, nodes, expiration, key, modifiedIndex, ttl, value);
```
nacx: Make lists immutable.

etcd/src/main/java/org/jclouds/etcd/fallbacks/EtcdFallbacks.java
```diff
@@ -35,4 +35,15 @@ public Boolean createOrPropagate(Throwable t) throws Exception {
          throw propagate(t);
       }
    }
+   
+   public static final class NullOnKeyNonFoundAnd404 implements Fallback<Object> {
+      public Object createOrPropagate(Throwable t) throws Exception {
+         if (checkNotNull(t, "throwable") != null && 
+               t.getMessage().contains("Could not find key") && 
+               returnValueOnCodeOrNull(t, true, equalTo(404)) != null) {
```
nacx: Use `HttpUtils.contains404(t)`

etcd/src/main/java/org/jclouds/etcd/features/KeysApi.java
```diff
@@ -0,0 +1,60 @@
...
+@Consumes(MediaType.APPLICATION_JSON)
+@Path("/{jclouds.api-version}/keys")
+public interface KeysApi {
+
+	@Named("keys:create")
+	@PUT
+	@Path("/{key}")
+	Key createKey(@PathParam("key") String key, 
+	         @FormParam("value") String value, 
+	         @Nullable @FormParam("ttl") Integer seconds);
```
nacx: It would look cleaner to have an overloaded method, one without TTL and another with an `int` (primitive) TTL.

etcd/src/main/java/org/jclouds/etcd/features/KeysApi.java
```diff
@@ -0,0 +1,60 @@
...
+
+   @Named("keys:get-key")
```
nacx: Use just `keys:get`, to follow the convention used in the create method (and this is already prefixed as key).

etcd/src/main/java/org/jclouds/etcd/features/KeysApi.java
```diff
@@ -0,0 +1,60 @@
...
+   @GET
+   @Path("/{key}")
+   @Fallback(NullOnKeyNonFoundAnd404.class)
+   Key getKey(@PathParam("key") String key, 
+            @Nullable @QueryParam("recursive") Boolean recursive,
+            @Nullable @QueryParam("sorted") Boolean sorted);
```
nacx: Same as before. Allowing to pass those `(foo, null, null)` is kinda ugly. Consider providing an overloaded version that uses primitive types for the options.

etcd/src/main/java/org/jclouds/etcd/features/KeysApi.java
```diff
@@ -0,0 +1,60 @@
...
+   
+   @Named("keys:delete-key")
```
nacx: Same about naming.

etcd/src/main/java/org/jclouds/etcd/handlers/EtcdErrorHandler.java
```diff
@@ -54,7 +55,17 @@ public void handleError(HttpCommand command, HttpResponse response) {
...
+               if (command.getCurrentRequest().getMethod().equals("DELETE")) {
+                  if (command.getCurrentRequest().getEndpoint().getPath().startsWith("/" + EtcdApiMetadata.API_VERSION + "/keys") &&
+                        message.contains("Key not found")) {
+                        exception = new ResourceNotFoundException("Could not find key to delete: message=" + message, new HttpResponseException(command, response, message));
+                  }
+               } else if (command.getCurrentRequest().getMethod().equals("GET")) {
+                  if (command.getCurrentRequest().getEndpoint().getPath().startsWith("/" + EtcdApiMetadata.API_VERSION + "/keys") &&
+                        message.contains("Key not found")) {
+                        exception = new ResourceNotFoundException("Could not find key: message=" + message, new HttpResponseException(command, response, message));
+                  }
```
nacx: This seems inconsistent. the methods are already configured with a fallback that looks for the `Could not find key` string. Why should that be treated in the fallback and this message in the error handler? Are there really two different  messages to describe the same error?

nacx:
Thanks @cdancy!
Apart from the inline comments, the code must be formatted according to the [jclouds style guide](https://cwiki.apache.org/confluence/display/JCLOUDS/Coding+Standards).
Keep in mind that we need the tests in order to merge this!

cdancy:
Thanks @nacx. Going over these now. Yea I wanted to flesh out any issues here before adding tests so as not to make the initial review seem monstrous.

cdancy:
@nacx I've addressed the issues you noted above. I removed some of the functionality, specifically pertaining to the various path params, as they are relevant only to code that not yet been added but will come in later commits. Thought this was the best route instead of trying to explain away what they do and why they are there without any actual code to look at.

I also added the mock and live tests. A lot of this is boiler plate code so I hope it's easy enough to digest. If not let me know and I can break it up further. Once this commit is in, and the keys API foundation is in place, the subsequent commits will be smaller and more to the point.

Thanks AOT!

etcd/src/main/java/org/jclouds/etcd/domain/keys/Node.java
```diff
@@ -0,0 +1,61 @@
...
+   public abstract String value();
+
+   Node() {
+   }
+
+   @SerializedNames({ "createdIndex", "dir", "nodes", "expiration", "key", "modifiedIndex", "ttl", "value" })
+   public static Node create(int createdIndex, boolean dir, List<Node> nodes, String expiration, String key,
+         int modifiedIndex, int ttl, String value) {
+      if (nodes == null)
+         nodes = ImmutableList.of();
```
nacx: Just apply the conditional inline when passing the parameter. We'd better avoid changing the reference for input parameters.

etcd/src/test/java/org/jclouds/etcd/features/KeysApiLiveTest.java
```diff
@@ -0,0 +1,99 @@
...
+      String localKey = UUID.randomUUID().toString().replaceAll("-", "");
+      String localValue = UUID.randomUUID().toString().replaceAll("-", "");
+      Key createdKey = api().createKey(localKey, localValue, 1);
+      assertNotNull(createdKey);
+      assertNotNull(createdKey.node().expiration());
+
+      try {
+         Thread.sleep(3000);
+      } catch (InterruptedException e) {
+         e.printStackTrace();
```
nacx: Remove this. Let the test propagate the exception and fail.

etcd/src/test/java/org/jclouds/etcd/features/KeysApiLiveTest.java
```diff
@@ -0,0 +1,99 @@
...
+import static org.testng.Assert.assertTrue;
+
+import java.util.UUID;
+
+import org.jclouds.etcd.BaseEtcdApiLiveTest;
+import org.jclouds.etcd.domain.keys.Key;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+@Test(groups = "live", testName = "KeysApiLiveTest", singleThreaded = true)
```
nacx: Does this really need to be single threaded?

etcd/src/test/java/org/jclouds/etcd/features/KeysApiLiveTest.java
```diff
@@ -0,0 +1,99 @@
...
+         Thread.sleep(3000);
+      } catch (InterruptedException e) {
+         e.printStackTrace();
+      }
+
+      createdKey = api().getKey(localKey);
+      assertNull(createdKey);
+   }
+
+   @Test(dependsOnMethods = "testCreateKeyWithTTL")
```
nacx: This test uses different keys than the TTL one. There is no need to depend on it.

etcd/src/test/java/org/jclouds/etcd/features/KeysApiLiveTest.java
```diff
@@ -0,0 +1,99 @@
...
+      assertTrue(getKey.node().value().equals(value));
+   }
+
+   @Test(dependsOnMethods = "testGetKey")
+   public void testDeleteKey() {
+      Key deletedKey = api().deleteKey(key);
+      assertNotNull(deletedKey);
+      assertTrue(deletedKey.action().equals("delete"));
+      assertTrue(deletedKey.prevNode().value().equals(value));
+   }
```
nacx: Use `@Test(alwaysRun = true)` method to avoid leaking keys if get test fails.

etcd/src/test/java/org/jclouds/etcd/features/KeysApiLiveTest.java
```diff
@@ -0,0 +1,99 @@
...
+
+   @Test(dependsOnMethods = "testCreateKey")
+   public void testGetKey() {
+      Key getKey = api().getKey(key);
+      assertNotNull(getKey);
+      assertTrue(getKey.action().equals("get"));
+      assertTrue(getKey.node().value().equals(value));
+   }
+
+   @Test(dependsOnMethods = "testGetKey")
```
nacx: Use `@Test(alwaysRun = true)` method to avoid leaking keys if get test fails.

etcd/src/test/java/org/jclouds/etcd/features/KeysApiLiveTest.java
```diff
@@ -0,0 +1,99 @@
...
+   public void testDeleteKey() {
+      Key deletedKey = api().deleteKey(key);
+      assertNotNull(deletedKey);
+      assertTrue(deletedKey.action().equals("delete"));
+      assertTrue(deletedKey.prevNode().value().equals(value));
+   }
+
+   @Test(dependsOnMethods = "testDeleteKey")
```
nacx: Don't depend on other tests. Just try to get a random value.

etcd/src/test/java/org/jclouds/etcd/features/KeysApiLiveTest.java
```diff
@@ -0,0 +1,99 @@
...
+   public void testGetNonExistentKey() {
+      Key deletedKey = api().getKey(key);
+      assertNull(deletedKey);
+   }
+
+   @Test(dependsOnMethods = "testGetNonExistentKey")
```
nacx: Same here. There's no need to depend on any test.

etcd/src/test/java/org/jclouds/etcd/features/KeysApiMockTest.java
```diff
@@ -0,0 +1,124 @@
...
+
+      server.enqueue(new MockResponse().setBody(payloadFromResource("/keys-create.json")).setResponseCode(201));
+      EtcdApi etcdApi = api(server.getUrl("/"));
+      KeysApi api = etcdApi.keysApi();
+      try {
+         Key createdKey = api.createKey("hello", "world");
+         assertNotNull(createdKey);
+         assertTrue(createdKey.node().key().equals("/hello"));
+         assertTrue(createdKey.node().value().equals("world"));
+         assertSent(server, "PUT", "/" + EtcdApiMetadata.API_VERSION + "/keys/hello");
```
nacx: You need to verify that the body (the form params in this case) are also properly sent.

etcd/src/test/java/org/jclouds/etcd/features/KeysApiMockTest.java
```diff
@@ -0,0 +1,124 @@
...
+      } finally {
+         etcdApi.close();
+         server.shutdown();
+      }
+   }
```
nacx: Add the mock test for the create with TTL method.

nacx:
> A lot of this is boiler plate code so I hope it's easy enough to digest. If not let me know and I can break it up further.

This PR was not that big, so feel free to add the other key methods back.

etcd/src/test/java/org/jclouds/etcd/features/KeysApiLiveTest.java
cdancy: No. Just a copy over from another live test. Removing now.

etcd/src/test/java/org/jclouds/etcd/features/KeysApiLiveTest.java
cdancy: True enough. Removing now.

etcd/src/test/java/org/jclouds/etcd/features/KeysApiMockTest.java
cdancy: There are no form parameters being sent it's all on the request path:

https://github.com/coreos/etcd/blob/master/Documentation/api.md#setting-the-value-of-a-key

cdancy:
@nacx for this initial PR lets keep things simple and leave those other methods out. I want to get a good foundation built and going to make the subsequent PR's easier on not only myself but others reviewing.

etcd/src/main/java/org/jclouds/etcd/domain/keys/Node.java
```diff
@@ -0,0 +1,60 @@
...
+   public abstract String value();
+
+   Node() {
+   }
+
+   @SerializedNames({ "createdIndex", "dir", "nodes", "expiration", "key", "modifiedIndex", "ttl", "value" })
+   public static Node create(int createdIndex, boolean dir, List<Node> nodes, String expiration, String key,
+         int modifiedIndex, int ttl, String value) {
+      return new AutoValue_Node(createdIndex, dir,
+            ImmutableList.copyOf(nodes != null ? nodes : ImmutableList.<Node> of()), expiration, key, modifiedIndex,
```
nacx: Why copying an empty list? Why not just: `nodes != null ? ImmutableList.copyOf(nodes) : ImmutableList.<Node> of()`

etcd/src/test/java/org/jclouds/etcd/features/KeysApiLiveTest.java
```diff
@@ -0,0 +1,102 @@
...
+@Test(groups = "live", testName = "KeysApiLiveTest")
+public class KeysApiLiveTest extends BaseEtcdApiLiveTest {
+
+   private String key;
+   private String value;
+
+   @BeforeClass
+   protected void init() {
+      key = UUID.randomUUID().toString().replaceAll("-", "");
+      value = UUID.randomUUID().toString().replaceAll("-", "");
```
nacx: Just seen that this "key id formatting convention" is applied everywhere. Better create a static helper `randomKey()` to generate the key ids?

etcd/src/test/java/org/jclouds/etcd/features/KeysApiMockTest.java
nacx: And how does the server know that the value has to be "world"? That info is sent somewhere in the request, and we need to validate that.

cdancy:
@nacx latest comments should all be addressed.

On a side/separate note: any suggestions on how to force a specific live test-class to run AFTER all other live tests are completed? In this case the MembersApiLiveTest needs to run last or risk fudging up the other tests while they are attempting to modify the cluster. All of my attempts thus far have been in vain.

nacx:
lgtm now. Thanks @cdancy!
I'm sorry, I don't know a way to configure that. You'll have to play a bit with TestNG and the configurations. If you get stuck then ping me and I'll try to give a hand.

cdancy:
@nacx squash-and-rebase and we can call it done then?

nacx:
Yes :)

cdancy:
@nacx done!

nacx:
Pushed to master as [7b21c2e6](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/7b21c2e6). Thanks @cdancy!

cdancy:
Very awesome. Thanks @nacx. Now for round 2...

PR #235, opened 2016-02-17
Refactored ProfitBricks locations to not rely on existing DataCenters

nacx:
This change removes the `DataCenter - Location` mapping. The current version uses the DataCenter entity to represent locations where nodes can be deployed, but that requires that a DataCenter for each location already exists in order to let users deploy servers to those locations using the ComputeService, which is kind of annoying when you're getting started with a clean account.

The current PR changes this, and preconfigures the ProfitBricks regions and zones as jclouds properties (like AWS, Rackspace and many others do) and does not use a ProfitBricks specific entity to represent a Location. When provisioning nodes, jclouds will take care of creating the DataCenters if needed.

This has some benefits:

* Works out of the box without users having to pre-create anything to let jclouds work.
* Users can use the real location ids ("us/las", "de/fra", etc) when specifying the `locationId` in the template builder.
* There is no need for a custom implementation of the template builder that resolves the zones.
* Removes the need for the Location and DataCenter transformation functions.
* Allows users to override the properties to "restrict" the regions/zones jclouds can work with.
* Cleans up the ComputeService live tests so they don't need to precreate DataCenters. (Note that you might need to specify the `locationId` in the template property when running the live tests if the image being used exists only in one zone).

I haven't been able to run the complete ComputeService live test suite because provisioning was terribly slow. However, I've been able to properly list locations, images, and create nodes using the jclouds-cli compiled with the changes in this PR.

@devcsrj Could you please have a look at this, and give feedback about this functional change?

I think this location abstraction change makes a lot of sense and I'd pretty much like to have it merged before promoting the provider.


devcsrj:
Man you're a magician! :+1: 

The way you did the *create-datacenter-if-not-exist* was awesome. Wasn't familiar with how `CreateNodesWithGroupEncodedIntoNameThenAddToSet` worked before, and your [custom impl](https://github.com/jclouds/jclouds-labs/pull/235/files#diff-2de75f6f337c46cc3ee37511896f795fR57) gave me a lot of context.

Just one thing I'd like to ask, how did you get the correct ISO codes for the locations? (i.e.: `DE-BW`, `DE-HE`, `US-NV`)? I remember [asking](https://botbot.me/freenode/jclouds/2014-12-20/?msg=28000713&page=1) for this before (and still couldn't find a list now) and was only able to determine, `DE` for Germany, etc. Did you manually [match](https://www.iso.org/obp/ui/#iso:code:3166:DE), like look at a map or something? :smiley: 

[nit] Just some typos in the [`TemplateWithDataCenter`](https://github.com/jclouds/jclouds-labs/pull/235/files#diff-3db81757e47ede67f080ce2a0d44aef1R43).

This **definitely** needs to get merged before the promotion. Some serious refactoring stuff you did here. Thanks @nacx !

nacx:
ComputeService live tests are passing. I get a couple failures due to resource limits, but that's because of my (still) limited account. I'll squash and merge in a while.

nacx:
Pushed to master as [7b8adc62](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/7b8adc62).

PR #236, opened 2016-02-18
[JCLOUDS-1046] Remove h2-jdbc provider from source tree

kwart:
The h2-jdbc module was not listed in jclouds-labs/pom.xml. It means the maven builds of the project don't touch this module.


kwart:
Not so nice results. I though, the failures in the h2-jdbc module are related to my local environment, but they are not:

```
Failed tests: 
  H2JdbcBlobIntegrationTest>BaseBlobIntegrationTest.deleteMultipleObjects:551->BaseBlobStoreIntegrationTest.addBlobToContainer:275->BaseBlobStoreIntegrationTest.addBlobToContainer:287 » Rollback
  H2JdbcContainerIntegrationTest>BaseContainerIntegrationTest.testBlobNameEscaping:592 » Rollback
  H2JdbcContainerIntegrationTest>BaseContainerIntegrationTest.testClearWhenContentsUnderPath:167->BaseBlobStoreIntegrationTest.add5BlobsUnderPathAnd5UnderRootToContainer:282->BaseBlobStoreIntegrationTest.addBlobToContainer:287 » Rollback
  H2JdbcBlobIntegrationTest>BaseBlobIntegrationTest.deleteMultipleObjects:550->BaseBlobStoreIntegrationTest.addBlobToContainer:275->BaseBlobStoreIntegrationTest.addBlobToContainer:287 » Rollback
  H2JdbcBlobIntegrationTest>BaseBlobIntegrationTest.deleteMultipleObjects:550->BaseBlobStoreIntegrationTest.addBlobToContainer:275->BaseBlobStoreIntegrationTest.addBlobToContainer:287 » Rollback
  H2JdbcBlobIntegrationTest>BaseBlobIntegrationTest.deleteMultipleObjects:550->BaseBlobStoreIntegrationTest.addBlobToContainer:275->BaseBlobStoreIntegrationTest.addBlobToContainer:287 » Rollback
  H2JdbcBlobIntegrationTest>BaseBlobIntegrationTest.deleteMultipleObjects:550->BaseBlobStoreIntegrationTest.addBlobToContainer:275->BaseBlobStoreIntegrationTest.addBlobToContainer:287 » Rollback
  H2JdbcContainerIntegrationTest>BaseContainerIntegrationTest.testDelimiterList:556 » Rollback
  H2JdbcContainerIntegrationTest>BaseContainerIntegrationTest.testDirectory:222 » Rollback
  H2JdbcContainerIntegrationTest>BaseContainerIntegrationTest.testSetContainerAccess:522 » HttpResponse
  H2JdbcBlobIntegrationTest>BaseBlobIntegrationTest.testSetBlobAccess:677 » HttpResponse

Tests run: 88, Failures: 11, Errors: 0, Skipped: 3
```

kwart:
The new commit removes the `h2-jdbc` provider from the source tree.

From IRC discussion on 2016-02-22:

> \<jcacek> nacx, BTW, do you have any idea what to do with failing h2-jdbc module? https://github.com/jclouds/jclouds-labs/pull/236
> \<nacx> jcacek, I'd suggest just to remove that provider from jclouds
> \<nacx> I mean, the jdbc api is intended for testing purposes, such as the filesystem provider
> \<nacx> but (at least till now) there is no real need to have a jdbc based provider
> \<nacx> supporting different jpa libraries, etc
> \<nacx> so I'd say we could just remove that provider if we don't find anyone interested in crontributing to it and fixing it

nacx:
@andrewgaul @rcoedo any thoughts on this?

rcoedo:
When I implemented this api last summer, I had two requirements:
1. The api had to be independent from the actual database engine.
2. The blobs had to be stored in the database.

Since blob operations are usually not standard in SQL, the only sane way to meet these requirements was to add and object mapper. The jdbc api ended up being a jpa api.

The jdbc api was supposed to solve the testing incompatibilities we were having with the filesystem provider when storing metadata. I believe that adding an object mapper as a dependency makes the problem even worse.

In my honest opinion, it would be better to have a small provider tied to a single database engine (like h2 or sqlite) and use directly jdbc to store only the metadata in the database and the binary file in the filesystem, as @andrewgaul has suggested in the past.

In conclusion: yes, I'd remove it :+1:

nacx:
That makes sense. And the filesystem + db provider looks like an interesting project. Perhaps something to propose as a GSoC 2016 project? @andrewgaul how do you think is better to proceed with this?

rcoedo:
This is the last year that I am eligible for the GSoC program, maybe I can make a proposal.

andrewgaul:
I have been looking at the H2-JDBC provider and made several fixes 7b3c8788caa17df47fc914aec4f42af0111257ca, 7255d73ced667002d59985df8a371509989c958c, and d169a7bb7186473ed6f794311b507b4b4975ffeb.  We have one remaining test failure which appears to be some kind of schema creation issue when running the first test.

Addressing the broader question of whether this provider has utility, I believe using JDBC holds promise as other object stores like Swift use SQLite for its metadata and the filesystem for its data.  Our current filesystem provider has poor performance for list operations due to the way it emulates a key-value store and does not support metadata on Mac OS X, one of the most popular testing environments.  The JDBC provider addresses these issues although the implementation seems clunky to me.  Using JPA is a big hammer for the handful of queries we execute against likely only a single provider.  We should probably use a database per blobstore container instead of mixing all the container and blob data in a single database.  Storing blob data in the database is likely not the best solution.  I have not completely audited the code but we have several redundant queries, inefficient ETag calculation, and resource handling leaks.

In all, the current JDBC provider does not seem like the best path forward and I would prefer something simpler and H2 and JDBC-based instead of JPA.  However, it mostly works and could help users of S3Proxy on Mac OS X have a more complete experience than the filesystem provider.  We seem unlikely to replace JDBC in the near future so I would prefer to keep it unless it causes further issues.  I would also point out that other providers like filesystem had big problems and took years of hacking to arrive at our current useful state.

andrewgaul:
@zack-shoylev Should include you in the discussion as the mentor of this project.

zack-shoylev:
Ah, thanks @andrewgaul 

I think either approach can work (fixing current code or switching to database metadata storage and filesystem blob data storage (probably using uuids or similar). The problems I see with the former you have already mentioned. The problem with the latter is that splitting the data away from the metadata makes it very difficult to have transactional operations (think about rollback/commit), something which databases should internally handle fairly well.

andrewgaul:
I made one more fix 2b597284e7ecf035cffa3bc9a59f5fbce0f7fb6b and now all tests pass so I enabled the module in 2f4dc796c325e1362ea27612c94059e4f5fc3770.  If we want to continue discussing removal of the JDBC blobstore let's revisit it on the mailing list instead of code review.

nacx:
Thanks @andrewgaul !

PR #237, opened 2016-02-19
[JCLOUDS-1084] Docker live tests fixed and made more robust

kwart:
JIRA: https://issues.apache.org/jira/browse/JCLOUDS-1084

This commit fixes several issues in Docker live tests:
* `SshToCustomPortLiveTest` doesn't depend on installing additional packages (SSH server) to alpine-linux image. The image with preinstalled Dropbear and OpenSSH servers is used instead.
* `DOCKER_HOST` environment variable is not required by `pom.xml` (it's just used as a default value for `test.docker.endpoint` system property now)
* `MiscApiLiveTest` cleanup for `testBuildImageFromDockerfile` method. It removes the image created by building `Dockerfile` after the test.

### DIFF
--- a/docker/src/test/java/org/jclouds/docker/compute/SshToCustomPortLiveTest.java
+++ b/docker/src/test/java/org/jclouds/docker/compute/SshToCustomPortLiveTest.java
@@ -17,15 +17,15 @@
 package org.jclouds.docker.compute;
 
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.jclouds.compute.options.TemplateOptions.Builder.runAsRoot;
+import static org.jclouds.compute.options.RunScriptOptions.Builder.wrapInInitScript;
+import static org.jclouds.docker.internal.DockerTestUtils.consumeStreamSilently;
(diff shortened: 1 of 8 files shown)


docker/src/test/java/org/jclouds/docker/compute/SshToCustomPortLiveTest.java
```diff
@@ -221,4 +202,11 @@ private static void consumeStreamSilently(InputStream is) {
    private static String toTag(String repo, String tag) {
       return repo + (tag != null ? ":" + tag : "");
    }
+
+   private void removeImageIfExists(String imageTag) {
```
nacx: Make this statically available in some base test class so it can be reused instead of repeated.

nacx:
Just one very minor. lgtm. Thanks @kwart!

nacx:
Pushed to master as [8e142cd2](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/8e142cd2). Thanks @kwart!

PR #239, opened 2016-02-20
JCLOUDS-1085: support for waiting on key changes

cdancy:
KeysApi gained the ability to wait on changes in key state.

### DIFF
--- a/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/AzureComputeApi.java
+++ b/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/AzureComputeApi.java
@@ -22,6 +22,8 @@ import org.jclouds.azurecompute.arm.features.JobApi;
 import org.jclouds.azurecompute.arm.features.LocationApi;
 import org.jclouds.azurecompute.arm.features.ResourceGroupApi;
 import org.jclouds.azurecompute.arm.features.StorageAccountApi;
+import org.jclouds.azurecompute.arm.features.SubnetApi;
+import org.jclouds.azurecompute.arm.features.VirtualNetworkApi;
 import org.jclouds.rest.annotations.Delegate;
 
 import javax.ws.rs.PathParam;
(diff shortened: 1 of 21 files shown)


etcd/src/test/java/org/jclouds/etcd/features/KeysApiLiveTest.java
```diff
@@ -94,6 +94,40 @@ public void testDeleteNonExistentKey() {
...
+   @Test
+   public void testWaitKey() {
+      String localKey = randomString();
+      String localValue = randomString();
+      Key createdKey = api().createKey(localKey, localValue, 3);
+      assertNotNull(createdKey);
+      assertNotNull(createdKey.node().expiration());
+      assertTrue(createdKey.node().ttl() == 3);
+
+      createdKey = api().waitKey(localKey);
```
cdancy: @nacx does jclouds have any default timeout that trips should the call not return in a specified amount of time? And if so is there a way to disable it for these "wait" methods of the KeysApi?

etcd/src/test/java/org/jclouds/etcd/features/KeysApiMockTest.java
```diff
@@ -141,4 +141,40 @@ public void testDeleteNonExistentKey() throws Exception {
...
+         Key waitedOnKey = api.waitKey("hello", 2);
+         assertNotNull(waitedOnKey);
+         assertTrue(waitedOnKey.action().equals("expire"));
+         assertTrue(waitedOnKey.prevNode().value().equals("world"));
+         assertSent(server, "GET", "/" + EtcdApiMetadata.API_VERSION + "/keys/hello?wait=true&waitIndex=2");
+      } finally {
+         etcdApi.close();
+         server.shutdown();
+      }
+   }
```
nacx: Missing tests that exercise the fallbacks.

etcd/src/test/java/org/jclouds/etcd/features/KeysApiLiveTest.java
nacx: By default methods aren't executed with specific timeouts; just the HTTP connection timeouts controlled by the timeout properties in the jclouds `Constants` class.

PR #240, opened 2016-02-22
[JCLOUDS-1002] provide access to Docker container Config object in the Node template options

kwart:
JIRA: https://issues.apache.org/jira/browse/JCLOUDS-1002

Provide fine-grained control of container configuration to `DockerTemplateOptions`. If the `configBuilder` is set, then it's used and other members introduced in `DockerTemplateOptions` are ignored.

### DIFF
--- a/docker/src/main/java/org/jclouds/docker/compute/options/DockerTemplateOptions.java
+++ b/docker/src/main/java/org/jclouds/docker/compute/options/DockerTemplateOptions.java
@@ -22,8 +22,8 @@ import static com.google.common.base.Preconditions.checkNotNull;
 import java.util.List;
 import java.util.Map;
 
-import org.jclouds.compute.ComputeService;
 import org.jclouds.compute.options.TemplateOptions;
+import org.jclouds.docker.domain.Config;
 import org.jclouds.docker.internal.NullSafeCopies;
 import org.jclouds.domain.LoginCredentials;
 import org.jclouds.javax.annotation.Nullable;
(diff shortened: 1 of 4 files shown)


kwart:
I tried also the way witha new template options class, but I've failed on `TemplateBuilderImpl.build()` which copies properties to new instance of the old options class. So I've put the `Config.Builder` member variable to the old class (`DockerTemplateOptions`) at the end.

andreaturli:
I think it gives a better control on the container spec and it is a semplification of the options as well. I agree with @nacx - *TemplateOptions couple with low-level features offered by the specific cloud so they dont need to be portable.
@gkrvlt any other thoughts?

docker/src/main/java/org/jclouds/docker/compute/options/DockerTemplateOptions.java
```diff
@@ -235,6 +281,18 @@ public DockerTemplateOptions extraHosts(Map<String, String> extraHosts) {
...
    }
 
+   /**
+    * This method sets Config.Builder configuration object, which can be used as a replacement for all the other settings from this class.
+    * If a not-<code>null</code> instance is provided, then it the compu 
+    * @param configBuilder Config.Builder instance. This instance can be changed in this method! 
+    */
+   public DockerTemplateOptions configBuilder(Config.Builder configBuilder) {
+      this.configBuilder = configBuilder != null 
+            ? Config.builder().fromConfig(configBuilder.image(NO_IMAGE).build()).image(NO_IMAGE)
```
nacx: Why setting the `NO_IMAGE` twice?

nacx:
Just a minor comment. Change lgtm. @andreaturli good to merge?

kwart:
I forgot to remove the trailing call for setting the image in `configBuilder(Config.Builder)`  after I changed its value on the input instance. :)

It's fixed now.

@nacx  Thanks for the review.

nacx:
Pushed to master as [4ff9889f](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/4ff9889f). Thanks @kwart!

PR #242, opened 2016-02-29
Profitbricks REST - Image API

mirza-spc:
@jasminSPC


profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/ImageApiLiveTest.java
```diff
@@ -0,0 +1,55 @@
...
+      List<Image> images = imageApi().getList();
+
+      assertNotNull(images);
+      assertFalse(images.isEmpty());
+      assertTrue(images.size() > 1);
+   }
+   
+   private ImageApi imageApi() {
+      return api.imageApi();
+   }
```
nacx: Add live tests for the other methods in the image api

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/ImageApiMockTest.java
```diff
@@ -0,0 +1,95 @@
...
+      );
+      
+      imageApi().deleteImage("some-id");
+      assertEquals(server.getRequestCount(), 1);
+      assertSent(server, "DELETE", "/images/some-id");
+   }
+        
+   private ImageApi imageApi() {
+      return api.imageApi();
+   }
```
nacx: Add the missing tests for all the methods in the image api, and tests for the fallbacks for those methods that define them.

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/ImageApiLiveTest.java
mirza-spc: Remaining methods are Update and Delete. I'm not quite sure if PB has a test image for that purpose? Esp. since there is no Create call.

jasminSPC: https://devops.profitbricks.com/api/rest/#images
Update and delete can be done only on private images, and to create private you have to upload them using PB control panel. This wouldn't make sense in Live tests

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/ImageApiMockTest.java
mirza-spc: All the methods in the Image API? As far as I can see they are already there List, Get, Update, Delete. By fallbacks do you mean 404s? Because @devcsrj commented
`You may remove the test for 404 fallback, as this is already tested in the core.
Add a line note
`

devcsrj: Hi @mirza-spc ! You're right, there's no need to test for 404 fallbacks. :)

In this case, @nacx might've meant, the tests for the methods with `DepthOptions` param, like for getList and getImage.

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/ImageApiLiveTest.java
devcsrj: He might be pertaining to methods with the `DepthOptions` param (getList, getImage). :)

About the *update* and *delete* operations: Since the PB API doesn't have an endpoint to actually upload images (was this via FTP, IIRC?), would it make sense if we just remove these operations, and just leave the `GET` requests to keep it simple? This is the case for the [soap counterpart](https://github.com/jclouds/jclouds/blob/master/providers/profitbricks/src/main/java/org/jclouds/profitbricks/features/ImageApi.java#L42). Although I'm not aware as to how many users would actually depend on this feature, so... thoughts?

jasminSPC: That makes sense

Jasmin Gacić

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/ImageApiMockTest.java
nacx: To avoid confusion, jclouds provides a set of default fallbacks that will be executed if no fallback is defined on a method (typically these fallbacks throw an appropriate exception).

* There should be **no tests** for fallbacks in methods that don't explicitly declare a fallback, as we don't want to test what jclouds does by default; it is already tested in jclouds core.
* There **must** be tests for fallbacks in those methods where you explicitly declare one, as that method won't use the jclouds default and we want to make sure the right fallback takes place upon failed requests.

Rule of thumb is: if your method has a `@Fallback` annotation, it needs a test. Otherwise, fallback test is not needed.

nacx: > In this case, @nacx might've meant, the missing tests for the methods with DepthOptions param, like for getList and getImage.

That's also right. **Every** method needs a mock test.

devcsrj: > Rule of thumb is: if your method has a @Fallback annotation, it needs a test. Otherwise, fallback test is not needed.

+1.

So in this case: a 404/401 test won't be needed for `updateImage`, because it doesn't have a declared fallback (and jclouds will propagate an appropriate exception), but we need one for `getImage`, so that we can verify that the method's behavior indeed returns an empty list, instead of a `ResourceNotFoundException`/`AuthorizationException`.  

I got confused with this as well. Sorry for the confusion, @mirza-spc !

mirza-spc:
I've added the missing tests for mock tests. I guess we still haven't decided the course for the live tests?

devcsrj:
I have no numbers to determine how many users are currently using/will use those image UD operations via jclouds (if this argument make sense).

If this were a vote, I'd say remove the update and delete methods, because there's no way we can write an *acceptable* live test for 'em anyway. Even if a REST endpoint for upload image is added, I don't think waiting for the upload live test of a [50MB](http://www.damnsmalllinux.org/) image (smallest I can find) is feasible. :smile: 

With that said, now that the missing mock tests were added, this already LGTM, if you decide to just keep only getImage/s operations.

nacx:
I'd say remove them too. The primary objective here should be to add the minimum required methods to make it possible to implement the ComputeService interface.

Any other methods are low priority for jclouds, as we are all about cloud portability. What will make this a first class provider in jclouds is not implementing the entire ProfitBricks API, but properly implementing the jclouds interfaces, and that should be the focus. Let's forget about anything that doesn't help reach that goal until the provider is out of labs.

nacx:
lgtm. Thanks! Can you squash the commits into a single one to prepare the merge?

nacx:
Pushed to master as [f588d6b6](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/f588d6b6). Thanks!

PR #243, opened 2016-03-02
[JCLOUDS-1088] fix NPE in DeploymentToNodeMetadata.apply

Ladicek:
The NPE is caused by passing the value of `CloudService.location()` to `LocationPredicates.idEquals` which throws a NPE when the argument is `null`. However, `CloudService.location()` is `@Nullable` and is `null` when the cloud service is associated with an affinity group.

The solution in this commit checks if `CloudService.location()` is `null` and if it is, the location is obtained from the affinity group.

Fixes https://issues.apache.org/jira/browse/JCLOUDS-1088


azurecompute/src/main/java/org/jclouds/azurecompute/compute/functions/DeploymentToNodeMetadata.java
```diff
@@ -117,8 +118,11 @@ public NodeMetadata apply(final Deployment from) {
       // TODO: CloudService name is required (see JCLOUDS-849): waiting for JCLOUDS-853.
       final CloudService cloudService = api.getCloudServiceApi().get(from.name());
       if (cloudService != null) {
+         final String location = cloudService.location() != null
+                 ? cloudService.location()
+                 : api.getAffinityGroupApi().get(cloudService.affinityGroup()).location();
```
nacx: The affinityGroup field is also marked `@Nullable`, is it always present when the location field is missing?
Also, the AffinityGroupApi.get method returns null if the affinity group is not found. Should we protect the code against this case?

nacx:
Thanks @Ladicek! (FYI the build has failed due to a checkstyle error: [unused import](https://jclouds.ci.cloudbees.com/job/jclouds-labs-pull-requests/1054/org.apache.jclouds.labs$azurecompute/violations/file/src/main/java/org/jclouds/azurecompute/compute/functions/DeploymentToNodeMetadata.java/)).

azurecompute/src/main/java/org/jclouds/azurecompute/compute/functions/DeploymentToNodeMetadata.java
Ladicek: Well, to be absolutely honest, I'm not sure -- but the `CloudService` javadoc suggests that these two are mutually exclusive.

Ladicek:
Aww, sorry about that checkstyle error. Should be fixed now.

nacx:
The change lgtm. @andreaturli ?

nacx:
Pushed to master as [fe24698d](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/fe24698d). Thanks @Ladicek!

PR #244, opened 2016-03-04
Profitbricks REST - Snapshot API

mirza-spc:
@jasminSPC


profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/binder/snapshot/UpdateSnapshotRequestBinder.java
```diff
@@ -0,0 +1,97 @@
...
+import java.util.Map;
+import org.apache.jclouds.profitbricks.rest.binder.BaseProfitBricksRequestBinder;
+import org.apache.jclouds.profitbricks.rest.domain.Snapshot;
+import org.jclouds.http.HttpRequest;
+import org.jclouds.json.Json;
+
+public class UpdateSnapshotRequestBinder extends BaseProfitBricksRequestBinder<Snapshot.Request.UpdatePayload> {
+
+   final Map<String, Object> requestBuilder;
+   final Json jsonBinder;
```
nacx: I've noticed these properties are used in all binders. Would it make sense to move them tot he parent class? Also, is there any reason why they're not private?

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/SnapshotApiLiveTest.java
```diff
@@ -0,0 +1,165 @@
...
+import static org.testng.Assert.assertNotNull;
+import org.testng.annotations.AfterClass;
+import org.testng.annotations.BeforeClass;
+
+@Test(groups = "live", testName = "SnapshotApiLiveTest")
+public class SnapshotApiLiveTest extends BaseProfitBricksLiveTest {
+   
+   DataCenter dataCenter;
+   Volume testVolume;
+   Snapshot testSnapshot;
```
nacx: Make these variables private

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/SnapshotApiLiveTest.java
```diff
@@ -0,0 +1,165 @@
...
+      assertNotNull(testVolume);
+      assertVolumeAvailable(testVolume);
+      
+      testSnapshot = api.volumeApi().createSnapshot(
+         Volume.Request.createSnapshotBuilder()
+            .dataCenterId(testVolume.dataCenterId())
+            .volumeId(testVolume.id())
+            .name("test-snapshot")
+            .description("snapshot desc...")
+            .build());
```
nacx: Move this better to a `createSnapthot` test method.

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/SnapshotApiLiveTest.java
```diff
@@ -0,0 +1,165 @@
...
+      if (dataCenter != null)
+         deleteDataCenter(dataCenter.id());
+   }
+   
+   @Test
+   public void testList() {
+      List<Snapshot> snapshots = snapshotApi().getList();
+
+      assertNotNull(snapshots);
+      assertFalse(snapshots.isEmpty());
```
nacx: Assert that it actually contains the just created snapshot?

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/SnapshotApiLiveTest.java
```diff
@@ -0,0 +1,165 @@
...
+   @Test
+   public void testGetSnapshot() {
+      Snapshot snapshot = snapshotApi().getSnapshot(testSnapshot.id());
+
+      assertNotNull(snapshot);
+      assertEquals(snapshot.id(), testSnapshot.id());
+      assertEquals(snapshot.properties().name(), "test-snapshot");
+   }  
+   
+   @Test(dependsOnMethods = "testGetSnapshot")
```
nacx: Consider changing this to depend on the create snapshot test once it exists.

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/SnapshotApiLiveTest.java
```diff
@@ -0,0 +1,165 @@
...
+      snapshotApi().deleteSnapshot(testSnapshot.id());
+      assertSnapshotRemoved(testSnapshot);
+   }
+   
+   private SnapshotApi snapshotApi() {
+      return api.snapshotApi();
+   }
+   
+   private void assertVolumeAvailable(Volume volume) {
+      assertRandom(new Predicate<VolumeRef>() {
```
nacx: Just seen this... Would it be more convenient for this method to be named `assertPredicate`? The current name seems to mean that it will perform arbitrary assertions...

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/SnapshotApiMockTest.java
```diff
@@ -0,0 +1,162 @@
...
+            
+      assertEquals(server.getRequestCount(), 1);
+      assertSent(server, "PATCH", "/rest/snapshots/some-id", "{\"name\": \"new-snapshot-name\", \"description\": \"description...\"}");
+   }
+   
+   @Test
+   public void testDelete() throws InterruptedException {
+      server.enqueue(
+         new MockResponse().setBody("")
+      );
```
nacx: Instead of an empty body, wouldn't it be better to configure a response with the 204 status code?

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/SnapshotApiMockTest.java
```diff
@@ -0,0 +1,162 @@
...
+   
+   
+   @Test
+   public void testGetListWith404() throws InterruptedException {
+      server.enqueue(response404());
+      List<Snapshot> list = snapshotApi().getList(new DepthOptions().depth(1));
+      assertTrue(list.isEmpty());
+      assertEquals(server.getRequestCount(), 1);
+      assertSent(server, "GET", "/snapshots?depth=1");
+   }
```
nacx: Add a test for the 404 fallback for the other `getList` too.

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/SnapshotApiMockTest.java
```diff
@@ -0,0 +1,162 @@
...
+   public void testGetSnapshotWith404() throws InterruptedException {
+      server.enqueue(response404());
+
+      Snapshot snapshot = snapshotApi().getSnapshot("some-id");
+      
+      assertEquals(snapshot, null);
+
+      assertEquals(server.getRequestCount(), 1);
+      assertSent(server, "GET", "/snapshots/some-id");
+   }
```
nacx: Add a test for the 404 fallback for the other `getSnapshot` too.

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/features/SnapshotApi.java
```diff
@@ -19,29 +19,72 @@
...
 
 @Path("/snapshots")
 @RequestFilters(BasicAuthentication.class)
 public interface SnapshotApi extends Closeable {
    
+   @Named("snapshot:list")
+   @GET
+   @SelectJson("items")
+   @Fallback(EmptyListOnNotFoundOr404.class)
+   List<Snapshot> getList();
```
nacx: I've just noticed this pattern. In other jclouds APIs, the common name for list methods is just `list` (instead of getList). Mind doing the change?

nacx:
Thanks @mirza-spc! This looks pretty good; all comments are minors.

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/SnapshotApiLiveTest.java
mirza-spc: That is already tested in volume api, there is not create snapshot in snapshot api as per docs.

nacx: Oh, you're right. Dismiss this comment then! :)

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/features/SnapshotApi.java
mirza-spc: ok but you still have getSnapshot, should it be just get? or createSnapshot => create? that would seem along the lines of such a convention. or is it just that the list method is a special case?

nacx: Yes, that's the convention we try to use now, so it applies to all mehtods.

mirza-spc: this should mean that all of the methods of all of the apis in profitbricks rest should be updated to the convention format?

nacx: Yes, but let's leave that for another PR and just amend the ones in the snapshot api in this one.

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/SnapshotApiLiveTest.java
```diff
@@ -0,0 +1,176 @@
...
+      assertFalse(snapshots.isEmpty());
+      
+      Snapshot newSnapshot = null;
+      
+      for (Snapshot snapshot : snapshots) {
+         if (snapshot.id().equals(testSnapshot.id())) {
+            newSnapshot = snapshot;
+            break;
+         }
+      }
```
nacx: Better use:
```java
assertTrue(Iterables.any(snapshots, new Predicate<Snapshot>() {
   @Override public boolean apply(Snapshot input) {
      return input.id().equals(testSnapshot.id());
   }
}));
```

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/binder/snapshot/UpdateSnapshotRequestBinder.java
nacx: What about this comment? Something that makes sense for a separate PR? Also I've noticed some binders declare the variables in the default scope. If there is no good reason for that, they should be made private. We could address both in a new PR.

mirza-spc: I've refactored it out into the base class. Good thing for noticing it. I think they are only two binders that build the form encoded data, all others are using json.

nacx:
Thanks @mirza-spc! Mind squashing the commits to prepare the merge?

nacx:
Pushed to master as [ca57d90d](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/ca57d90d). Thanks!

PR #245, opened 2016-03-07
Volume API SSH keys

mirza-spc:
@jasminSPC


jasminSPC:
@nacx As per our email discussion this PR contains changes regarding SSH keys and the new endpoint. Do you think it would be a good idea to use this PR to rename all the functions that were named like getList and similar?

nacx:
> Do you think it would be a good idea to use this PR to rename all the functions that were named like getList and similar?

Let's use a specific PR for that.

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/domain/Volume.java
```diff
@@ -135,12 +136,14 @@ public static BusType fromValue(String value) {
...
          public abstract String name();
 
-         @Nullable
          public abstract String type();
-
+         
          public abstract int size();
 
          @Nullable
+         public abstract List<String> sshKeys();
```
nacx: Better use a `Set` to avoid duplicates?

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/binder/volume/CreateVolumeRequestBinderTest.java
```diff
@@ -42,6 +42,7 @@ public void testCreatePayload() {
               .dataCenterId("datacenter-id")
               .name("jclouds-volume")
               .size(3)
+              .type("HDD")
```
nacx: Should an enum work better here, so users have a constrained set of options to use?

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/VolumeApiMockTest.java
```diff
@@ -102,7 +104,35 @@ public void testCreate() throws InterruptedException {
...
+   }
+   
+   @Test
+   public void testCreateWithSsh() throws InterruptedException {
+      server.enqueue(
+         new MockResponse().setBody(stringFromResource("/volume/get.json"))
+      );
+      
+      List<String> sshKeys = new ArrayList<String>();
+      sshKeys.add("hQGOEJeFL91EG3+l9TtRbWNjzhDVHeLuL3NWee6bekA=");
```
nacx: Is this meant to be the contents of an ssh public key? Is this a real value, or could we use a real one here?

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/VolumeApiLiveTest.java
```diff
@@ -60,6 +60,7 @@ public void testCreateVolume() {
               .name("jclouds-volume")
               .size(3)
               .licenceType(LicenceType.LINUX)
+              .type("HDD")
```
nacx: Configure some ssh key too

nacx:
Thanks @jasminSPC and @mirza-spc! This is a long awaited feature!

devcsrj:
Nice! With this, pre-configuring password-ready images will no longer be needed. +1

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/VolumeApiMockTest.java
mirza-spc: As far as I've been given information, the key should fit into a string inside the list/set. So yes, I think that the real value should be here. This one seems to be a 256 bit key

mirza-spc:
Live test for ssh key is still not passing, even though it is supposed to work as in https://github.com/StackPointCloud/profitbricks-sdk-go/blob/master/volume_test.go#L11

Bad request is the response. I've got no API docs for this.

nacx:
If the live test fails, then any user trying to use this feature will experience the same issue. We need to fix it before merging.

Try enabling the [jclouds.wire logger](http://jclouds.apache.org/reference/logging/) so you can see the exact HTTP request that is being made and compare it by the one generated by the Go test. Hopefully you'll see where they differ and have a hint to fix the test.

jasminSPC:
We found the issue, thanks.

On Tue, Mar 8, 2016 at 6:55 PM, Ignasi Barrera <notifications@github.com>
wrote:

> If the live test fails, then any user trying to use this feature will
> experience the same issue. We need to fix it before merging.
>
> Try enabling the jclouds.wire logger
> <http://jclouds.apache.org/reference/logging/> so you can see the exact
> HTTP request that is being made and compare it by the one generated by the
> Go test. Hopefully you'll see where they differ and have a hint to fix the
> test.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/jclouds/jclouds-labs/pull/245#issuecomment-193890303>.
>



-- 
Jasmin Gacić

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/binder/BaseProfitBricksRequestBinder.java
```diff
@@ -16,20 +16,18 @@
...
-
 import com.google.common.base.Strings;
-
 import org.jclouds.io.MutableContentMetadata;
 import org.jclouds.io.payloads.BaseMutableContentMetadata;
+import static com.google.common.base.Preconditions.checkNotNull;
 
 public abstract class BaseProfitBricksRequestBinder<T> implements MapBinder {
 
+   protected final String endpointPath = "/rest/v2";
```
nacx: Don't hardcode this. Extract it from the provider metadata. To do so:

* Add a parameter to the constructor: `Supplier<URI> endpointSupplier`. This will provide the endpoint configured in the metadata.
* Modify subclasses to inject the endpoint supplier, by adding the supplier parameter as: `@Provider Supplier<URI> endpointSupplier` (and making sure the constructor is annotated with `@Inject`). Note that subclasses add the qualifier annotation to get the right supplier injected. It would be redundant in the parent class as it is abstract and shouldn't care about implementations using injection or not.

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/VolumeApiLiveTest.java
```diff
@@ -53,13 +55,19 @@ public void teardownTest() {
...
 
       testVolume = volumeApi().createVolume(
               Volume.Request.creatingBuilder()
               .dataCenterId(dataCenter.id())
               .name("jclouds-volume")
-              .size(3)
+              .size(4)
               .licenceType(LicenceType.LINUX)
+              .type(VolumeType.HDD)
+              .image("362c903a-df61-11e5-80a4-52540005ab80")
```
nacx: Is this a global image that is supposed to exist?

nacx:
Almost there! Just a couple final comments.

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/VolumeApiLiveTest.java
jasminSPC: Only specific images support sshKeys. The other option was to fetch all public images and then filter them by name "Ubuntu-15.05...".

jasminSPC: Actually I take it back, whenever PB updates their images the UUIDs change. So I think we will have to grab images by name and location.

nacx: Live tests take time. If there is a way to avoid hardcoding image ids (that eventually change when the provider updates the image list), then I'd go for that approach, even if that implies performing some additional calls. The time overhead introduced won't be that much compared to the amount of time live tests to run. WDYT?

jasminSPC: I'm afraid of introducing more complexity to the unit test. If we add this this test wouldn't be atomic any more. 
These tests won't be ran too often so it wouldn't hurt to change them every time we run them.

nacx: The same way you create the datacenter in the before class, can't you get the list of available images and get a proper one?

Once everything is in place we'll configure our CI system to run live tests weekly. Apart from that, we try to run live tests in every release, to validate everything is OK. Having to do a change at release time if an image id has changed can be kinda frustrating for the releaser :)

jasminSPC: In that case we'll have to add it :)

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/binder/BaseProfitBricksRequestBinder.java
mirza-spc: I keep geeting the `No implementation for com.google.common.base.Supplier<java.net.URI> annotated with @org.jclouds.location.Provider() was bound.`

nacx: Hmm weird. That provider is bound by default in the [LocationModule](https://github.com/jclouds/jclouds/blob/master/core/src/main/java/org/jclouds/location/config/LocationModule.java#L85-L92) which is included by default in the [RestModule](https://github.com/jclouds/jclouds/blob/master/core/src/main/java/org/jclouds/rest/config/RestModule.java#L51) (the ruperclass of the `HttpApiModule`), so you should already have it.
Are you experiencing this issue when running the tests? If so, are you using a custom (small) Injector or a custom module set? If that is the case, try including the `LocationModule` in the lsit of modules used to create that small injector.

mirza-spc: Yes, I'm trying to run the binder test. `UpdateImageRequestBinderTest.java`. To achieve this, I have 

`Injector injector = Guice.createInjector(new GsonModule(), new LocationModule());`. 

However I still get an error:
 ```20) No implementation for java.lang.Long annotated with @com.google.inject.name.Named(value=jclouds.session-interval) was bound.
  at org.jclouds.location.config.LocationModule.zoneIdsSupplier(LocationModule.java:128)```

nacx: Oh, you can bind a value for test by adding a new module, like this:
```java
Injector injector = Guice.createInjector(new GsonModule(), new LocationModule(), new AbstractModule() {
   @Override protected void configure() {
      bind(String.class).annotatedWith(Names.named(Constants.PROPERTY_SESSION_INTERVAL)).toInstance("60");
   }
});
```
If it fails, could you paste the entire failure in a Gist?

mirza-spc: Thanks for your help, but I solved it in a different way, because I still kept getting further errors with the last injector snippet.
`protected final String endpoint = new ProfitBricksProviderMetadata().getEndpoint();`
and changed a little bit the request builder part
```R request = (R) fromRequest.toBuilder().endpoint(endpointPath).build();
      request = (R) request.toBuilder().replacePath(request.getEndpoint().getPath() + String.format("images/%s", imageId)).build();```

Comment (nacx) [inline profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/binder/BaseProfitBricksRequestBinder.java]: In fact that is not a good option, as when building the context, jclouds substitutes several values and properties from the metadata based on the properties passed to the context builder (it is not a use case right now here, but even when the provider metadata sets the endpoint users can override it when creating the context)., so you really need to use the injection approach to inject the right endpoint.

Please change the constructor back to inject the `@Provider Supplier<URI> endpointSupplier`. Regarding the test, I've had a look and there is no need to configure the `LocationModule`; just binding the supplier used in the base binder class is enough:

```java
Injector injector = Guice.createInjector(new GsonModule(), new AbstractModule() {
   @Override protected void configure() {
      bind(new TypeLiteral<Supplier<URI>>() {
      }).annotatedWith(Provider.class).toInstance(
         Suppliers.ofInstance(URI.create(new ProfitBricksProviderMetadata().getEndpoint())));
      }
   });
```
I've applied the injection approach and configured the `UpdateImageRequestBinderTest` like this and tests passed. Could you apply the same pattern to the rest of the tests?

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/binder/BaseProfitBricksRequestBinder.java
```diff
@@ -16,24 +16,24 @@
...
-
 import org.jclouds.io.MutableContentMetadata;
 import org.jclouds.io.payloads.BaseMutableContentMetadata;
+import java.net.URI;
+import static com.google.common.base.Preconditions.checkNotNull;
+import com.google.common.base.Supplier;
 
 public abstract class BaseProfitBricksRequestBinder<T> implements MapBinder {
 
+   protected Supplier<URI> endpointSupplier;
```
nacx: This should be declared final and added as a parameter to the constructor. You want to make sure this is always set by subclasses, ottherwise NPEs might appear (there is no need to annotate with `@Inject` or `@Provider` here; the way the supplier is used is up to the subclasses).

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/binder/BaseProfitBricksRequestBinder.java
```diff
@@ -16,24 +16,24 @@
...
    protected final String paramName;
 
    protected BaseProfitBricksRequestBinder(String paramName) {
       this.paramName = checkNotNull(paramName, "Initialize 'paramName' in constructor");
+      URI t = URI.create("https://www.profitbricks.com/");
```
nacx: Remove this

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/binder/volume/CreateSnapshotRequestBinderTest.java
```diff
@@ -34,7 +41,14 @@
...
    public void testCreatePayload() {
       
-      Injector injector = Guice.createInjector(new GsonModule());
+      Injector injector = Guice.createInjector(new GsonModule(), new AbstractModule() {
+         @Override protected void configure() {
+            bind(new TypeLiteral<Supplier<URI>>() {
+            }).annotatedWith(Provider.class).toInstance(
+               Suppliers.ofInstance(URI.create(new ProfitBricksProviderMetadata().getEndpoint())));
+            }
+         });
```
nacx: This will be needed in all binder tests. Create a base test class for all binder tests that initializes the injector variable in a `@BeforeClass` method, so tests can just use it and focus on tests.

nacx:
Good to see all tests are passing now! There are a few final comments. Thanks @mirza-spc!

nacx:
lgtm! Mind squashing the commits and rebase (better after squashing all commits into a single one to see less conflicts) so we can merge it?

nacx:
Pushed to master as [bb9dfbaf(http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/bb9dfbaf). Thanks and apologies for the delay! Don't hesitate to ping us and comment if the PR gets unattended!

PR #247, opened 2016-03-18
Update DockerComputeServiceAdapter.java

zack-shoylev:
Is this needed for the compute service adapter?

It's causing issues against one of the docker service providers I am testing with.


nacx:
@zack-shoylev push the branch to the ASF remote, otherwise the mirror job will delete it :)

Back to the PR, let's make that flag configurable from the Docker template options.

zack-shoylev:
@nacx I figured that out about the mirror. Fun github stuff :)

I like the idea of  doing this through the template options. I will look into that.

PR #248, opened 2016-03-18
Use non-privileged mode for compute

zack-shoylev:
Is this needed for the compute service adapter?

It's causing issues against one of the docker service providers I am testing with.

### DIFF
--- a/docker/src/main/java/org/jclouds/docker/compute/options/DockerTemplateOptions.java
+++ b/docker/src/main/java/org/jclouds/docker/compute/options/DockerTemplateOptions.java
@@ -37,7 +37,7 @@ import com.google.common.collect.ImmutableMap;
  * Contains options supported by the
  * {@link org.jclouds.compute.ComputeService#createNodesInGroup(String, int, TemplateOptions)
  * createNodes} operation on the <em>docker</em> provider.
- * 
+ *
  * <h2>Usage</h2>
  *
  * The recommended way to instantiate a DockerTemplateOptions object is to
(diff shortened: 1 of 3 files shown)


nacx:
See comment: https://github.com/jclouds/jclouds-labs/pull/247#issuecomment-198543632

docker/src/main/java/org/jclouds/docker/compute/strategy/DockerComputeServiceAdapter.java
```diff
@@ -109,7 +109,7 @@ public DockerComputeServiceAdapter(DockerApi api) {
 
          HostConfig.Builder hostConfigBuilder = HostConfig.builder()
                  .publishAllPorts(true)
-                 .privileged(true);
+                 .privileged(false);
```
andreaturli: probably better to add a property in the DockerTemplateOptions

docker/src/main/java/org/jclouds/docker/compute/options/DockerTemplateOptions.java
```diff
@@ -94,9 +94,10 @@
    protected Map<Integer, Integer> portBindings = ImmutableMap.of();
    protected String networkMode;
    protected Map<String, String> extraHosts = ImmutableMap.of();
+   protected Boolean privileged;
```
nacx: Better make this a primitive boolean; there's no need to have to deal with nulls

nacx:
Thanks @zack-shoylev! Just a comment; apart from that lgtm!

zack-shoylev:
merged

PR #249, opened 2016-03-22
Profitbricks REST - NIC API

mirza-spc:
@jasminSPC

### DIFF
--- a/profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/binder/nic/CreateNicRequestBinder.java
+++ b/profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/binder/nic/CreateNicRequestBinder.java
@@ -14,59 +14,61 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.apache.jclouds.profitbricks.rest.binder.volume;
+package org.apache.jclouds.profitbricks.rest.binder.nic;
 
+import static com.google.common.base.Preconditions.checkNotNull;
+import com.google.common.base.Supplier;
 import com.google.inject.Inject;
+import java.net.URI;
 import java.util.HashMap;
 import java.util.Map;
 import org.apache.jclouds.profitbricks.rest.binder.BaseProfitBricksRequestBinder;
-import org.apache.jclouds.profitbricks.rest.domain.Volume;
+import org.apache.jclouds.profitbricks.rest.domain.Nic;
 import org.jclouds.http.HttpRequest;
 import org.jclouds.json.Json;
-import com.google.common.base.Supplier;
-import java.net.URI;
 import org.jclouds.location.Provider;
-import static com.google.common.base.Preconditions.checkNotNull;
 
-public class CreateVolumeRequestBinder extends BaseProfitBricksRequestBinder<Volume.Request.CreatePayload> {
+public class CreateNicRequestBinder extends BaseProfitBricksRequestBinder<Nic.Request.CreatePayload> {
 
    private String dataCenterId;
+   private String serverId;
 
    @Inject
-   CreateVolumeRequestBinder(Json jsonBinder, @Provider Supplier<URI> endpointSupplier) {
-      super("volume", jsonBinder, endpointSupplier);
+   CreateNicRequestBinder(Json jsonBinder,  @Provider Supplier<URI> endpointSupplier) {
+      super("nic", jsonBinder, endpointSupplier);
    }
 
    @Override
-   protected String createPayload(Volume.Request.CreatePayload payload) {
+   protected String createPayload(Nic.Request.CreatePayload payload) {
       
+      checkNotNull(payload, "payload");
       checkNotNull(payload.dataCenterId(), "dataCenterId");
-      checkNotNull(payload.type(), "type");
+      checkNotNull(payload.serverId(), "serverId");
 
       dataCenterId = payload.dataCenterId();
-      
+      serverId = payload.serverId();
+            
       Map<String, Object> properties = new HashMap<String, Object>();
       
-      properties.put("type", payload.type());
-      
-      properties.put("size",  payload.size());
+      properties.put("lan",  payload.lan());
       
       if (payload.name() != null)
          properties.put("name", payload.name());
       
-      if (payload.bus() != null)
-         properties.put("bus", payload.bus());
+      if (payload.ips() != null && !payload.ips().isEmpty())
+         properties.put("ips", payload.ips());
       
-      if (payload.sshKeys() != null)
-         properties.put("sshKeys", payload.sshKeys());
+      if (payload.dhcp() != null)
+         properties.put("dhcp", payload.dhcp());
       
-      if (payload.imagePassword() != null)
-         properties.put("imagePassword", payload.imagePassword());
+      if (payload.firewallActive() != null)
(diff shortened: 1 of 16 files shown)


profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/binder/nic/CreateNicRequestBinder.java
```diff
@@ -0,0 +1,85 @@
...
+      
+      if (payload.firewallrules() != null) {
+         Map<String, Object> entities = new HashMap<String, Object>();
+         entities.put("firewallrules", payload.firewallrules());
+         properties.put("entities", entities);
+      }
+      
+      requestBuilder.put("properties", properties);
+      
+      return (new Gson()).toJson(requestBuilder);
```
nacx: Use the injected `Json`.

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/binder/nic/UpdateNicRequestBinder.java
```diff
@@ -0,0 +1,73 @@
...
+      if (payload.name() != null)
+         requestBuilder.put("name", payload.name());
+      
+      if (payload.ips() != null && !payload.ips().isEmpty())
+         requestBuilder.put("ips", payload.ips());
+      
+      if (payload.dhcp() != null)
+         requestBuilder.put("dhcp", payload.dhcp());
+      
+      return (new Gson()).toJson(requestBuilder);
```
nacx: Same here; use the injected one.

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/features/NicApi.java
```diff
@@ -0,0 +1,126 @@
...
+
+@Path("/datacenters/{dataCenterId}/servers/{serverId}/nics")
+@RequestFilters(BasicAuthentication.class)
+public interface NicApi extends Closeable {
+   
+   @Named("nic:list")
+   @GET
+   @SelectJson("items")
+   @Fallback(EmptyListOnNotFoundOr404.class)
+   List<Nic> getList(@PathParam("dataCenterId") String dataCenterId, @PathParam("serverId") String serverId);
```
nacx: This was already discussed. Let's start using a consistent naming in new APIs ()feel free to modify the existing ones if you want). Rename this to just "list", the get operations to just "get", etc. This is the NicApi, so there is no need for the method names to include the context of what entity is being managed.

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/features/NicApi.java
```diff
@@ -0,0 +1,126 @@
...
+   
+   @Named("nic:delete")
+   @DELETE
+   @Path("/{nicId}")
+   @Fallback(Fallbacks.VoidOnNotFoundOr404.class)
+   void deleteNic(@PathParam("dataCenterId") String dataCenterId, @PathParam("serverId") String serverId, @PathParam("nicId") String nicId);
+   
+   static final class NicParser extends ParseJson<Nic> {
+      
+      final ParseId parseService;
```
nacx: Make this private.

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/NicApiLiveTest.java
```diff
@@ -0,0 +1,153 @@
...
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertFalse;
+import static org.testng.Assert.assertNotNull;
+
+@Test(groups = "live", testName = "NicApiLiveTest")
+public class NicApiLiveTest extends BaseProfitBricksLiveTest {
+   
+   DataCenter dataCenter;
+   Server testServer;
+   Nic testNic;
```
nacx: Make all these variables private unless there is a real need to have it package-visible.

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/NicApiLiveTest.java
```diff
@@ -0,0 +1,153 @@
...
+      assertNicAvailable(testNic);
+      
+      Nic nic = nicApi().getNic(dataCenter.id(), testServer.id(), testNic.id());
+      
+      assertEquals(nic.properties().name(), "apache-nic");
+   }
+   
+
+   @Test(dependsOnMethods = "testUpdateNic")
+   public void testDeleteNic() {
```
nacx: We should be able to test this even if the update test fails.

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/NicApiLiveTest.java
```diff
@@ -0,0 +1,153 @@
...
+         public boolean apply(String args) {
+            String[] params = args.split(",");
+            Nic nic = nicApi().getNic(params[0], params[1], params[2]);
+            
+            if (nic == null || nic.metadata() == null)
+               return false;
+            
+            return nic.metadata().state() == State.AVAILABLE;
+         }
+      }, complexId(nic.dataCenterId(), testServer.id(), nic.id()));
```
nacx: Why assembling this into a String to just split it immediately? Why not passing directly the Nic object to the predicate?

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/NicApiLiveTest.java
mirza-spc: But if the update test fails, we don't consider the tests passing? Also I'd still have to wait since I don't really want to update and delete at the same time.

nacx: Tests don't pass but we need to know which pass and which don't. Just add an `alwaysRun = true` to the test annotation in this method.

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/NicApiLiveTest.java
```diff
@@ -0,0 +1,152 @@
...
+   }
+   
+   private void assertNicRemoved(Nic nic) {
+      assertPredicate(new Predicate<String>() {
+         @Override
+         public boolean apply(String args) {
+            String[] params = args.split(",");
+            return nicApi().get(params[0], params[1], params[2]) == null;
+         }
+      }, complexId(nic.dataCenterId(), testServer.id(), nic.id()));
```
nacx: Use the Nic for the predicate here too.

nacx:
Just one comment left. Thanks for the quick fixes!

nacx:
Thanks! Can you squash the commits before merging?

mirza-spc:
Commits are squashed!

nacx:
Pushed to master as [e95de6a7](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/e95de6a7). Thanks @mirza-spc!

PR #251, opened 2016-03-29
Profitbricks REST - LAN API

mirza-spc:
@jasminSPC

### DIFF
--- a/profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/LanApiLiveTest.java
+++ b/profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/LanApiLiveTest.java
@@ -1,0 +1,141 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.jclouds.profitbricks.rest.features;
+
+import com.google.common.base.Predicate;
+import com.google.common.collect.Iterables;
+import java.util.List;
+import org.apache.jclouds.profitbricks.rest.domain.DataCenter;
+import org.apache.jclouds.profitbricks.rest.domain.Snapshot;
+import org.apache.jclouds.profitbricks.rest.domain.State;
+import org.apache.jclouds.profitbricks.rest.domain.Lan;
+import org.apache.jclouds.profitbricks.rest.internal.BaseProfitBricksLiveTest;
+import org.testng.annotations.AfterClass;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertFalse;
+import static org.testng.Assert.assertNotNull;
+import static org.testng.Assert.assertTrue;
+
+@Test(groups = "live", testName = "LanApiLiveTest")
+public class LanApiLiveTest extends BaseProfitBricksLiveTest {
+   
+   private DataCenter dataCenter;
+   private Lan testLan;
+   private Snapshot testSnapshot;
+  
+   @BeforeClass
+   public void setupTest() {
+      dataCenter = createDataCenter();
+   }
+   
+   @AfterClass(alwaysRun = true)
+   public void teardownTest() {
+      if (dataCenter != null)
+         deleteDataCenter(dataCenter.id());
+   }
+     
+   @Test
+   public void testCreateLan() {
+      assertNotNull(dataCenter);
+            
+      testLan = lanApi().create(
+              Lan.Request.creatingBuilder()
+              .dataCenterId(dataCenter.id())
+              .name("jclouds-lan")
+              .build());
+
+      assertNotNull(testLan);
+      assertEquals(testLan.properties().name(), "jclouds-lan");
+      assertLanAvailable(testLan);
+   }
+   
+
+   @Test(dependsOnMethods = "testCreateLan")
+   public void testGetLan() {
+      Lan lan = lanApi().get(dataCenter.id(), testLan.id());
+
+      assertNotNull(lan);
+      assertEquals(lan.id(), testLan.id());
+   }
+
+   @Test(dependsOnMethods = "testCreateLan")
+   public void testList() {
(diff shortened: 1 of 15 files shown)


profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/LanApiLiveTest.java
```diff
@@ -0,0 +1,137 @@
...
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertFalse;
+import static org.testng.Assert.assertNotNull;
+
+@Test(groups = "live", testName = "LanApiLiveTest")
+public class LanApiLiveTest extends BaseProfitBricksLiveTest {
+   
+   DataCenter dataCenter;
+   Lan testLan;
+   Snapshot testSnapshot;
```
nacx: Make these variables private

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/LanApiLiveTest.java
```diff
@@ -0,0 +1,137 @@
...
+      assertEquals(lan.id(), testLan.id());
+   }
+
+   @Test(dependsOnMethods = "testCreateLan")
+   public void testList() {
+      List<Lan> lans = lanApi().list(dataCenter.id());
+
+      assertNotNull(lans);
+      assertFalse(lans.isEmpty());
+      assertEquals(lans.size(), 1);
```
nacx: A more robust assertion would be to check that the list *contains* the created Lan. Just to avoid race conditions if at some point we create something else in this datacenter.

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/LanApiLiveTest.java
```diff
@@ -0,0 +1,137 @@
...
+   }
+   
+   @Test(dependsOnMethods = "testGetLan")
```
nacx: Can this just depend on the create test? This will reduce the chances this test is skipped.

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/LanApiLiveTest.java
```diff
@@ -0,0 +1,137 @@
...
+   
+   private void assertLanRemoved(Lan lan) {
+      assertPredicate(new Predicate<String>() {
+         @Override
+         public boolean apply(String args) {
+            String[] params = args.split(",");
+            return lanApi().get(params[0], params[1]) == null;
+         }
+      }, complexId(lan.dataCenterId(), lan.id()));
+   }
```
nacx: Same thing here about this ugly complex id thing. Get rid of this pattern, please.

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/LanApiLiveTest.java
```diff
@@ -0,0 +1,137 @@
...
+              .build());
+
+      assertLanAvailable(testLan);
+      
+      Lan lan = lanApi().get(dataCenter.id(), testLan.id());
+      
+      assertEquals(lan.properties().isPublic(), false);
+   }
+   
+   @Test(dependsOnMethods = "testUpdateLan")
```
nacx: Is an `alwaysRun = True` needed here, or the tear down that removes the datacenter is enough?

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/LanApiMockTest.java
```diff
@@ -0,0 +1,148 @@
...
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertNotNull;
+import static org.testng.Assert.assertTrue;
+import org.testng.annotations.Test;
+
+@Test(groups = "unit", testName = "LanApiMockTest", singleThreaded = true)
+public class LanApiMockTest extends BaseProfitBricksApiMockTest {
+   
+   @Test
+   public void testGetList() throws InterruptedException {
```
nacx: Rename tests to `testList`

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/LanApiMockTest.java
```diff
@@ -0,0 +1,148 @@
...
+   }
+   
+   @Test
+   public void testGetListWith404() throws InterruptedException {
+      server.enqueue(response404());
+      List<Lan> list = lanApi().list("datacenter-id", new DepthOptions().depth(1));
+      assertTrue(list.isEmpty());
+      assertEquals(server.getRequestCount(), 1);
+      assertSent(server, "GET", "/datacenters/datacenter-id/lans?depth=1");
+   }
```
nacx: You need to test each method independently. Missing tests:
* List with options when response is 200.
* List without options when response is 404.
* Get with options when response is 200.
* Get with options when response is 404.

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/LanApiLiveTest.java
mirza-spc: I would say so, because all the things in the datacenter get removed with it.

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/features/LanApiLiveTest.java
```diff
@@ -0,0 +1,141 @@
...
+      assertNotNull(lans);
+      assertFalse(lans.isEmpty());
+      assertTrue(Iterables.any(lans, new Predicate<Lan>() {
+         @Override public boolean apply(Lan input) {
+            return input.id().equals(testLan.id());
+         }
+      }));
+   }
+   
+   @Test(dependsOnMethods = "testCreateLan", alwaysRun = true)
```
nacx: We really don't want to execute this if the lan was not created, so let's remove the `alwaysRun = true` from this test.

nacx:
Just one last comment. Mind fixing it and directly squash the commits to prepare the merge? Thx!

nacx:
Pushed to master as [ff028de1](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/ff028de1). Thanks @mirza-spc!

jasminSPC:
@nacx I see this This pull request is closed, but the StackPointCloud:pb-lan-api **branch has unmerged commits.** is this how it is supposed to be?

nacx:
It is OK. I don't "git merge" contributions for two reasons:

* To avoid unnecessary merge commits.
* To properly preserve authorship, so the commit properly reflects the author and the committer (when using git merge the committer is set tot he author even when he's not the one that checked in the code).

For this reason I usually merge contributions with a cherry-pick (which changes the commit hash), and that's why Github does not detect it as merged.

Also, commits are pushed to the Apache Git repos and mirrored to GitHub, so it might be a while until commits appear here, but don't worry, everything is OK. If in doubt, please, double-check and compare the commit that is in master with the one in your branch!

nacx:
BTW, that's why I always put a link to the commit in the Apache git repo.

jasminSPC:
Got it, thanks

Jasmin Gacić

PR #252, opened 2016-04-01
Added volumesFrom to Docker template options

grkvlt:
Added new `volumesFrom` option to `DockerTemplateOptions` and tidies up some of the `@Nullable` entries in the configuration domain objects. Release 1.9.x version added as #253.


andreaturli:
same comments as #252 
probably you don't need to create 2 PRs for master and 1.9.x but once master is merged you can certainly ask to backport it to 1.9.x. This way I think we reduce the number of PRs. Thoughts?

andreaturli:
thanks @grkvlt. Merged at [master](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/31381953)

PR #253, opened 2016-04-01
Added volumesFrom to Docker template options

grkvlt:
See #252 for more information.

### DIFF
--- a/docker/src/main/java/org/jclouds/docker/domain/Config.java
+++ b/docker/src/main/java/org/jclouds/docker/domain/Config.java
@@ -64,7 +64,7 @@ public abstract class Config {
 
    public abstract String image();
 
-   public abstract Map<String, ?> volumes();
+   @Nullable public abstract Map<String, ?> volumes();
 
    @Nullable public abstract String workingDir();
 
@@ -76,9 +76,9 @@ public abstract class Config {
 
    @Nullable public abstract HostConfig hostConfig();
 
-   public abstract List<String> binds();
+   @Nullable public abstract List<String> binds();
 
-   public abstract List<String> links();
+   @Nullable public abstract List<String> links();
 
    public abstract List<Map<String, String>> lxcConf();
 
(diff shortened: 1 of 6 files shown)


grkvlt:
@andreaturli @nacx can you please review this and #252 which update the available options for Docker

docker/src/main/java/org/jclouds/docker/domain/Config.java
```diff
@@ -117,14 +117,14 @@ public static Config create(String hostname, String domainname, String user, int
...
-              copyWithNullOf(entrypoint), image, copyOf(volumes), workingDir, networkDisabled,
+              copyWithNullOf(entrypoint), image, copyWithNullOf(volumes), workingDir, networkDisabled,
               copyOf(exposedPorts), copyOf(securityOpts), hostConfig,
-              copyOf(binds), copyOf(links), copyOf(lxcConf), copyOf(portBindings), publishAllPorts, privileged,
-              copyWithNullOf(dns), dnsSearch, volumesFrom, copyOf(capAdd), copyOf(capDrop), copyOf(restartPolicy));
+              copyWithNullOf(binds), copyWithNullOf(links), copyOf(lxcConf), copyOf(portBindings), publishAllPorts, privileged,
+              copyWithNullOf(dns), copyWithNullOf(dnsSearch), copyWithNullOf(volumesFrom), copyWithNullOf(capAdd), copyWithNullOf(capDrop), copyOf(restartPolicy));
    }
 
    public static Builder builder() {
```
andreaturli: [minor] it would be nice to use Auto Value builder available since 1.1 https://github.com/google/auto/tree/master/value#builders

docker/src/main/java/org/jclouds/docker/domain/HostConfig.java
```diff
@@ -79,16 +78,16 @@ public Builder toBuilder() {
    public static final class Builder {
```
andreaturli: same here, see https://github.com/google/auto/tree/master/value#builders

andreaturli:
thanks @grkvlt looks good, just a couple of minor comments

docker/src/main/java/org/jclouds/docker/domain/Config.java
grkvlt: Yes, but moving to AutoValue builders should be a separate PR, I think

grkvlt:
@andreaturli if you're OK with these PRs I suggest I could make a _separate_ patch using the AutoValue builders, since it won't be adding any functionality. That way these can be merged sooner. WDYT?

andreaturli:
Sounds good to me. I'll take care of the builders later

andreaturli:
thanks @grkvlt, merged at [1.9.x](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/dbea9c3e)

PR #254, opened 2016-04-06
Azure ARM fixes

nacx:
This fixes the directory structure of the Azure ARM provider. It contained a directory named `azurecompute.arm` instead of being a hierarchy, which caused the javadoc generation to fail.

I've also added the suggestions mentioned in the first pull request:

* Removed the tenant id property from the pom.xml and added the oauth resource so there is no need to manually set it.
* Updated the build status image to point to the jclouds Jenkins.
* Simplified the README to reflect the generic live tests instructions, to show how to run all them at once (this is what we expect and what we'll put in our CI system).
* Removed the `oauth.endpoint` property from the provider metadata. It can't be preconfigured as it requires the tenant id, so we'd better fail fast and report the missing property instead of silently try to connect to a wrong endpoint if users don't specify the property when creating the context.

@ritazh Comments are very welcome if any of the changes here are not appropriate!


azurecompute-arm/README.md
```diff
@@ -42,68 +43,22 @@ Run the following commands to assign roles to the service principal
...
-mvn -Dtest=OSImageApiLiveTest -Dtest.azurecompute-arm.identity=<Application-id> -Dtest.azurecompute-arm.subscriptionid=<Subscription-id> -Dtest.azurecompute-arm.credential=<password> -Dtest.azurecompute-arm.endpoint="https://management.azure.com" -Dtest.jclouds.oauth.resource="https://management.azure.com/" -Dtest.oauth.endpoint="https://login.microsoftonline.com/<Tenant-id>/oauth2/token" test
-
-# DeploymentApiLiveTest:
-
-mvn -Dtest=DeploymentApiLiveTest -Dtest.azurecompute-arm.identity=<Application-id> -Dtest.azurecompute-arm.subscriptionid=<Subscription-id> -Dtest.azurecompute-arm.credential=<password> -Dtest.azurecompute-arm.endpoint="https://management.azure.com/" -Dtest.jclouds.oauth.resource="https://management.azure.com/" -Dtest.oauth.endpoint="https://login.microsoftonline.com/<Tenant-id>/oauth2/token" test
-
+mvn clean verify -Plive \
+    -Dtest.azurecompute-arm.identity=<Application-id> \
+    -Dtest.azurecompute-arm.credential=<password> \
+    -Dtest.azurecompute-arm.subscriptionid=<Subscription-id>\
```
ritazh: @nacx This is great!!! 
Since we are using the [endpoint property](https://github.com/jclouds/jclouds-labs/blob/master/azurecompute-arm/src/main/java/org/jclouds/azurecompute.arm/AzureComputeProviderMetadata.java#L76), I think we should now use `-Dtest.azurecompute-arm.endpoint=<https://management.azure.com/Subscription-id>` which includes the subscriptionid, instead of `-Dtest.azurecompute-arm.subscriptionid=<Subscription-id>`. WDYT?

nacx: +1. Will change. Live test invocation should be as close as possible to what users will need to provide when creating the context. Thanks for having a look!

nacx:
Comments addressed  @ritazh. Note that the build will fail until https://github.com/jclouds/jclouds/pull/943 is merged, as in the last commit I've already fixed the OAuthScopes binding.

nacx:
Ok, the OAuth changes are merged and deployed to the snapshot repository. Let's see if the build is happy now: rebuild please!

nacx:
Squashed and pushed to master as [17f7c452](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/17f7c452).

nacx:
With the package fix the javadoc and the CI build should be back to green in short.

PR #256, opened 2016-04-12
JCLOUDS-664 Azurecompute-arm LocationApi

ritazh:
@nacx @andreaturli 
This PR is part of the foundations needed for interacting with Azure Resource Manager (ARM) Rest APIs. The Location API is the second API.

FYI
@jmspring @jtjk @isalento


azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/features/LocationApi.java
```diff
@@ -0,0 +1,57 @@
...
+@RequestFilters(OAuthFilter.class)
+@QueryParams(keys = "api-version", values = "2015-11-01")
+@Produces(MediaType.APPLICATION_JSON)
+@Consumes(MediaType.APPLICATION_JSON)
+public interface LocationApi {
+
+   /**
+    * The List Locations operation lists all of the data center locations that are valid for your subscription.
+    */
+   @Named("ListLocations")
```
andreaturli: [minor] jclouds tends to use "location:list"

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/LocationApiLiveTest.java
```diff
@@ -0,0 +1,39 @@
...
+
+import static org.testng.Assert.assertTrue;
+
+@Test(groups = "live", testName = "LocationApiLiveTest")
+public class LocationApiLiveTest extends BaseAzureComputeApiLiveTest {
+
+   @Test
+   public void testList() {
+      for (Location location : api().list()) {
+          assertTrue(!location.id().isEmpty());
```
andreaturli: could you please add a "checkLocation(location);" where you assertNotNull on all the fields of Location?

andreaturli:
great job @ritazh 
only few minor comments from my side: please address them so we can merge this PR as well.

ritazh:
@andreaturli @nacx Please review the latest.

andreaturli:
great! merging now, thanks @ritazh

andreaturli:
merged at [master](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/559fdecf)

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/Region.java
```diff
@@ -0,0 +1,90 @@
...
+      return iso3166Code;
+   }
+
+   public static Region byName(final String name) {
+      Preconditions.checkNotNull(name);
+
+      Region result = null;
+      for (Region region : values()) {
+         if (name.equals(region.name)) {
+            result = region;
```
nacx: [minor] Just `return region` here to avoid unnecessary iterations.

nacx:
Thanks @ritazh! This looks great.

One small question regarding the future implementation of the ComputeService interface: I've seen you've populated the ISO code list in the metadata (great!). When implementing the ComputeService, there are two approaches to get the list of available locations:

* Dynamically, in the [ComputeServiceAdapter#listLocations](https://github.com/jclouds/jclouds/blob/master/compute/src/main/java/org/jclouds/compute/ComputeServiceAdapter.java#L129) method (like in the [DigitalOcean2 provider](https://github.com/jclouds/jclouds/blob/master/providers/digitalocean2/src/main/java/org/jclouds/digitalocean2/compute/DigitalOcean2ComputeServiceAdapter.java#L165-L174)).
* Statically, by providing the location properties, like in AWS or [ProfitBricks](https://github.com/jclouds/jclouds/blob/master/providers/profitbricks/src/main/java/org/jclouds/profitbricks/ProfitBricksProviderMetadata.java#L64-L74).

Do all subscriptions have access to all regions in the Region enum? And are they generally available? If that is the case, we might want to preconfigure the regions in the metadata like in AWS or ProfitBricks, as that gives users more control by letting them override the properties to make jclouds just work with a specific subset of regions (this has some use cases, for example [JCLOUDS-540](https://issues.apache.org/jira/browse/JCLOUDS-540).

Which do you think would be the best approach for the ARM provider?

ritazh:
@andreaturli Seems 559fdec is missing changes made based on your feedback from the latest commit in PR #256. I will create another PR for those changes and adding @nacx's suggestion for `Region`.

ritazh:
@nacx Regarding list of locations, from the official Azure REST API [doc](https://msdn.microsoft.com/en-US/library/azure/dn790540.aspx):
"This API provides all of the locations that are available for resource providers; however, each resource provider may support a subset of this list." It might make more sense to do this dynamically since it will change over time. 

Here is how we are implementing [listLocations](https://github.com/ritazh/jclouds-labs/blob/azurecompute-arm-dev/azurecompute-arm-dev/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java#L416-L418) in `AzureComputeServiceAdapter.java` now. There is also [LocationToLocation.java](https://github.com/ritazh/jclouds-labs/blob/azurecompute-arm-dev/azurecompute-arm-dev/src/main/java/org/jclouds/azurecompute/arm/compute/functions/LocationToLocation.java). WDYT?

nacx:
That looks fine. +1 to implement it dynamically given the docs. Thanks for sharing!

PR #259, opened 2016-04-13
JCLOUDS-664 Azurecompute-arm StorageAccountApi

ritazh:
@nacx @andreaturli 
This PR is for StorageAccountApi, part of the foundations needed for interacting with Azure Resource Manager (ARM) Rest APIs.

FYI
@jmspring @jtjk @isalento


azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/CreateStorageServiceParams.java
```diff
@@ -0,0 +1,65 @@
...
+    * characters.
+    */
+   @Nullable
+   public abstract Map<String, String> properties();
+
+   @SerializedNames({"location", "tags", "properties"})
+   public static CreateStorageServiceParams create(
+           final String location, @Nullable final Map<String, String> tags, @Nullable final Map<String, String> properties) {
+
+      return new AutoValue_CreateStorageServiceParams(location, tags == null ? ImmutableMap.<String, String>builder().build() : ImmutableMap.copyOf(tags), properties == null ? ImmutableMap.<String, String>builder().build() : ImmutableMap.copyOf(properties));
```
nacx: Since you're enforcing non-null maps here, remove the `@Nullable` annotations from the variables.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/StorageService.java
```diff
@@ -0,0 +1,208 @@
...
+   public enum AccountType {
+
+      Standard_LRS,
+      Standard_ZRS,
+      Standard_GRS,
+      Standard_RAGRS,
+      Premium_LRS,
+      UNRECOGNIZED;
+
+      public static AccountType fromString(final String text) {
```
nacx: Better rename to `fromValue` so we can benefit from it during deserialization. jclouds will call the `fromValue` when deserializing values if the default `enum.valueOf` fails (see [EnumTypeAdapterThatReturnsFromValue](https://github.com/jclouds/jclouds/blob/master/core/src/main/java/org/jclouds/json/internal/EnumTypeAdapterThatReturnsFromValue.java), the default enum type adapter used in jclouds).

I've just noticed this, but it would be great if you could rename any other similar methods in the enums introduced in the other ARM PRs, if any.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/StorageService.java
```diff
@@ -0,0 +1,208 @@
...
+         if (text != null) {
+            for (AccountType type : AccountType.values()) {
+               if (text.equalsIgnoreCase(type.name())) {
+                  return type;
+               }
+            }
+         }
+         return UNRECOGNIZED;
+      }
+   }
```
nacx: This method is repeated in every enum. Consider creating a helper class with method to encapsulate this logic. Something like:
```java
@SuppressWarnings("unchecked")
public static <T extends Enum<T>> Enum<T> fromValueOrDefault(String text, Enum<T> defaultValue) {
   if (text != null) {
      EnumSet<T> elements = EnumSet.allOf(defaultValue.getClass());
      for (Enum<T> element : elements) {
         if (text.equalsIgnoreCase(element.name())) {
            return element;
         }
      }
   }
   return defaultValue;
}
```

And then just call it from all enums:
```java
public static AccountType fromValue(final String text) {
   return fromValueOrDefault(text, AccountType.UNRECOGNIZED);
}
```

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/StorageService.java
```diff
@@ -0,0 +1,208 @@
...
+
+   public enum RegionStatus {
+
+       Available,
+       Unavailable,
+       available,
+       unavailable,
```
nacx: Are the lower case versions really needed?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/StorageService.java
```diff
@@ -0,0 +1,208 @@
...
+      @SerializedNames({"accountType", "creationTime", "primaryEndpoints",  "primaryLocation",
+              "provisioningState", "secondaryEndpoints", "secondaryLocation", "statusOfPrimary", "statusOfSecondary"})
+      public static StorageServiceProperties create(final AccountType accountType, final Date creationTime,
+              final Map<String, String> primaryEndpoints, final String primaryLocation, final Status provisioningState,
+              final Map<String, String> secondaryEndpoints, final String secondaryLocation,
+              final RegionStatus statusOfPrimary, final RegionStatus statusOfSecondary) {
+
+         return new AutoValue_StorageService_StorageServiceProperties(accountType, creationTime,
+                 primaryEndpoints, primaryLocation, provisioningState,
+                 secondaryEndpoints, secondaryLocation, statusOfPrimary, statusOfSecondary);
```
nacx: Do we want to have null maps here? Can't we follow the same pattern and default to empty ones?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/StorageService.java
```diff
@@ -0,0 +1,208 @@
...
+    * Specifies the properties of the storage account.
+    */
+   public abstract StorageServiceProperties storageServiceProperties();
+
+
+   @SerializedNames({"id", "name", "location", "tags", "type", "properties"})
+   public static StorageService create(final String id,  final String name,  final String location,
+                                       @Nullable final Map<String, String> tags,  final String type,
+                                       final StorageServiceProperties storageServiceProperties) {
+      return new AutoValue_StorageService(id,  name,  location,  tags == null ? ImmutableMap.<String, String>builder().build() : ImmutableMap.copyOf(tags), type, storageServiceProperties);
```
nacx: Remove the `@Nullable` annotation from the `tags` member variable.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/StorageServiceUpdateParams.java
```diff
@@ -0,0 +1,185 @@
...
+         if (text != null) {
+            for (Status status : Status.values()) {
+               if (text.equalsIgnoreCase(status.name())) {
+                  return status;
+               }
+            }
+         }
+         return UNRECOGNIZED;
+      }
+   }
```
nacx: These enums are the same than the previous ones. Extract them to top-level type classes.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/StorageServiceUpdateParams.java
```diff
@@ -0,0 +1,185 @@
...
+
+   @Nullable
```
nacx: I'm not sure if this annotation has any effect when configured at class level?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/StorageServiceUpdateParams.java
```diff
@@ -0,0 +1,185 @@
...
+      @SerializedNames({"accountType", "creationTime", "primaryEndpoints", "primaryLocation",
+              "provisioningState", "secondaryEndpoints", "secondaryLocation", "statusOfPrimary", "statusOfSecondary"})
+      public static StorageServiceUpdateProperties create(final AccountType accountType, final Date creationTime,
+              final Map<String, String> primaryEndpoints, final String primaryLocation, final Status provisioningState,
+              final Map<String, String> secondaryEndpoints, final String secondaryLocation,
+              final RegionStatus statusOfPrimary, final RegionStatus statusOfSecondary) {
+
+         return new AutoValue_StorageServiceUpdateParams_StorageServiceUpdateProperties(accountType, creationTime,
+                 primaryEndpoints == null ? null : ImmutableMap.copyOf(primaryEndpoints), primaryLocation, provisioningState,
+                 secondaryEndpoints == null ? null : ImmutableMap.copyOf(secondaryEndpoints), secondaryLocation, statusOfPrimary, statusOfSecondary);
```
nacx: Same about nullable Maps. Remove the annotation if they're initialized to empty.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/StorageServiceUpdateParams.java
```diff
@@ -0,0 +1,185 @@
...
+      }
+   }
+
+   /**
+    * Specifies the tags of the storage account.
+    */
+   @Nullable
```
nacx: Remove.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/features/StorageAccountApi.java
```diff
@@ -0,0 +1,160 @@
...
+
+/**
+ * The Azure Resource Management API includes operations for managing the storage accounts in your subscription.
+ *
+ * @see <a href="https://msdn.microsoft.com/en-us/library/mt163683.aspx">docs</a>
+ */
+@Path("/")
+@RequestFilters(OAuthFilter.class)
+@QueryParams(keys = "api-version", values = "2015-06-15")
+@Produces(MediaType.APPLICATION_JSON)
```
nacx: This will set the Content-Type header, which is not used in GET methods. Move this annotation to just the methods that generate a body.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/features/StorageAccountApi.java
```diff
@@ -0,0 +1,160 @@
...
+
+   /**
+    * https://msdn.microsoft.com/en-us/library/mt163652.aspx
+    * DELETE
+    */
+   @Named("storageaccount:delete")
+   @DELETE
+   @ResponseParser(StatusCodeParser.class)
+   @Path("/resourcegroups/{resourceGroup}/providers/Microsoft.Storage/storageAccounts/{storageAccountName}")
+   String delete(@PathParam("storageAccountName") String storageAccountName);
```
nacx: Looking at the docs, it does not seem to return a URL to track an ongoing delete operation; just 200 or 204. WDYT about just returning `void` and remove the response parser? Any other response status code would then throw an exception.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/StorageAccountApiLiveTest.java
```diff
@@ -0,0 +1,134 @@
...
+   }
+
+   @Test()
+   public void testListAll() {
+      List<StorageService> storages = api().list();
+      assertTrue(storages.size() > 0);
+      for (StorageService storage : storages) {
+         check(storage);
+      }
+   }
```
nacx: This test is repeated.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/StorageAccountApiLiveTest.java
```diff
@@ -0,0 +1,134 @@
...
+   private void check(final StorageService storage) {
+      assertNotNull(storage.id());
+      assertNotNull(storage.name());
+      assertNotNull(storage.storageServiceProperties());
+      assertNotNull(storage.storageServiceProperties().accountType());
+      assertFalse(storage.storageServiceProperties().primaryEndpoints().isEmpty());
+      assertNotNull(storage.storageServiceProperties().creationTime());
+   }
+
+   @Test()
```
nacx: Should this better depend on the create test, to make sure there is content?

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/StorageAccountApiLiveTest.java
```diff
@@ -0,0 +1,134 @@
...
+              "property_value"), ImmutableMap.of("accountType", StorageService.AccountType.Standard_ZRS.toString()));
+      while (storage == null) {
+         try {
+            Thread.sleep(25 * 1000);
+         } catch (InterruptedException e) {
+            e.printStackTrace();
+         }
+         storage = api().create(NAME, LOCATION, ImmutableMap.of("property_name", "property_value"),
+                 ImmutableMap.of("accountType", StorageService.AccountType.Standard_ZRS.toString()));
+      }
```
nacx: Which are the semantics of the create operations? It looks kinda weird to attempt the create operation N times instead of calling a `get` operation when waiting.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/StorageAccountApiMockTest.java
```diff
@@ -0,0 +1,288 @@
...
+      server.enqueue(jsonResponse("/storageAccounts.json"));
+
+      final StorageAccountApi storageAPI = api.getStorageAccountApi(resourceGroup);
+
+      List<StorageService> list = storageAPI.list();
+      assertEquals(list, expected());
+
+      assertSent(server, "GET", "/subscriptions/" + subsriptionId +
+              "/resourcegroups/resourceGroup/providers/Microsoft.Storage/storageAccounts?api-version=2015-06-15");
+   }
```
nacx: Add a test that exercises the 404 fallback for the list method.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/StorageAccountApiMockTest.java
```diff
@@ -0,0 +1,288 @@
...
+      final StorageAccountApi storageAPI = api.getStorageAccountApi(resourceGroup);
+
+      assertEquals(storageAPI.create("name-of-storage-account", "westus",
+              ImmutableMap.of("property_name", "property_value"),
+              ImmutableMap.of("accountType", StorageService.AccountType.Premium_LRS.toString())),
+              getCreateResponse());
+
+      assertSent(server, "PUT", "/subscriptions/" + subsriptionId +
+              "/resourcegroups/resourceGroup/providers/Microsoft.Storage/" +
+              "storageAccounts/name-of-storage-account?api-version=2015-06-15");
```
nacx: For methods that generate a body we also need to verify that the generated body is the expected one.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/StorageAccountApiMockTest.java
```diff
@@ -0,0 +1,288 @@
...
+   public void testIsAvailable() throws Exception {
+      server.enqueue(jsonResponse("/isavailablestorageservice.json"));
+
+      final StorageAccountApi storageAPI = api.getStorageAccountApi(resourceGroup);
+
+      assertEquals(storageAPI.isAvailable("TESTSTORAGE"),
+              Availability.create("true"));
+
+      assertSent(server, "POST", "/subscriptions/" + subsriptionId +
+              "/providers/Microsoft.Storage/checkNameAvailability?api-version=2015-06-15");
```
nacx: Also verify the body here.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/StorageAccountApiMockTest.java
```diff
@@ -0,0 +1,288 @@
...
+      server.enqueue(jsonResponse("/storageaccountkeys.json"));
+
+      final StorageAccountApi storageAPI = api.getStorageAccountApi(resourceGroup);
+
+      assertEquals(storageAPI.regenerateKeys("TESTSTORAGE", "key1"), StorageServiceKeys.create(
+              "bndO7lydwDkMo4Y0mFvmfLyi2f9aZY7bwfAVWoJWv4mOVK6E9c/exLnFsSm/NMWgifLCfxC/c6QBTbdEvWUA7w==",
+              "/jMLLT3kKqY4K+cUtJTbh7pCBdvG9EMKJxUvaJJAf6W6aUiZe1A1ulXHcibrqRVA2RJE0oUeXQGXLYJ2l85L7A=="));
+
+      assertSent(server, "POST", "/subscriptions/" + subsriptionId + "/resourcegroups/" + resourceGroup +
+              "/providers/Microsoft.Storage/storageAccounts/TESTSTORAGE/regenerateKey?api-version=2015-06-15");
```
nacx: Add body check.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/StorageAccountApiMockTest.java
```diff
@@ -0,0 +1,288 @@
...
+      StorageServiceUpdateParams.StorageServiceUpdateProperties.create(null, null, null, null, null, null, null, null,
+              null);
+
+      final StorageServiceUpdateParams params = storageAPI.update("TESTSTORAGE", props,
+              ImmutableMap.of("another_property_name", "another_property_value"));
+
+      assertTrue(params.tags().containsKey("another_property_name"));
+
+      assertSent(server, "PATCH", "/subscriptions/" + subsriptionId + "/resourcegroups/" + resourceGroup +
+         "/providers/Microsoft.Storage/storageAccounts/TESTSTORAGE?api-version=2015-06-15");
```
nacx: Add body check.

nacx:
Thanks @ritazh!

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/features/StorageAccountApi.java
ritazh: @nacx If we just return void, we are not preserving the response from the server. What if users want to check the status of a DELETE operation? Returning 204 would let them know the storage account they are trying to delete no longer exist? Just a thought...

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/StorageAccountApiLiveTest.java
ritazh: Using `URIParser` and `ParseJobStatus` since this returns 202 then 200 when done. 
```java
URI uri = api().create(NAME, LOCATION, ImmutableMap.of("property_name",
              "property_value"), ImmutableMap.of("accountType", StorageService.AccountType.Standard_ZRS.toString()));
      if (uri != null){
         assertTrue(uri.toString().contains("api-version"));

         boolean jobDone = Predicates2.retry(new Predicate<URI>() {
            @Override public boolean apply(URI uri) {
               return ParseJobStatus.JobStatus.DONE == api.getJobApi().jobStatus(uri);
            }
         }, 60 * 1 * 1000 /* 1 minute timeout */).apply(uri);
         assertTrue(jobDone, "delete operation did not complete in the configured timeout");
      }
      final StorageService service = api().get(NAME);
```

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/StorageAccountApiLiveTest.java
ritazh: Removed.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/StorageAccountApiMockTest.java
ritazh: Added.

ritazh:
@nacx Made updates per your feedback. Ready for review. 😄

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/features/StorageAccountApi.java
nacx: The pattern we use in most APIs is to let the delete method return a boolean, and then use a fallback like "false on 404". You could change the method in that way to return false on 204.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/StorageService.java
```diff
@@ -0,0 +1,187 @@
...
+
+      Standard_LRS,
+      Standard_ZRS,
+      Standard_GRS,
+      Standard_RAGRS,
+      Premium_LRS,
+      UNRECOGNIZED;
+
+       public static AccountType fromValue(final String text) {
+           return (AccountType) GetEnumValue.fromValueOrDefault(text, AccountType.UNRECOGNIZED);
```
nacx: Is the cast needed? The type should be already resolved by the second parameter.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/StorageService.java
```diff
@@ -0,0 +1,187 @@
...
+
+       /**
+        * Specifies the time that the storage account was created.
+        */
+       public abstract Date creationTime();
+
+       /**
+        * Specifies the endpoints of the storage account.
+        */
+       @Nullable
```
nacx: Remove this annotation

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/StorageService.java
```diff
@@ -0,0 +1,187 @@
...
+      /**
+       * provisioningState for the storage group
+       */
+      @Nullable
+      public abstract Status provisioningState();
+
+       /**
+        * Specifies the secondary endpoints of the storage account.
+        */
+       @Nullable
```
nacx: Remove this annotation

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/StorageService.java
```diff
@@ -0,0 +1,187 @@
...
+   /**
+    * Specifies the location of the storage account.
+    */
+   @Nullable
+   public abstract String location();
+
+   /**
+    * Specifies the tags of the storage account.
+    */
+   @Nullable
```
nacx: Remove this annotation

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/StorageServiceUpdateParams.java
nacx: Still needs to be removed.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/features/StorageAccountApi.java
nacx: Also consider removing the StatusCodeParser class if we go for this approach and it is then not used.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/StorageAccountApiLiveTest.java
```diff
@@ -0,0 +1,122 @@
...
+              "property_value"), ImmutableMap.of("accountType", StorageService.AccountType.Standard_ZRS.toString()));
+      if (uri != null){
+         assertTrue(uri.toString().contains("api-version"));
+
+         boolean jobDone = Predicates2.retry(new Predicate<URI>() {
+            @Override public boolean apply(URI uri) {
+               return ParseJobStatus.JobStatus.DONE == api.getJobApi().jobStatus(uri);
+            }
+         }, 60 * 1 * 1000 /* 1 minute timeout */).apply(uri);
+         assertTrue(jobDone, "delete operation did not complete in the configured timeout");
```
nacx: "create operation"

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/internal/BaseAzureComputeApiLiveTest.java
```diff
@@ -56,10 +81,45 @@ private void deleteResourceGroup(String name) {
...
+              "property_value"), ImmutableMap.of("accountType", StorageService.AccountType.Standard_ZRS.toString()));
+      if (uri != null){
+         assertTrue(uri.toString().contains("api-version"));
+
+         boolean jobDone = Predicates2.retry(new Predicate<URI>() {
+            @Override public boolean apply(URI uri) {
+               return ParseJobStatus.JobStatus.DONE == api.getJobApi().jobStatus(uri);
+            }
+         }, 60 * 1 * 1000 /* 1 minute timeout */).apply(uri);
+         assertTrue(jobDone, "delete operation did not complete in the configured timeout");
```
nacx: "create operation"

nacx:
Some final comments. Thanks @ritazh!

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/StorageService.java
ritazh: Yes...compiler is returning: "error: incompatible types: Enum<AccountType> cannot be converted to AccountType" without it.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/StorageService.java
ritazh: @nacx 
So I did remove this annotation in this class. But in order to support create and update requests without tags, I added `@Nullable` in the [StorageAccountApi](https://github.com/ritazh/jclouds-labs/blob/azurecomputearm-storageaccountapi/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/features/StorageAccountApi.java#L90).

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/features/StorageAccountApi.java
ritazh: @nacx After adding `@Fallback(Fallbacks.FalseOnNotFoundOr204.class)` for delete, in `StorageAccountApiMockTest`, setting response code to 204 does not seem to work. Same test for `FalseOnNotFoundOr204` works fine. Similarly `FalseOnNotFoundOr422` also never returns false. What am I missing here?

```java
server.enqueue(new MockResponse().setResponseCode(204));

      final StorageAccountApi storageAPI = api.getStorageAccountApi(resourceGroup);

      boolean status = storageAPI.delete("TESTSTORAGE");
      assertFalse(status);
```

jmspring:
@nacx Using 204 for a fallback is not going to work, from my understanding of the code.  If you step through BaseHttpCommandExecutorService.invoke where the HTTP request is made.  Only in the path where an error code is not 2XX, is the path to check for fallbacks triggered.

Specifically, the BaseHttpCommandExecutorService.shouldContinue method checks for the need to continue or not, if not, an error is generated and that error is then propagated back at the end of the BaseHttpCommandExecutorService.invoke method and the Fallback triggered.

Thoughts?

jmspring:
Also @nacx - I think there is a benefit of being able to distinguish between a 200 and 204, mostly in the case to allow the caller to maybe catch the case where the resource specified should be there, but maybe the name was wrong.

nacx:
> Thoughts?

Oh, you're right, and it's my fault for not having properly advised and later detected it in the review. Fallbacks are for "failed" requests, and Response Parsers are for successful ones. In this case, to go for the boolean approach we should simply leave the method wihtout any fallback, as jclouds will [return true for 2xx responses](https://github.com/jclouds/jclouds/blob/master/core/src/main/java/org/jclouds/rest/internal/TransformerForRequest.java#L116-L117). By removing the fallback we'd let the method to fail otherwise. Another option would be to create a ["FalseOn204" response parser function](https://github.com/jclouds/jclouds/blob/master/core/src/main/java/org/jclouds/rest/annotations/ResponseParser.java) and annotate the method accordingly, so it returns `false` on a 204 response, `true` on 200, and an exception otherwise.

WDYT about implementing it as a response parser? (And apologies for my previous bad advice).

ritazh:
@nacx Latest commit includes `@ResponseParser(FalseOn204.class)` for `delete`. Ready for another round of review. Thanks!

nacx:
Pushed to master as [b82378c2](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/b82378c2). Thanks!

nacx:
I've also pushed [jclouds#e5f65a47](http://git-wip-us.apache.org/repos/asf/jclouds/commit/e5f65a47) to remove the fallback for the 204, as it won't never apply. Apologies for my confusion!

PR #260, opened 2016-04-14
Make DockerTemplateOptions values null safe

grkvlt:
Do not reset `entrypoint` or `commands` options to an empty list, which overrides any Dockerfile provided values.

### DIFF
--- a/docker/src/main/java/org/jclouds/docker/compute/options/DockerTemplateOptions.java
+++ b/docker/src/main/java/org/jclouds/docker/compute/options/DockerTemplateOptions.java
@@ -83,15 +83,15 @@ public class DockerTemplateOptions extends TemplateOptions implements Cloneable 
    private static final String NO_IMAGE = "jclouds-placeholder-for-image";
 
    protected List<String> dns = ImmutableList.of();
-   protected String hostname;
-   protected Integer memory;
-   protected Integer cpuShares;
-   protected List<String> entrypoint = ImmutableList.of();
-   protected List<String> commands = ImmutableList.of();
+   @Nullable protected String hostname;
+   @Nullable protected Integer memory;
+   @Nullable protected Integer cpuShares;
+   @Nullable List<String> entrypoint;
+   @Nullable List<String> commands;
    protected Map<String, String> volumes = ImmutableMap.of();
-   protected List<String> env = ImmutableList.of();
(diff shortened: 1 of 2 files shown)


grkvlt:
@andreaturli This and #261 are more changes needed for Clocker

docker/src/main/java/org/jclouds/docker/compute/options/DockerTemplateOptions.java
```diff
@@ -83,15 +83,15 @@
...
-   protected String hostname;
-   protected Integer memory;
-   protected Integer cpuShares;
-   protected List<String> entrypoint = ImmutableList.of();
-   protected List<String> commands = ImmutableList.of();
+   @Nullable protected String hostname;
+   @Nullable protected Integer memory;
+   @Nullable protected Integer cpuShares;
+   @Nullable List<String> entrypoint;
+   @Nullable List<String> commands;
```
nacx: Out of curiosity, does enforcing empty lists here (and never null) cause an undesired behaviour when creating containers using the docker compute service adapter?

grkvlt: No, we do an `isEmpty()` check on the non `@Nullable` values before setting them in the `createContainer` API call

nacx: Which is the purpose of making them nullable then? Having null collections is a bad pattern and a potential source of NPEs; we should try to avoid them.

nacx:
Apologies for the delay. I've finally had time to get back to this and properly read the Brooklyn issue and understand the fix. Pushed to master as [44ddd688](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/44ddd688). Thanks @grkvlt!

docker/src/main/java/org/jclouds/docker/compute/options/DockerTemplateOptions.java
grkvlt: FYI, the JSON that the Docker API recieves will treat null differently to an empty list. So for the entrypoint value, null means 'do nothing, just use the values from the image' but an empty list means 'change this value to an empty list, and discard the value from the image' which is why we need to be `@Nullable` here.

PR #261, opened 2016-04-14
Make DockerTemplateOptions values null safe

grkvlt:
Do not reset `entrypoint` or `commands` options to an empty list, which overrides any Dockerfile provided values.

Same as #260 but rebased for `1.9.x` branch.


nacx:
Pushed to 1.9.x as [5c892910](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/5c892910). Thanks @grkvlt!

PR #263, opened 2016-04-25
Add Docker template option and update configuration

grkvlt:
- Add `openStdin` option to `DockerTemplateOptions`
- Set all port bindings explicitly on container creation


nacx:
Pushed to master as [1eed49f6](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/1eed49f6). Thanks @grkvlt!

PR #265, opened 2016-04-27
JCLOUDS-664 Azurecompute-arm NicApi PublicIPApi and VMApi

ritazh:
@nacx @andreaturli 
This PR is for `NetworkInterfaceCardApi`, `PublicIPAddressApi`, and `VirtualMachineApi`, part of the foundations needed for interacting with Azure Resource Manager (ARM) Rest APIs.

FYI
@jtjk @isalento @jmspring

### DIFF
--- a/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/AvailabilitySet.java
+++ b/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/AvailabilitySet.java
@@ -1,0 +1,104 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jclouds.azurecompute.arm.domain;
+
+import com.google.auto.value.AutoValue;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import org.jclouds.javax.annotation.Nullable;
+import org.jclouds.json.SerializedNames;
+import java.util.Map;
+import java.util.List;
+
+/**
+ * AvailabilitySet for subscription
+ */
+@AutoValue
+public abstract class AvailabilitySet {
+
+   @AutoValue
+   public abstract static class AvailabilitySetProperties {
+
+      /**
+       * A platform Update Domain Count
+       */
+      public abstract int platformUpdateDomainCount();
+
+      /**
+       * A platform Fault Domain Count
+       */
+      public abstract int platformFaultDomainCount();
+
+      /**
+       * A list of virtual machines in availability set
+       */
+      @Nullable
+      public abstract List<AvailabilitySetVirtualMachine> virtualMachines();
+
+      @SerializedNames({"platformUpdateDomainCount", "platformFaultDomainCount", "virtualMachines"})
+      public static AvailabilitySetProperties create(final int platformUpdateDomainCount,
+                                                     final int platformFaultDomainCount,
+                                                     List<AvailabilitySetVirtualMachine> virtualMachines) {
+         return new AutoValue_AvailabilitySet_AvailabilitySetProperties(platformUpdateDomainCount,
+                 platformFaultDomainCount,
+                 virtualMachines == null ? null : ImmutableList.copyOf(virtualMachines));
+      }
+   }
+
+   /**
+    * The id of the availability set
+    */
+   public abstract String id();
+
+   /**
+    * The name of the availability set.
+    */
+   @Nullable
+   public abstract String name();
+
+   /**
+    * The name of the availability set.
+    */
+   @Nullable
+   public abstract String type();
+
+   /**
+    * The location of the availability set
+    */
+   @Nullable
+   public abstract String location();
+
+   /**
+    * Specifies the tags of the availability set
+    */
+   @Nullable
+   public abstract Map<String, String> tags();
+
+   /**
+    * Specifies the properties of the availability set
+    */
+   @Nullable
+   public abstract AvailabilitySetProperties properties();
+
+
+   @SerializedNames({"id", "name", "type", "location", "tags", "properties"})
+   public static AvailabilitySet create(final String id, final String name, final String type, final String location,
+                                        final Map<String, String> tags, AvailabilitySetProperties properties) {
+      return new AutoValue_AvailabilitySet(id, name, type, location, tags == null ? null : ImmutableMap.copyOf(tags), properties);
+   }
+}
--- a/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/DiagnosticsProfile.java
+++ b/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/DiagnosticsProfile.java
@@ -1,0 +1,68 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jclouds.azurecompute.arm.domain;
+
+import com.google.auto.value.AutoValue;
+import org.jclouds.javax.annotation.Nullable;
+import org.jclouds.json.SerializedNames;
+
+@AutoValue
+public abstract class DiagnosticsProfile {
+
+   @AutoValue
+   public abstract static class BootDiagnostics{
+
+      public abstract boolean enabled();
+
+      @Nullable
+      public abstract String storageUri();
+
+      @SerializedNames({"enabled", "storageUri"})
+      public static BootDiagnostics create(final boolean enabled, final String storageUri) {
+         return builder()
+                 .enabled(enabled)
+                 .storageUri(storageUri)
+                 .build();
+      }
+
+      public static Builder builder() {
+         return new AutoValue_DiagnosticsProfile_BootDiagnostics.Builder();
+      }
+      @AutoValue.Builder
+      public abstract static class Builder {
+         public abstract Builder enabled(boolean enabled);
+         public abstract Builder storageUri(String storageUri);
+         public abstract BootDiagnostics build();
+      }
+   }
+
+   public abstract BootDiagnostics bootDiagnostics();
+
+   @SerializedNames({"bootDiagnostics"})
+   public static DiagnosticsProfile create(final BootDiagnostics  bootDiagnostics) {
+      return builder().bootDiagnostics(bootDiagnostics).build();
+   }
+   public static Builder builder() {
+      return new AutoValue_DiagnosticsProfile.Builder();
+   }
+
+   @AutoValue.Builder
+   public abstract static class Builder {
+      public abstract Builder bootDiagnostics(BootDiagnostics bootDiagnostics);
+      public abstract DiagnosticsProfile build();
+   }
+}
--- a/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/NetworkInterfaceCardProperties.java
+++ b/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/NetworkInterfaceCardProperties.java
@@ -1,0 +1,78 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jclouds.azurecompute.arm.domain;
+
+import com.google.auto.value.AutoValue;
+import com.google.common.collect.ImmutableList;
+import org.jclouds.javax.annotation.Nullable;
+import org.jclouds.json.SerializedNames;
+
+import java.util.List;
+
+@AutoValue
+public abstract class NetworkInterfaceCardProperties {
+
+   @Nullable
+   public abstract String provisioningState();
+
+   @Nullable
+   public abstract String resourceGuid();
+
+   @Nullable
+   public abstract Boolean enableIPForwarding();
+
+   @Nullable
+   public abstract List<IpConfiguration> ipConfigurations();
+
+   @SerializedNames({"provisioningState", "resourceGuid", "enableIPForwarding", "ipConfigurations"})
+   public static NetworkInterfaceCardProperties create(final String provisioningState, final String resourceGuid, final Boolean enableIPForwarding, final List<IpConfiguration> ipConfigurations) {
+      NetworkInterfaceCardProperties.Builder builder = NetworkInterfaceCardProperties.builder()
+              .provisioningState(provisioningState)
+              .resourceGuid(resourceGuid)
+              .enableIPForwarding(enableIPForwarding)
+              .ipConfigurations(ipConfigurations == null ? null : ImmutableList.copyOf(ipConfigurations));
+
+      return builder.build();
+   }
+
+   public static Builder builder() {
+
+      return new AutoValue_NetworkInterfaceCardProperties.Builder();
+   }
+
+   @AutoValue.Builder
+   public abstract static class Builder {
+      public abstract Builder provisioningState(String provisioningState);
+
+      public abstract Builder resourceGuid(String resourceGuid);
+
+      public abstract Builder enableIPForwarding(Boolean enableIPForwarding);
+
+      public abstract Builder ipConfigurations(List<IpConfiguration> ipConfigurations);
+
+      abstract List<IpConfiguration> ipConfigurations();
+
+      abstract NetworkInterfaceCardProperties autoBuild();
+
+      public NetworkInterfaceCardProperties build() {
+         ipConfigurations(ipConfigurations() != null ? ImmutableList.copyOf(ipConfigurations()) : null);
+         return autoBuild();
+      }
+   }
+}
+
+
--- a/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/NetworkProfile.java
+++ b/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/NetworkProfile.java
@@ -1,0 +1,55 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jclouds.azurecompute.arm.domain;
+
+import com.google.auto.value.AutoValue;
+import com.google.common.collect.ImmutableList;
+import org.jclouds.json.SerializedNames;
+
+import java.util.List;
+
+@AutoValue
+public abstract class NetworkProfile {
+
+   /**
+    * List of network interfaces
+    */
+   public abstract List<IdReference> networkInterfaces();
+
+   @SerializedNames({"networkInterfaces"})
+   public static NetworkProfile create(final List<IdReference> networkInterfaces) {
+      return builder().networkInterfaces(networkInterfaces).build();
+   }
+
+   public static Builder builder() {
+      return new AutoValue_NetworkProfile.Builder();
+   }
+
+   @AutoValue.Builder
+   public abstract static class Builder {
+      public abstract Builder networkInterfaces(List<IdReference> networkInterfaces);
+
+      abstract List<IdReference> networkInterfaces();
+
+      abstract NetworkProfile autoBuild();
+
+      public NetworkProfile build() {
+         networkInterfaces(networkInterfaces() != null ? ImmutableList.copyOf(networkInterfaces()) : ImmutableList.<IdReference>of());
+         return autoBuild();
+      }
+   }
+}
--- a/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/OSProfile.java
+++ b/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/OSProfile.java
@@ -1,0 +1,236 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jclouds.azurecompute.arm.domain;
+
+import com.google.auto.value.AutoValue;
+import com.google.common.collect.ImmutableList;
+import org.jclouds.javax.annotation.Nullable;
+import org.jclouds.json.SerializedNames;
+
+import java.util.List;
+import java.util.Map;
+import com.google.common.collect.ImmutableMap;
+
+@AutoValue
+public abstract class OSProfile {
+
+   @AutoValue
+   public abstract static class LinuxConfiguration {
+
+      @AutoValue
+      public abstract static class SSH {
+
+         @AutoValue
+         public abstract static class SSHPublicKey {
+
+            @Nullable
+            public abstract String path();
+
+            @Nullable
+            public abstract String keyData();
+
+            @SerializedNames({"path", "keyData"})
+            public static SSHPublicKey create(final String path, final String keyData) {
+
+               return new AutoValue_OSProfile_LinuxConfiguration_SSH_SSHPublicKey(
+                       path, keyData);
+            }
+         }
+
+         /**
+          * The list of public keys and paths
+          */
+         @Nullable
+         public abstract List<SSHPublicKey> publicKeys();
+
+         @SerializedNames({"publicKeys"})
+         public static SSH create(final List<SSHPublicKey> publicKeys) {
+
+            return new AutoValue_OSProfile_LinuxConfiguration_SSH(
+                    publicKeys);
+         }
+      }
+
+      /**
+       * The authentication method password or ssh
+       */
+      public abstract String disablePasswordAuthentication();
+
+      /**
+       * ssh keys
+       */
+      @Nullable
+      public abstract SSH ssh();
+
+      @SerializedNames({"disablePasswordAuthentication", "ssh"})
+      public static LinuxConfiguration create(final String disablePasswordAuthentication,
+                                              final SSH ssh) {
+
+         return new AutoValue_OSProfile_LinuxConfiguration(disablePasswordAuthentication,
+                 ssh);
+      }
+   }
+
+   @AutoValue
+   public abstract static class WindowsConfiguration {
+
+      @AutoValue
+      public abstract static class WinRM {
+
+         /**
+          * Map of different settings
+          */
+         public abstract Map<String, String> listeners();
+
+         @SerializedNames({"listeners"})
+         public static WinRM create(final Map<String, String> listeners) {
+            return new AutoValue_OSProfile_WindowsConfiguration_WinRM(listeners == null ? ImmutableMap.<String, String>of() : ImmutableMap.copyOf(listeners));
+         }
+      }
+
+      @AutoValue
+      public abstract static class AdditionalUnattendContent {
+
+         public abstract String pass();
+
+         public abstract String component();
+
+         public abstract String settingName();
+
+         public abstract String content();
+
+         @SerializedNames({"pass", "component", "settingName", "content"})
+         public static AdditionalUnattendContent create(final String pass, final String component,
+                                                        final String settingName,
+                                                        final String content) {
+
+            return new AutoValue_OSProfile_WindowsConfiguration_AdditionalUnattendContent(
+                    pass, component, settingName, content);
+         }
+      }
+
+      /**
+       * The provision VM Agent true of false.
+       */
+      public abstract boolean provisionVMAgent();
+
+      /**
+       * winR
+       */
+      @Nullable
+      public abstract WinRM winRM();
+
+      /**
+       * unattend content
+       */
+      @Nullable
+      public abstract AdditionalUnattendContent additionalUnattendContent();
+
+      /**
+       * is automatic updates enabled
+       */
+      public abstract boolean enableAutomaticUpdates();
+
+      /**
+       * list of certificates
+       */
+      @Nullable
+      public abstract List<String> secrets();
+
+      @SerializedNames({"provisionVMAgent", "winRM", "additionalUnattendContent", "enableAutomaticUpdates",
+              "secrets"})
+      public static WindowsConfiguration create(final boolean provisionVMAgent, final WinRM winRM,
+                                                final AdditionalUnattendContent additionalUnattendContent,
+                                                final boolean enableAutomaticUpdates, final List<String> secrets) {
+
+         return new AutoValue_OSProfile_WindowsConfiguration(provisionVMAgent, winRM,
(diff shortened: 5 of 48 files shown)


azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/AvailabilitySet.java
```diff
@@ -0,0 +1,104 @@
...
+      @Nullable
+      public abstract List<AvailabilitySetVirtualMachine> virtualMachines();
+
+      @SerializedNames({"platformUpdateDomainCount", "platformFaultDomainCount", "virtualMachines"})
+      public static AvailabilitySetProperties create(final int platformUpdateDomainCount,
+                                                     final int platformFaultDomainCount,
+                                                     List<AvailabilitySetVirtualMachine> virtualMachines) {
+         return new AutoValue_AvailabilitySet_AvailabilitySetProperties(platformUpdateDomainCount,
+                 platformFaultDomainCount,
+                 virtualMachines);
```
nacx: Initialize to null or immutable copy

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/DiagnosticsProfile.java
```diff
@@ -0,0 +1,73 @@
...
+import org.jclouds.javax.annotation.Nullable;
+import org.jclouds.json.SerializedNames;
+
+@AutoValue
+public abstract class DiagnosticsProfile {
+
+   @AutoValue
+   public abstract static class BootDiagnostics {
+
+      public abstract Boolean enabled();
```
nacx: If not nullable, change to primitive?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/NetworkInterfaceCardProperties.java
```diff
@@ -0,0 +1,72 @@
...
+   public abstract String provisioningState();
+
+   @Nullable
+   public abstract String resourceGuid();
+
+   @Nullable
+   public abstract Boolean enableIPForwarding();
+
+   @Nullable
+   public abstract ImmutableList<IpConfiguration> ipConfigurations();
```
nacx: Change the type to the `List` interface

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/NetworkInterfaceCardProperties.java
```diff
@@ -0,0 +1,72 @@
...
+
+   @AutoValue.Builder
+   public abstract static class Builder {
+      public abstract Builder provisioningState(String provisioningState);
+
+      public abstract Builder resourceGuid(String resourceGuid);
+
+      public abstract Builder enableIPForwarding(Boolean enableIPForwarding);
+
+      public abstract Builder ipConfigurations(List<IpConfiguration> ipConfigurations);
```
nacx: For builders using lists, we want to enforce immutable lists while keeping clean signatures with `List` interface types. You can use [this pattern](https://github.com/jclouds/jclouds-labs-openstack/blob/master/openstack-heat/src/main/java/org/jclouds/openstack/heat/v1/options/UpdateStack.java#L67-L76) that uses hidden methods in the builder to enforce immutability. Also apply this pattern to all builder methods that receive collection parameters.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/NetworkProfile.java
```diff
@@ -0,0 +1,48 @@
...
+
+import java.util.List;
+
+@AutoValue
+public abstract class NetworkProfile {
+
+   /**
+    * List of network interfaces
+    */
+   public abstract ImmutableList<IdReference> networkInterfaces();
```
nacx: Change type to `List` and enforce immutability in the builder.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/OSProfile.java
```diff
@@ -0,0 +1,235 @@
...
+   @AutoValue
+   public abstract static class WindowsConfiguration {
+
+      @AutoValue
+      public abstract static class WinRM {
+
+         /**
+          * Map of different settings
+          */
+         public abstract Map<String, String> listeners();
```
nacx: Initialize it if non-nullable.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/OSProfile.java
```diff
@@ -0,0 +1,235 @@
...
+      public abstract List<String> secrets();
+
+      @SerializedNames({"provisionVMAgent", "winRM", "additionalUnattendContent", "enableAutomaticUpdates",
+              "secrets"})
+      public static WindowsConfiguration create(final boolean provisionVMAgent, final WinRM winRM,
+                                                final AdditionalUnattendContent additionalUnattendContent,
+                                                final boolean enableAutomaticUpdates, final List<String> secrets) {
+
+         return new AutoValue_OSProfile_WindowsConfiguration(provisionVMAgent, winRM,
+                 additionalUnattendContent, enableAutomaticUpdates, secrets);
```
nacx: Make secrets an immutable list if present.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/StorageProfile.java
```diff
@@ -0,0 +1,73 @@
...
+   /**
+    * The image reference of the storage profile
+    */
+   public abstract OSDisk osDisk();
+
+   /**
+    * The list of the data disks of the storage profile
+    */
+   @Nullable
+   public abstract ImmutableList<DataDisk> dataDisks();
```
nacx: Apply the immutable collection build pattern here too

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineInstance.java
```diff
@@ -0,0 +1,71 @@
...
+
+   @Nullable
+   public abstract List<VirtualMachineStatus> statuses();
+
+
+   @SerializedNames({"platformUpdateDomain", "platformFaultDomain", "statuses"})
+   public static VirtualMachineInstance create(final String platformUpdateDomain, final String platformFaultDomain,
+                                               final List<VirtualMachineStatus> statuses) {
+
+      return new AutoValue_VirtualMachineInstance(platformUpdateDomain, platformFaultDomain, statuses);
```
nacx: Initialize to immutable if present.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/features/VirtualMachineApi.java
```diff
@@ -0,0 +1,140 @@
...
+
+/**
+ * The Virtual Machine API includes operations for managing the virtual machines in your subscription.
+ *
+ * @see <a href="https://msdn.microsoft.com/en-us/library/azure/mt163630.aspx">docs</a>
+ */
+@Path("/resourceGroups/{resourceGroup}/providers/Microsoft.Compute/virtualMachines")
+@RequestFilters(OAuthFilter.class)
+@QueryParams(keys = "api-version", values = "2015-06-15")
+@Produces(MediaType.APPLICATION_JSON)
```
nacx: Move this annotation at method level just for the methods that actually produce a body.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/features/VirtualMachineApi.java
```diff
@@ -0,0 +1,140 @@
...
+   @Fallback(Fallbacks.NullOnNotFoundOr404.class)
+   URI delete(@PathParam("name") String name);
+
+   /**
+    * The Restart Virtual Machine operation
+    */
+   @Named("RestartVirtualMachine")
+   @POST
+   @Path("/{name}/restart")
+   @Fallback(Fallbacks.VoidOnNotFoundOr404.class)
```
nacx: We don't want to silently succeed if the VM does not exist. Remove the 404 fallbacks in all state change operations.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/NetworkInterfaceCardApiLiveTest.java
```diff
@@ -0,0 +1,135 @@
...
+      final NetworkInterfaceCardApi nicApi = api.getNetworkInterfaceCardApi(resourcegroup);
+
+
+      //Create properties object
+      final NetworkInterfaceCardProperties networkInterfaceCardProperties =
+              NetworkInterfaceCardProperties.create(null, null, null,
+                      Arrays.asList(IpConfiguration.create("myipconfig", null, null, null,
+                              IpConfigurationProperties.create(null, null, "Dynamic", IdReference.create(subnetID), null))
+                      )
+              );
```
nacx: Good candidates to use builders?

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/NetworkInterfaceCardApiLiveTest.java
```diff
@@ -0,0 +1,135 @@
...
+
+      final NetworkInterfaceCardApi nicApi = api.getNetworkInterfaceCardApi(resourcegroup);
+
+      List<NetworkInterfaceCard> nicList = nicApi.list();
+
+      assertTrue(nicList.size() > 0);
+   }
+
+
+   @Test(groups = "live", dependsOnMethods = "listNetworkInterfaceCards", alwaysRun = true)
```
nacx: This must depend on the "get" test too.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/NetworkInterfaceCardApiLiveTest.java
```diff
@@ -0,0 +1,135 @@
...
+   }
+
+   @Test(groups = "live", dependsOnMethods = "createNetworkInterfaceCard")
+   public void listNetworkInterfaceCards() {
+
+      final NetworkInterfaceCardApi nicApi = api.getNetworkInterfaceCardApi(resourcegroup);
+
+      List<NetworkInterfaceCard> nicList = nicApi.list();
+
+      assertTrue(nicList.size() > 0);
```
nacx: Assert instead that it contains the created nic.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/NetworkInterfaceCardApiMockTest.java
```diff
@@ -0,0 +1,127 @@
...
+      List<NetworkInterfaceCard> nicList = nicApi.list();
+      String path = String.format("/subscriptions/%s/resourcegroups/%s/providers/Microsoft.Network/networkInterfaces?%s", subscriptionid, resourcegroup, apiVersion);
+
+      assertSent(server, "GET", path);
+      assertTrue(nicList.size() == 2);
+      assertTrue(nicList.get(0).properties().ipConfigurations().size() > 0);
+      assertEquals(nicList.get(0).properties().ipConfigurations().get(0).properties().privateIPAllocationMethod(), "Dynamic");
+      assertTrue(nicList.get(1).properties().ipConfigurations().size() > 0);
+      assertEquals(nicList.get(1).properties().ipConfigurations().get(0).properties().privateIPAllocationMethod(), "Static");
+   }
```
nacx: Add tests for the list and get fallbacks.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/PublicIPAddressApiLiveTest.java
```diff
@@ -0,0 +1,127 @@
...
+
+   @Test(groups = "live", dependsOnMethods = "deletePublicIPAddressResourceDoesNotExist")
+   public void createPublicIPAddress() {
+
+      final PublicIPAddressApi ipApi = api.getPublicIPAddressApi(resourcegroup);
+
+      final Map<String, String> tags = ImmutableMap.of("testkey", "testvalue");
+
+      PublicIPAddressProperties properties =
+              PublicIPAddressProperties.create(null, null, "Static", 4, null, null);
```
nacx: Another good candidate to use builders.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/PublicIPAddressApiLiveTest.java
```diff
@@ -0,0 +1,127 @@
...
+
+      //Poll until resource is ready to be used
+      while (ip.properties().provisioningState().equals("Updating")) {
+         try {
+            Thread.sleep(10 * 1000);
+         } catch (InterruptedException e) {
+            e.printStackTrace();
+         }
+         ip = ipApi.get(publicIpAddressName);
+      }
```
nacx: Use the jclouds `predicates2.retry` construct. If this is a common thing to be done, consider creating the predicate and make it injectable so other classes can benefit from it.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/PublicIPAddressApiLiveTest.java
```diff
@@ -0,0 +1,127 @@
...
+   public void listPublicIPAddresses() {
+
+      final PublicIPAddressApi ipApi = api.getPublicIPAddressApi(resourcegroup);
+
+      List<PublicIPAddress> ipList = ipApi.list();
+
+      assertTrue(ipList.size() > 0);
+   }
+
+   @Test(groups = "live", dependsOnMethods = "listPublicIPAddresses", alwaysRun = true)
```
nacx: Must depend on the get test too.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/PublicIPAddressApiMockTest.java
```diff
@@ -0,0 +1,155 @@
...
+   public void listPublicIPAddresses() throws InterruptedException {
+      server.enqueue(jsonResponse("/PublicIPAddressList.json"));
+
+      final PublicIPAddressApi ipApi = api.getPublicIPAddressApi(resourcegroup);
+      List<PublicIPAddress> ipList = ipApi.list();
+
+      String path = String.format("/subscriptions/%s/resourcegroups/%s/providers/Microsoft.Network/publicIPAddresses?%s", subscriptionid, resourcegroup, apiVersion);
+      assertSent(server, "GET", path);
+      assertEquals(ipList.size(), 4);
+   }
```
nacx: Add tests for the list and get fallbacks.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/VirtualMachineApiLiveTest.java
```diff
@@ -0,0 +1,269 @@
...
+   }
+
+   @Test
+   public void testCreate() {
+      StorageAccountApi storageApi = api.getStorageAccountApi(getResourceGroupName());
+      StorageService storageAccount = storageApi.get(getStorageServiceName());
+      String blob = storageAccount.storageServiceProperties().primaryEndpoints().get("blob");
+
+      String id = "/subscriptions/" + subscriptionid + "/resourceGroups/" + getResourceGroupName() +
+              "/providers/" + "Microsoft.Compute/virtualMachines/" + getName();
```
nacx: Will users have to manually build this string every time a VM has to be created? It doesn't look very usable. I think it could be a way to get rid of this ID parameter in the method.

First, change the payload defined in the interface to something like the following, to configure the ID template, and remove the ID parameter from the method signature:

```java
@Payload("%7B\"id\":\"/subscriptions/{subscriptionId}/resourceGroups/{resourceGroup}" 
    + "/providers/Microsoft.Compute/virtualMachines/{name}\","
    + "\"name\":\"{name}\",\"type\":\"Microsoft.Compute/virtualMachines\","
    + "\"location\":\"{location}\",\"tags\":%7B%7D,\"properties\":{properties}%7D")
```

Now we just need to resolve the `subscriptionId` and `resourceGroup` parameters and put them on the payload. The resourceGroup has been already passed to the `api.getVirtualMachineApi` method so we should have it, and the subscriptionId is part of the endpoint. We could automatically get them as follows:

In the `AzureComputeHttpApiModule`, add a method that provides the subscription id, as follows:

```java
@Provides
@Named("subscription-id")
protected String provideSubscriptionId(ProviderMetadata providerMetadata) {
  // Return the subscription id extracted from the providerMetadata.getEndpoint()
}
```

Then create a custom binder that extends the `BindToJsonPayload` like

```java
@Singleton
public class BindSubscriptionIdAndResourceName extends BindToJsonPayload {
   private final String subscriptionId;

   @Inject public BindSubscriptionIdAndResourceName(Json jsonBinder,
         @Named("subscription-id") String subscriptionId) {
      super(jsonBinder);
      this.subscriptionId = subscriptionId;
   }
   
   @Override public <R extends HttpRequest> R bindToRequest(R request, Map<String, Object> postParams) {
      checkArgument(request instanceof GeneratedHttpRequest, "this binder only supports GeneratedHttpRequests");
      checkArgument(GeneratedHttpRequest.class.cast(request).getCaller().isPresent(), "this binder is only for delegate apis");

      // Get the value for the resource group name from the caller API
      Invocation caller = GeneratedHttpRequest.class.cast(request).getCaller().get();
      String resourceGroupName = (String) caller.getArgs().get(0);

      // Populate the computed values to the payload param map so they are bound to the payload
      postParams.put("resourceGroup", resourceGroupName);
      postParams.put("subscriptionId", subscriptionId);

      return super.bindToRequest(request, postParams);
   }
}
```

Finally, change the `@MapBinder` annotation in the method to use this new binder class.

I haven't tested it, but it should work or at least give a hint on how we can infer the subscription id and resource name in the method so users don't have to generate that ID.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/VirtualMachineApiLiveTest.java
```diff
@@ -0,0 +1,269 @@
...
+            Thread.sleep(120 * 1000); // Creation takes long time. Recheck later
+         } catch (InterruptedException e) {
+            e.printStackTrace();
+         }
+         vm = api().get(getName());
+         status = vm.properties().provisioningState();
+         stopCounter++;
+         if (stopCounter == 15)
+            break;
+      }
```
nacx: Same as before, use the jclouds `Predictes2.retry` construct, and if it makes sense create a predicate and make it injectable for later reuse. The `AzureComputeHttpApiModule` is a good place to declare such reusable predicates.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/VirtualMachineApiLiveTest.java
```diff
@@ -0,0 +1,269 @@
...
+            try {
+               Thread.sleep(5 * 1000); // Starting takes long time. Recheck later
+            } catch (InterruptedException e) {
+               e.printStackTrace();
+            }
+            stopCounter++;
+            if (stopCounter == 40)
+               break;
+         }
+      }
```
nacx: Same here about reusable predicates.

nacx: Also, instead of using a counter, let's better use a timeout and take benefit of the `Predicates2.retry`.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/VirtualMachineApiLiveTest.java
```diff
@@ -0,0 +1,269 @@
...
+      assertEquals(status, "VM running");
+   }
+
+   @Test(dependsOnMethods = "testCreate")
+   public void testList() {
+      List<VirtualMachine> list = api().list();
+      for (VirtualMachine machine : list) {
```
nacx: I'd just assert that it contains the created VM.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/VirtualMachineApiLiveTest.java
```diff
@@ -0,0 +1,269 @@
...
+         assertTrue(!machine.name().isEmpty());
+      }
+   }
+
+   @Test(dependsOnMethods = "testRestart")
```
nacx: Also `alwaysRun = true`? This will need to depend on other methods too, owhterwise this could be run *before* the list or get tests.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/VirtualMachineApiMockTest.java
```diff
@@ -0,0 +1,161 @@
...
+@Test(groups = "unit", testName = "VirtualMachineApiMockTest", singleThreaded = true)
+public class VirtualMachineApiMockTest extends BaseAzureComputeApiMockTest {
+
+   public void testGet() throws Exception {
+      server.enqueue(jsonResponse("/virtualmachine.json"));
+      final VirtualMachineApi vmAPI = api.getVirtualMachineApi("groupname");
+      assertEquals(vmAPI.get("windowsmachine"), getVM());
+      assertSent(server, "GET", "/subscriptions/SUBSCRIPTIONID/resourceGroups/groupname/providers/Microsoft.Compute" +
+              "/virtualMachines/windowsmachine?api-version=2015-06-15");
+   }
```
nacx: Add the 404 fallback tests and the getInstanceDetails tests

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/VirtualMachineApiMockTest.java
```diff
@@ -0,0 +1,161 @@
...
+
+   public void testList() throws Exception {
+      server.enqueue(jsonResponse("/virtualmachines.json"));
+      final VirtualMachineApi vmAPI = api.getVirtualMachineApi("groupname");
+      assertEquals(vmAPI.list(), getVMList());
+      assertSent(server, "GET", "/subscriptions/SUBSCRIPTIONID/resourceGroups/groupname/providers/Microsoft.Compute" +
+              "/virtualMachines?api-version=2015-06-15");
+   }
```
nacx: Add the list fallback tests too.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/VirtualMachineApiMockTest.java
```diff
@@ -0,0 +1,161 @@
...
+
+   public void testCreate() throws Exception {
+      server.enqueue(jsonResponse("/createvirtualmachineresponse.json"));
+      String id = "/subscriptions/SUBSCRIPTIONID/" +
+              "resourceGroups/groupname/providers/Microsoft.Compute/virtualMachines/windowsmachine";
+      final VirtualMachineApi vmAPI = api.getVirtualMachineApi("groupname");
+      VirtualMachine vm = vmAPI.create("windowsmachine", id, "windowsmachine", "westus", getProperties());
+      assertEquals(vm, getVM());
+      assertSent(server, "PUT", "/subscriptions/SUBSCRIPTIONID/resourceGroups/groupname/providers/Microsoft.Compute" +
+              "/virtualMachines/windowsmachine?api-version=2015-06-15&validating=false");
```
nacx: The generated body needs to be verified too

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/VirtualMachineApiMockTest.java
```diff
@@ -0,0 +1,161 @@
...
+      final VirtualMachineApi vmAPI = api.getVirtualMachineApi("groupname");
+
+      URI uri = vmAPI.delete("windowsmachine");
+
+      assertEquals(server.getRequestCount(), 1);
+      assertNull(uri);
+
+      assertSent(server, "DELETE", "/subscriptions/SUBSCRIPTIONID/resourceGroups/groupname/providers/Microsoft.Compute" +
+              "/virtualMachines/windowsmachine?api-version=2015-06-15");
+   }
```
nacx: Add a test for the delete method that does not involve the fallback.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/internal/BaseAzureComputeApiLiveTest.java
```diff
@@ -74,6 +83,17 @@ protected String getEndpoint() {
...
+   protected String getSubscriptionId() {
+      String subscriptionid = null;
+      String endpoint = null;
+      endpoint = getEndpoint();
+      if (endpoint != null) {
+         subscriptionid = endpoint.substring(endpoint.lastIndexOf("/") + 1);
+      }
+      assertNotNull(subscriptionid);
+      return subscriptionid;
+   }
```
nacx: Once the subscription id is bound to the Guice context, change this to get it from the injector.

nacx:
Thanks @ritazh!

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/NetworkInterfaceCardApiMockTest.java
ritazh: Done.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/NetworkInterfaceCardProperties.java
ritazh: Done.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/PublicIPAddressApiMockTest.java
ritazh: Done.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/VirtualMachineApiMockTest.java
ritazh: Done.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/VirtualMachineApiMockTest.java
ritazh: Done.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/VirtualMachineApiMockTest.java
ritazh: Done.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/internal/BaseAzureComputeApiLiveTest.java
ritazh: This is actually not needed, since we have updated `VirtualMachineApi.create()`'s payload as `@Payload("%7B\"location\":\"{location}\",\"tags\":%7B%7D,\"properties\":{properties}%7D")`. Hence we are still using `BindToJsonPayload`.

ritazh:
@nacx Ready for another round of review.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/config/AzureComputeHttpApiModule.java
```diff
@@ -54,4 +57,15 @@ protected void configure() {
       super.configure();
       bind(OAuthScopes.class).toInstance(OAuthScopes.NoScopes.create());
    }
+   @Provides
+   @Named("subscriptionId")
+   protected String provideSubscriptionId(ProviderMetadata providerMetadata) {
+      String subscriptionid = null;
+      String endpoint = null;
+      endpoint = providerMetadata.getEndpoint();
+      if (endpoint != null) {
```
nacx: The endpoint can't be null (jclouds enforces that when building the context), so you could remove this conditional and the endpoint variable.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/DiagnosticsProfile.java
```diff
@@ -0,0 +1,68 @@
...
+import org.jclouds.javax.annotation.Nullable;
+import org.jclouds.json.SerializedNames;
+
+@AutoValue
+public abstract class DiagnosticsProfile {
+
+   @AutoValue
+   public abstract static class BootDiagnostics{
+
+      public abstract Boolean enabled();
```
nacx: Make primitive or mark as `@Nullable`.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/NetworkInterfaceCardProperties.java
nacx: This pattern still needs to be applied? That is, providing those hidden methods so we can change the build method to enforce an immutable list. Otherwise the we can't guarantee collection immutability. (Apply this to other builders that accept collection parameters).

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/OSProfile.java
nacx: Still TBD?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineInstance.java
```diff
@@ -0,0 +1,71 @@
...
+   }
+
+   @Nullable
+   public abstract String platformUpdateDomain();
+
+   @Nullable
+   public abstract String platformFaultDomain();
+
+   @Nullable
+   public abstract ImmutableList<VirtualMachineStatus> statuses();
```
nacx: Change type to `List`.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/features/NetworkInterfaceCardApi.java
```diff
@@ -0,0 +1,78 @@
...
+   @Named("networkinterfacecard:list")
+   @SelectJson("value")
+   @GET
+   @Fallback(EmptyListOnNotFoundOr404.class)
+   List<NetworkInterfaceCard> list();
+
+   @Named("networkinterfacecard:create_or_update")
+   @Path("/{networkinterfacecardname}")
+   @MapBinder(BindToJsonPayload.class)
+   @PUT
```
nacx: Add the produces annotation here at method level

nacx: Dismiss this comment. The binder already sets the content-type to application/json, so there is no need for the produces annotation here.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/VirtualMachineApiLiveTest.java
```diff
@@ -0,0 +1,241 @@
...
+   public void testStop() {
+      api().stop(getName());
+      //Poll until resource is ready to be used
+      boolean jobDone = Predicates2.retry(new Predicate<String>() {
+         @Override public boolean apply(String name) {
+            String status = "";
+            List<VirtualMachineInstance.VirtualMachineStatus> statuses = api().getInstanceDetails(name).statuses();
+            for (int c = 0; c < statuses.size(); c++) {
+               if (statuses.get(c).code().substring(0, 10).equals("PowerState")) {
+                  status = statuses.get(c).displayStatus();
```
nacx: `break` here?

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/VirtualMachineApiMockTest.java
```diff
@@ -0,0 +1,241 @@
...
+      VirtualMachineInstance actual = vmAPI.getInstanceDetails("windowsmachine");
+      VirtualMachineInstance expected = getVMInstance();
+
+      assertEquals(actual.statuses().get(0).code(), expected.statuses().get(0).code());
+      assertEquals(actual.statuses().get(0).displayStatus(), expected.statuses().get(0).displayStatus());
+      assertEquals(actual.statuses().get(0).level(), expected.statuses().get(0).level());
+      assertEquals(actual.statuses().get(0).time().toString(), expected.statuses().get(0).time().toString());
+      assertSent(server, "GET", "/subscriptions/SUBSCRIPTIONID/resourceGroups/groupname/providers/Microsoft.Compute" +
+              "/virtualMachines/windowsmachine/instanceView?api-version=2015-06-15");
+   }
```
nacx: Add the fallback test for this one.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/internal/BaseAzureComputeApiLiveTest.java
nacx: If this method is no longer needed, then remove it. If it is needed, then change it to get the subscriptionId from the injector, now that it is bound to the context:
```java
return injector.getInstance(Key.get(String.class, Names.named("subscriptionId")));
```
You can initialize the injector into a protected variable in the base test class. You can look at the [BaseDigitalOcean2ApiLiveTest](https://github.com/jclouds/jclouds/blob/master/providers/digitalocean2/src/test/java/org/jclouds/digitalocean2/internal/BaseDigitalOcean2ApiLiveTest.java) for an example.

nacx:
Just some final minor comments @ritazh! Thanks! Look forward to having this merged!

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/internal/BaseAzureComputeApiLiveTest.java
ritazh: @nacx `getSubscriptionId` is needed but only for two live tests, not in the actual provider code. We don't actually need `subscriptionId` to be injected anywhere in the provider code. Note the above payload no longer needs it. Do we have inject this value?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/NetworkInterfaceCardProperties.java
ritazh: Sorry I missed this. I have added the pattern to all builder methods that receive collection parameters. Thanks!

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/OSProfile.java
ritazh: Done.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/VirtualMachineApiLiveTest.java
ritazh: Done. Added `break` to all the for loops in this test.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/VirtualMachineApiMockTest.java
ritazh: Done.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/config/AzureComputeHttpApiModule.java
ritazh: Removed this `@provides` since we no longer need it in this provider code.

ritazh:
@nacx Thanks for the feedback. Ready for another round of review.

nacx:
And [merged](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/3be36289)! Thanks @ritazh!

PR #270, opened 2016-05-20
JCLOUDS-1005: Backblaze B2 skeleton and bucket ops

andrewgaul:


andrewgaul:
@nacx @zack-shoylev Could you review the skeleton for me?  B2 is the first provider I have written and I want to make sure it follows best practices.

b2/src/main/java/org/jclouds/b2/B2ApiMetadata.java
```diff
@@ -0,0 +1,85 @@
...
+import org.jclouds.blobstore.BlobStoreContext;
+import org.jclouds.labs.b2.config.B2HttpApiModule;
+import org.jclouds.rest.internal.BaseHttpApiMetadata;
+
+import com.google.auto.service.AutoService;
+import com.google.common.collect.ImmutableSet;
+import com.google.inject.Module;
+
+@AutoService(ApiMetadata.class)
+// TODO: B2ProviderMetadata?
```
andrewgaul: Is the difference between API and provider meaningful?

b2/src/main/java/org/jclouds/b2/filters/RequestAuthorization.java
```diff
@@ -0,0 +1,91 @@
...
+
+      // Replace fixed URL with API URL, except for unit tests tests
+      URI endpoint = request.getEndpoint();
+      if (!endpoint.getHost().equals("localhost")) {
+         endpoint = URI.create(auth.apiUrl() + endpoint.getPath() + (endpoint.getQuery() == null ? "" : "?" + endpoint.getQuery()));
+      }
+
+      request = request.toBuilder()
+            .endpoint(endpoint)
+            // TODO: squirrel away accountId in bogus header so ReplacePayloadAccountId can use it
```
andrewgaul: B2 requires the accountId in the JSON payload of requests but we don't want to include it in API calls.  Can I more cleanly pass this information to `ReplacePayloadAccountId`?

b2/src/main/java/org/jclouds/b2/features/BucketApi.java
```diff
@@ -0,0 +1,87 @@
...
+   @Path("/b2api/v1/b2_list_buckets")
+   @Payload("{\"accountId\":\"{accountId}\"}")
+   @RequestFilters(ReplacePayloadAccountId.class)
+   @Consumes(APPLICATION_JSON)
+   BucketList listBuckets();
+
+   static final class ReplacePayloadAccountId implements HttpRequestFilter {
+      @Override
+      public HttpRequest filter(HttpRequest request) throws HttpException {
+         String accountId = request.getFirstHeaderOrNull("accountId");  // TODO: hack
```
andrewgaul: See below.

b2/src/test/java/org/jclouds/b2/features/BucketApiMockTest.java
```diff
@@ -0,0 +1,239 @@
...
+   }
+
+   /**
+    * Ensures the request has a json header.
+    *
+    * @param request
+    * @see RecordedRequest
+    */
+   private void assertContentTypeIsJSON(RecordedRequest request) {
+      // TODO: where to add this?  something is stripping it off
```
andrewgaul: Something strips off my Content-Type -- any idea how to force this?

andrewgaul:
Failures due to jclouds/jclouds#959.

b2/src/main/java/org/jclouds/b2/B2ApiMetadata.java
nacx: Yes. The api metadata should just describe the API, and the provider metadata should add to the API the real endpoint (apis can provide a default one), and the default configuration. For providers that have their own API this makes less sense, but I think it is still good to separate both. HAve a look at the digitalocean2 provider for an example.

You may also want to use the same provider id for the api and provider, so just remove this annotation once the provider metadata exists, to avoid having conflicts.

b2/pom.xml
```diff
@@ -0,0 +1,162 @@
...
+  <!-- TODO: when out of labs, switch to org.jclouds.api -->
+  <groupId>org.apache.jclouds.labs</groupId>
+  <artifactId>b2</artifactId>
+  <name>Apache jclouds B2 API</name>
+  <description>BlobStore binding to the Backblaze B2 API</description>
+  <packaging>bundle</packaging>
+
+  <properties>
+    <test.aws.identity>FIXME_IDENTITY</test.aws.identity>
+    <test.aws.credential>FIXME_CREDENTIAL</test.aws.credential>
```
nacx: aws? :)

b2/pom.xml
```diff
@@ -0,0 +1,162 @@
...
+    <test.b2.build-version />
+    <test.b2.identity>${test.aws.identity}</test.b2.identity>
+    <test.b2.credential>${test.aws.credential}</test.b2.credential>
```
nacx: Same here

b2/src/main/java/org/jclouds/b2/B2Client.java
```diff
@@ -0,0 +1,32 @@
...
+package org.jclouds.labs.b2;
+
+import java.io.Closeable;
+
+import org.jclouds.labs.b2.features.AuthorizationApi;
+import org.jclouds.labs.b2.features.BucketApi;
+import org.jclouds.rest.annotations.Delegate;
+
+/** Provides access to Backblaze B2 resources via their REST API. */
+public interface B2Client extends Closeable {
```
nacx: Convention is to name B2Api

b2/src/main/java/org/jclouds/b2/features/AuthorizationApi.java
```diff
@@ -0,0 +1,54 @@
...
+
+import com.google.common.base.Charsets;
+import com.google.common.io.BaseEncoding;
+
+public interface AuthorizationApi {
+   @Named("b2_authorize_account")
+   @GET
+   @Path("/b2api/v1/b2_authorize_account")
+   @Headers(keys = "Authorization", values = "{accountId}:{applicationKey}")
+   @RequestFilters(Base64EncodeAuthorization.class)
```
nacx: Remove the `@Headers` annotation and use the existing filter: `@RequestFilter(BasicAuthentication.class)`.
Also, since the accountId is the "identity", you could create a class that extends the jclouds `MapBinder` (the one that binds the values tot he request the request) and has the Credentials injected. The mapbinder could then update the path with the identity, and you remove the need for the first "accountId" parameter.

b2/src/main/java/org/jclouds/b2/features/AuthorizationApi.java
```diff
@@ -0,0 +1,54 @@
...
+   @Consumes(APPLICATION_JSON)
+   Authorization authorizeAccount(@PathParam("accountId") String accountId, @PathParam("applicationKey") String applicationKey);
+
+   static final class Base64EncodeAuthorization implements HttpRequestFilter {
```
nacx: This class is not needed; jclouds already provides it.

b2/src/main/java/org/jclouds/b2/filters/RequestAuthorization.java
nacx: With a minor modification in jclouds core: yes.

Change the `RestAnnotationProcessor` to have the Credentials injected. Then, add the following line after [these ones](https://github.com/jclouds/jclouds/blob/master/core/src/main/java/org/jclouds/rest/internal/RestAnnotationProcessor.java#L229-L230):
```java
tokenValues.put(Constants.PROPERTY_IDENTITY, credentials.identity);
```

With this change in place, you will be able to declare the methods like:
```java
@Named("b2_list_buckets")
@POST
@Path("/b2api/v1/b2_list_buckets")
@PayloadParams(keys = {"accountId"}, values = {"{jclouds.identity}"})
@Consumes(APPLICATION_JSON)
BucketList listBuckets();
```
Removing the need for the filters and hacks, as jclouds will already substitute the identity in the parameter value.

nacx:
@andrewgaul I've commented about some structural things. Hope the hints help :) I'll go through the entire PR during the weekend!

b2/src/main/java/org/jclouds/b2/B2Client.java
andrewgaul: Done.

b2/pom.xml
andrewgaul: Done.

b2/pom.xml
andrewgaul: Done.

b2/src/main/java/org/jclouds/b2/features/AuthorizationApi.java
andrewgaul: Done.

b2/src/main/java/org/jclouds/b2/B2ApiMetadata.java
andrewgaul: Done.

b2/src/main/java/org/jclouds/b2/filters/RequestAuthorization.java
andrewgaul: This suggestion will take me a little while to investigate.  I will push an amended commit addressing all other comments.

andrewgaul: Done.

b2/src/test/java/org/jclouds/b2/features/BucketApiMockTest.java
andrewgaul: Done.

b2/src/main/java/org/jclouds/b2/features/BucketApi.java
andrewgaul: Done.

b2/src/main/java/org/jclouds/b2/features/AuthorizationApi.java
andrewgaul: Done.

andrewgaul:
I addressed all outstanding review comments.  Tests will succeed after the next Maven snapshot update.

b2/pom.xml
```diff
@@ -0,0 +1,160 @@
...
+      <id>apache-snapshots</id>
+      <url>https://repository.apache.org/content/repositories/snapshots</url>
+      <releases>
+        <enabled>false</enabled>
+      </releases>
+      <snapshots>
+        <enabled>true</enabled>
+      </snapshots>
+    </repository>
+  </repositories>
```
nacx: This is already defined in the parent pom

b2/src/main/java/org/jclouds/b2/B2ApiMetadata.java
```diff
@@ -0,0 +1,83 @@
...
+   }
+
+   public static class Builder extends BaseHttpApiMetadata.Builder<B2Api, Builder> {
+
+      protected Builder() {
+         super(B2Api.class);
+         id("b2")
+                 .name("Backblaze B2 API")
+                 .identityName("Access Key ID")
+                 .credentialName("Secret Access Key")
```
nacx: In their naming, these are the "Account Id" and "Application Key"

b2/src/main/java/org/jclouds/b2/B2ResponseException.java
```diff
@@ -0,0 +1,41 @@
...
+
+public class B2ResponseException extends HttpResponseException {
+
+   private static final long serialVersionUID = 1L;
+   private final B2Error error;
+
+   public B2ResponseException(HttpCommand command, HttpResponse response, B2Error error) {
+      super("request " + command.getCurrentRequest().getRequestLine() + " failed with code " + response.getStatusCode()
+            + ", error: " + error.toString(), command, response);
+      this.error = checkNotNull(error, "error");
```
nacx: This check should be performed *before* calling its toString() method.

b2/src/main/java/org/jclouds/b2/domain/BucketList.java
```diff
@@ -0,0 +1,33 @@
...
+
+import com.google.auto.value.AutoValue;
+
+@AutoValue
+public abstract class BucketList {
+   public abstract List<Bucket> buckets();
+
+   @SerializedNames({"buckets"})
+   public static BucketList create(List<Bucket> buckets) {
+      return new AutoValue_BucketList(buckets);
```
nacx: Enforce an immutable list by using copyOf

b2/src/main/java/org/jclouds/b2/features/BucketApi.java
```diff
@@ -0,0 +1,78 @@
...
+   @Produces(APPLICATION_JSON)
+   Bucket createBucket(@PayloadParam("bucketName") String bucketName, @PayloadParam("bucketType") BucketType bucketType);
+
+   @Named("b2_delete_bucket")
+   @POST
+   @Path("/b2api/v1/b2_delete_bucket")
+   @MapBinder(BindToJsonPayload.class)
+   @PayloadParams(keys = {"accountId"}, values = {"{jclouds.identity}"})
+   @Consumes(APPLICATION_JSON)
+   @Produces(APPLICATION_JSON)
```
nacx: Worth adding a `@Fallback(NullOnNotFoundOr404.class)`? It is a pretty common practice in jclouds to avoid failing if the resource one is trying to delete is already gone.

nacx: Hmm having a deeper look seems that this api never returns a 404?

b2/src/main/java/org/jclouds/b2/filters/RequestAuthorization.java
```diff
@@ -0,0 +1,89 @@
...
+import com.google.common.base.Function;
+import com.google.common.base.Supplier;
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
+
+@Singleton
+public class RequestAuthorization implements HttpRequestFilter {
+   @Resource
+   @Named(Constants.LOGGER_SIGNATURE)
```
nacx: I'm not sure about this logger, s this is not signing anything. Create a custom logger for b2?

b2/src/main/java/org/jclouds/b2/filters/RequestAuthorization.java
```diff
@@ -0,0 +1,89 @@
...
+   Logger signatureLog = Logger.NULL;
+
+   private final Supplier<Credentials> creds;
+   private final HttpUtils utils;
+   private final LoadingCache<Credentials, Authorization> cache;
+
+   @Inject
+   public RequestAuthorization(@Provider Supplier<Credentials> creds, final B2Api b2Api, HttpUtils utils) {
```
nacx: As per Guice best practices, make this package-private by removing the modifier.

b2/src/main/java/org/jclouds/b2/filters/RequestAuthorization.java
```diff
@@ -0,0 +1,89 @@
...
+      this.creds = checkNotNull(creds, "creds");
+      this.utils = checkNotNull(utils, "utils");
```
nacx: Checks are redundant. The injector already enforces non-null values.

b2/src/main/java/org/jclouds/b2/handlers/ParseB2ErrorFromJsonContent.java
```diff
@@ -0,0 +1,51 @@
...
+import org.jclouds.labs.b2.B2ResponseException;
+import org.jclouds.labs.b2.domain.B2Error;
+
+import com.google.inject.Inject;
+import com.google.inject.TypeLiteral;
+
+public class ParseB2ErrorFromJsonContent extends ParseJson<B2Error> implements HttpErrorHandler {
+
+   @Inject
+   public ParseB2ErrorFromJsonContent(Json json) {
```
nacx: Make this package private

b2/src/main/java/org/jclouds/b2/handlers/ParseB2ErrorFromJsonContent.java
```diff
@@ -0,0 +1,51 @@
...
+      super(json, TypeLiteral.get(B2Error.class));
+   }
+
+   private static Exception refineException(B2Error error, Exception exception) {
+      if ("bad_json".equals(error.code())) {
+         return new IllegalArgumentException(error.message(), exception);
+      } else {
```
nacx: Also try to refine 404 errors to a ResourceNotFoundException to let the 404 fallbacks work

b2/src/test/java/org/jclouds/b2/features/BucketApiMockTest.java
```diff
@@ -0,0 +1,236 @@
...
+   }
+
+   @SuppressWarnings("serial")
+   public B2Api api(String uri, String provider, Properties overrides) {
+      return ContextBuilder.newBuilder(provider)
+            .credentials("ACCOUNT_ID", "APPLICATION_KEY")
+            .endpoint(uri)
+            .overrides(overrides)
+            .modules(modules)
+            .buildApi(new TypeToken<B2Api>(getClass()) {});
```
nacx: Just use `.buildApi(B2Api.class);`?

b2/src/test/java/org/jclouds/b2/internal/BaseB2ApiLiveTest.java
```diff
@@ -0,0 +1,45 @@
...
+   @Override
+   protected ApiMetadata createApiMetadata() {
+      return new B2ApiMetadata();
+   }
+
+   @Override
+   protected B2Api create(Properties props, Iterable<Module> modules) {
+      Injector injector = newBuilder().modules(modules).overrides(props).buildInjector();
+      return injector.getInstance(B2Api.class);
+   }
```
nacx: Is there a real need to override this method? It looks like it does the same than the parent?

b2/src/main/java/org/jclouds/b2/filters/RequestAuthorization.java
```diff
@@ -0,0 +1,89 @@
...
+
+   @Override
+   public HttpRequest filter(HttpRequest request) throws HttpException {
+      utils.logRequest(signatureLog, request, ">>");
+      Credentials creds = this.creds.get();
+      Authorization auth = cache.getUnchecked(creds);
+
+      // Replace fixed URL with API URL, except for unit tests tests
+      URI endpoint = request.getEndpoint();
+      if (!endpoint.getHost().equals("localhost")) {
```
nacx: Can we get rid of this, which couples the code to the tests? Can we configure the tests, and the json returned by the authorization requests to return an api url that can make the tests work without this conditional?

b2/pom.xml
andrewgaul: Done.

b2/src/main/java/org/jclouds/b2/B2ApiMetadata.java
andrewgaul: Done.

b2/src/main/java/org/jclouds/b2/B2ResponseException.java
andrewgaul: Done.

b2/src/main/java/org/jclouds/b2/domain/BucketList.java
andrewgaul: Done.

b2/src/main/java/org/jclouds/b2/filters/RequestAuthorization.java
andrewgaul: Done.

b2/src/main/java/org/jclouds/b2/filters/RequestAuthorization.java
andrewgaul: Done.

b2/src/main/java/org/jclouds/b2/handlers/ParseB2ErrorFromJsonContent.java
andrewgaul: Done.

b2/src/test/java/org/jclouds/b2/features/BucketApiMockTest.java
andrewgaul: Done.

b2/src/test/java/org/jclouds/b2/internal/BaseB2ApiLiveTest.java
andrewgaul: Done.

b2/src/main/java/org/jclouds/b2/filters/RequestAuthorization.java
andrewgaul: I changed this but the alternative might not be better; we need to munge the port to get MockWebServer to work correctly.

b2/src/main/java/org/jclouds/b2/handlers/ParseB2ErrorFromJsonContent.java
andrewgaul: Let's address this in the upcoming object pull request which will encounter 404 errors.

b2/src/main/java/org/jclouds/b2/features/BucketApi.java
andrewgaul: I added `testDeleteAlreadyDeletedBucket` to test this.  B2 return 400 with a `bad_bucket_id`.  I modified `ParseB2ErrorFromJsonContent` to turn this into `ContainerNotFoundException`.

b2/src/main/java/org/jclouds/b2/filters/RequestAuthorization.java
andrewgaul: I agree that B2 authorization is not strictly signing but as is it goes where all the other authentication logs go.

b2/src/main/java/org/jclouds/b2/filters/RequestAuthorization.java
andrewgaul: Re-reading this, I agree with you and will remove it.

b2/src/main/java/org/jclouds/b2/features/BucketApi.java
```diff
@@ -0,0 +1,78 @@
...
+@RequestFilters(RequestAuthorization.class)
+@BlobScope(CONTAINER)
+public interface BucketApi {
+   @Named("b2_create_bucket")
+   @POST
+   @Path("/b2api/v1/b2_create_bucket")
+   @MapBinder(BindToJsonPayload.class)
+   @PayloadParams(keys = {"accountId"}, values = {"{jclouds.identity}"})
+   @Consumes(APPLICATION_JSON)
+   @Produces(APPLICATION_JSON)
```
nacx: [minor] Since all methods produce and consume the same, you can move this annotations to class level and remove them from each method.

b2/src/main/java/org/jclouds/b2/features/BucketApi.java
nacx: Now that it turns it into that exception, adding the `@Fallback(NullOnNotFoundOr404.class)` would already work (as it looks for 404 errors or exceptions that are instanceof ResoueceNotFoundException).
With that fallback method would return just `null` if the object didn't already exist instead of throwing an exception, which is a pretty common pattern in most of the jclouds APIs. Worth adding the fallback?

nacx:
Just a couple final minor comments. LGTM!

b2/src/main/java/org/jclouds/b2/features/BucketApi.java
andrewgaul: Done.

b2/src/main/java/org/jclouds/b2/features/BucketApi.java
andrewgaul: Sorry I do not understand this comment.  B2 does not return 404:

https://www.backblaze.com/b2/docs/b2_delete_bucket.html

`ParseB2ErrorFromJsonContent` already maps `bad_bucket_id` to `ContainerNotFoundException` which `testDeleteAlreadyDeletedBucket` exercises so what else can we do here?

nacx: Yes, it does not return a 404, but it actually throws a ResourceNotFoundException when the referenced bucket is not found (which in practice, for jclouds is just the same as if it had returned a 404).

What I mean, is that a common practice in jclouds is to avoid failing on delete operations if the resource to be deleted does not exist. [The S3 API](https://github.com/jclouds/jclouds/blob/master/apis/s3/src/main/java/org/jclouds/s3/S3Client.java#L207) and many others behave this way, and I think it would be good to follow that pattern here too.

If you annotate this method with `@Fallback(NullOnNotFoundOr404.class)`, then the method will not fail and just return `null` when the resource to be deleted does not exist, just like most of the other jclouds APIs. The fallback will not only take place if the response is a 404, [but also if the call generated a ResoueceNotFoundException](https://github.com/jclouds/jclouds/blob/master/core/src/main/java/org/jclouds/Fallbacks.java#L138-L140). Don't be "distracted" by the "Or404" in the fallback name; it checks for exceptions too, and your error handler already takes care of properly setting it when appropriate, so APIs can use this kind of callbacks.

andrewgaul: Done.

andrewgaul:
Addressed all outstanding review comments.

b2/src/main/java/org/jclouds/b2/config/B2HttpApiModule.java
```diff
@@ -0,0 +1,77 @@
...
+   @Override
+   protected void bindErrorHandlers() {
+      bind(HttpErrorHandler.class).annotatedWith(Redirection.class).to(ParseB2ErrorFromJsonContent.class);
+      bind(HttpErrorHandler.class).annotatedWith(ClientError.class).to(ParseB2ErrorFromJsonContent.class);
+      bind(HttpErrorHandler.class).annotatedWith(ServerError.class).to(ParseB2ErrorFromJsonContent.class);
+   }
+
+   @Provides
+   @Singleton
+   static Supplier<Authorization> provideAuthorizationCache(final B2Api b2Api) {
```
andrewgaul: I made some changes to the authorization cache to support upcoming commits.

b2/src/main/java/org/jclouds/b2/config/B2HttpApiModule.java
```diff
@@ -0,0 +1,77 @@
...
+                  return b2Api.getAuthorizationApi().authorizeAccount();
+               }
+            }));
+      return new Supplier<Authorization>() {
+            @Override
+            public Authorization get() {
+               return cache.getUnchecked(fakeKey);
+            }
+         };
+   }
```
nacx: A common practice in jclouds when you are providing a supplier on top of a cache and binding it to the Guice contex, is to use the jclouds memoized supplier. It already takes care of properly dealing with auth exceptions. It would be something like:
```java
@Provides
@Singleton
protected final Supplier<Authorization> provideAuthorizationSupplier(final B2Api b2Api) {
   return new Supplier<Authorization>() {
      @Override public Authorization get() {
         return b2Api.getAuthorizationApi().authorizeAccount();
      }
   };
}

@Provides
@Singleton
@Memoized
protected final Supplier<Authorization> provideAuthorizationCache(
      AtomicReference<AuthorizationException> authException,
      @Named(PROPERTY_SESSION_INTERVAL) long seconds,
      Supplier<Authorization>  uncached) {
   return MemoizedRetryOnTimeOutButNotOnAuthorizationExceptionSupplier.create(
      authException, uncached, seconds, TimeUnit.SECONDS);
```

With this you can inject both suppliers. If you want to inject the cache, you just need to qualify the variable as `@Memoized`.

Also note that the `@Provides` methods are better marked final, as Guice 4 does not allow to override them. I've also used the `PROPERTY_SESSION_INTERVAL` property to configure the auth expiration. Feel free to consider if that makes sense in this provider. If it does, you could configure the 23 hours default value for that property in the api metadata.

b2/src/test/java/org/jclouds/b2/features/BucketApiMockTest.java
```diff
@@ -0,0 +1,233 @@
...
+         assertThat(response.bucketName()).isEqualTo("any_name_you_pick");
+         assertThat(response.bucketType()).isEqualTo(BucketType.ALL_PRIVATE);
+
+         assertThat(server.getRequestCount()).isEqualTo(2);
+         assertAuthentication(server);
+         assertRequest(server.takeRequest(), "POST", "/b2api/v1/b2_delete_bucket", "/delete_bucket_request.json");
+      } finally {
+         server.shutdown();
+      }
+   }
```
nacx: Add a test that exercises the fallback?

nacx:
Only a couple comments, but good to merge for me! This is looking great!

b2/src/main/java/org/jclouds/b2/config/B2HttpApiModule.java
andrewgaul: Done.

b2/src/test/java/org/jclouds/b2/features/BucketApiMockTest.java
andrewgaul: Done.

andrewgaul:
Addressed all outstanding review comments.

andrewgaul:
@nacx Thanks for your feedback and help on this pull request!

PR #271, opened 2016-05-23
Support supplying key and certificate to Docker as data

grkvlt:
Support supplying key and certificate to Docker as data

PR #272, opened 2016-05-23
Support supplying key and certificate to Docker as data

grkvlt:
See #271 for equivalent changes in 1.9.x branch


andreaturli:
hi @grkvlt could you just comment on why is this needed?

grkvlt:
@andreaturli This was to allow passing in the certificate data directly if it was not available as a file, and makes it easier in some cases to use the jclouds driver with Brooklyn

andreaturli:
cool, thanks.
lgtm. Merging now

andreaturli:
merged at [master](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/c5d845d3)

PR #273, opened 2016-05-23
JCLOUDS-664 Azurecompute-arm DeploymentApi OSImageApi VMSizeApi

ritazh:
@nacx @andreaturli 
This PR is for DeploymentApi, OSImageApi, and VMSizeApi, part of the foundations needed for interacting with Azure Resource Manager (ARM) Rest APIs.

FYI
@jtjk @isalento @jmspring

### DIFF
--- a/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/ComputeNode.java
+++ b/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/ComputeNode.java
@@ -14,23 +14,18 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.jclouds.azurecompute.arm.config;
+package org.jclouds.azurecompute.arm.domain;
 
-/**
- * Configuration properties and constants used in Azure Resource Manager connections.
- */
-public class AzureComputeProperties {
+import org.jclouds.azurecompute.arm.util.GetEnumValue;
 
-   public static final String OPERATION_TIMEOUT = "jclouds.azurecompute.arm.operation.timeout";
+public class ComputeNode {
+   public enum Status {
+      GOOD,
+      BAD,
+      UNRECOGNIZED;
 
-   public static final String OPERATION_POLL_INITIAL_PERIOD = "jclouds.azurecompute.arm.operation.poll.initial.period";
-
-   public static final String OPERATION_POLL_MAX_PERIOD = "jclouds.azurecompute.arm.operation.poll.max.period";
-
-   public static final String TCP_RULE_FORMAT = "jclouds.azurecompute.arm.tcp.rule.format";
-
-   public static final String TCP_RULE_REGEXP = "jclouds.azurecompute.arm.tcp.rule.regexp";
-
-   public static final String STORAGE_API_VERSION = "2015-06-15";
-
+      public static Status fromValue(final String text) {
+         return (Status) GetEnumValue.fromValueOrDefault(text, Status.UNRECOGNIZED);
+      }
+   }
 }
--- a/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/Deployment.java
+++ b/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/Deployment.java
@@ -1,0 +1,254 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jclouds.azurecompute.arm.domain;
+
+import static com.google.common.collect.ImmutableList.copyOf;
+
+import java.util.List;
+import java.util.Map;
+
+import com.google.common.collect.ImmutableMap;
+import org.jclouds.azurecompute.arm.util.GetEnumValue;
+import org.jclouds.domain.JsonBall;
+import org.jclouds.javax.annotation.Nullable;
+
+import com.google.auto.value.AutoValue;
+import org.jclouds.json.SerializedNames;
+
+@AutoValue
+public abstract class Deployment {
+
+   public enum ProvisioningState {
+      ACCEPTED,
+      READY,
+      CANCELED,
+      FAILED,
+      DELETED,
+      SUCCEEDED,
+      RUNNING,
+      UNRECOGNIZED;
+
+      public static ProvisioningState fromValue(final String text) {
+         return (ProvisioningState) GetEnumValue.fromValueOrDefault(text, ProvisioningState.UNRECOGNIZED);
+      }
+   }
+
+   public enum DeploymentMode {
+      INCREMENTAL,
+      COMPLETE,
+      UNRECOGNIZED;
+
+      public static DeploymentMode fromValue(final String text) {
+         return (DeploymentMode) GetEnumValue.fromValueOrDefault(text, DeploymentMode.UNRECOGNIZED);
+      }
+   }
+
+   @AutoValue
+   public abstract static class TypeValue {
+      public abstract String type();
+
+      public abstract String value();
+
+      @SerializedNames({"type", "value"})
+      public static TypeValue create(final String type, final String value) {
+         return new AutoValue_Deployment_TypeValue(type, value);
+      }
+   }
+
+   @AutoValue
+   public abstract static class ProviderResourceType {
+      @Nullable
+      public abstract String resourceType();
+
+      @Nullable
+      public abstract List<String> locations();
+
+      @Nullable
+      public abstract List<String> apiVersions();
+
+      @Nullable
+      public abstract Map<String, JsonBall> properties();
+
+      @SerializedNames({"resourceType", "locations", "apiVersions", "properties"})
+      public static ProviderResourceType create(final String resourceType,
+                                                final List<String> locations,
+                                                final List<String> apiVersions,
+                                                @Nullable final Map<String, JsonBall> properties) {
+         return new AutoValue_Deployment_ProviderResourceType(resourceType,
+                 locations == null ? null : copyOf(locations),
+                 apiVersions == null ? null : copyOf(apiVersions),
+                 properties == null ? ImmutableMap.<String, JsonBall>builder().build() : ImmutableMap.copyOf(properties));
+      }
+   }
+
+   @AutoValue
+   public abstract static class Provider {
+      @Nullable
+      public abstract String id();
+
+      @Nullable
+      public abstract String namespace();
+
+      @Nullable
+      public abstract String registrationState();
+
+      @Nullable
+      public abstract List<ProviderResourceType> resourceTypes();
+
+      @SerializedNames({"id", "namespace", "registrationState", "resourceTypes"})
+      public static Provider create(final String id,
+                                    final String namespace,
+                                    final String registrationState,
+                                    final List<ProviderResourceType> resourceTypes) {
+         return new AutoValue_Deployment_Provider(id, namespace, registrationState, resourceTypes == null ? null : copyOf(resourceTypes));
+      }
+   }
+
+   @AutoValue
+   public abstract static class Dependency {
+      @Nullable
+      public abstract List<Dependency> dependencies();
+
+      @Nullable
+      public abstract List<Dependency> dependsOn();
+
+      @Nullable
+      public abstract String id();
+
+      @Nullable
+      public abstract String resourceType();
+
+      @Nullable
+      public abstract String resourceName();
+
+      @SerializedNames({"dependencies", "dependsOn", "id", "resourceType", "resourceName"})
+      public static Dependency create(final List<Dependency> dependencies,
+                                      final List<Dependency> dependsOn,
+                                      final String id,
+                                      final String resourceType,
+                                      final String resourceName) {
+         return new AutoValue_Deployment_Dependency(dependencies == null ? null : copyOf(dependencies),
+               dependsOn == null ? null : copyOf(dependsOn), id, resourceType, resourceName);
+      }
+   }
+
+   @AutoValue
+   public abstract static class ContentLink {
+      public abstract String uri();
+
+      @Nullable
+      public abstract String contentVersion();
+
+      @SerializedNames({"uri", "contentVersion"})
+      public static ContentLink create(final String uri, final String contentVersion) {
+         return new AutoValue_Deployment_ContentLink(uri, contentVersion);
+      }
+   }
+
+   @AutoValue
+   public abstract static class DeploymentProperties {
+      @Nullable
+      public abstract String provisioningState();
+
+      @Nullable
+      public abstract String correlationId();
+
+      @Nullable
+      public abstract String timestamp();
+
+      @Nullable
+      public abstract Map<String, JsonBall> outputs();
+
+      @Nullable
+      public abstract List<Provider> providers();
+
+      @Nullable
+      public abstract List<Dependency> dependencies();
+
+      @Nullable
+      public abstract Map<String, JsonBall> template();
+
+      @Nullable
+      public abstract ContentLink templateLink();
+
+      @Nullable
+      public abstract Map<String, JsonBall> parameters();
+
+      @Nullable
+      public abstract ContentLink parametersLink();
+
+      public abstract String mode();
+
+      // The entries below seem to be dynamic/not documented in the specification
+      @Nullable
+      public abstract String duration();
+
+      @Nullable
+      public abstract List<Map<String, String>> outputResources();
+
+      @SerializedNames({"provisioningState", "correlationId", "timestamp", "outputs", "providers", "dependencies", "template", "templateLink", "parameters", "parametersLink", "mode", "duration", "outputResources"})
+      public static DeploymentProperties create(final String provisioningState,
+                                                final String correlationId,
+                                                final String timestamp,
+                                                @Nullable final Map<String, JsonBall> outputs,
+                                                final List<Provider> providers,
+                                                final List<Dependency> dependencies,
+                                                final Map<String, JsonBall> template,
+                                                final ContentLink templateLink,
+                                                final Map<String, JsonBall> parameters,
+                                                final ContentLink parametersLink,
+                                                final String mode,
+                                                final String duration,
+                                                final List<Map<String, String>> outputResources) {
+         return new AutoValue_Deployment_DeploymentProperties(provisioningState,
+                                                              correlationId,
+                                                              timestamp,
+                                                              outputs == null ? ImmutableMap.<String, JsonBall>builder().build() : ImmutableMap.copyOf(outputs),
+                                                              providers == null ? null : copyOf(providers),
+                                                              dependencies == null ? null : copyOf(dependencies),
+                                                              template == null ? ImmutableMap.<String, JsonBall>builder().build() : ImmutableMap.copyOf(template),
+                                                              templateLink,
+                                                              parameters == null ? ImmutableMap.<String, JsonBall>builder().build() : ImmutableMap.copyOf(parameters),
+                                                              parametersLink,
+                                                              mode,
+                                                              duration,
+                                                              outputResources == null ? null : copyOf(outputResources));
+      }
+   }
+
+   /**
+    * The ID associated with the template deployment.
+    */
+   @Nullable
+   public abstract String id();
+
+   /**
+    * The name associated with the template deployment.
+    */
+   public abstract String name();
+
+   /**
+    * Properties of the deployment.
+    */
+   @Nullable
+   public abstract DeploymentProperties properties();
+
+   @SerializedNames({"id", "name", "properties"})
+   public static Deployment create(final String id, final String name, final DeploymentProperties properties) {
+      return new AutoValue_Deployment(id, name, properties);
+   }
+}
--- a/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/DeploymentTemplate.java
+++ b/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/DeploymentTemplate.java
@@ -1,0 +1,105 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jclouds.azurecompute.arm.domain;
+
+import com.google.auto.value.AutoValue;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import org.jclouds.javax.annotation.Nullable;
+
+import java.util.List;
+import java.util.Map;
+import org.jclouds.json.SerializedNames;
+
+@AutoValue
+public abstract class DeploymentTemplate {
+
+   //Empty placeholders as we want to generate the empty JSON object
+   @AutoValue
+   public abstract static class Parameters {
+      public static Parameters create() {
+         return new AutoValue_DeploymentTemplate_Parameters();
+      }
+   }
+
+   public abstract String schema();
+
+   public abstract String contentVersion();
+
+   public abstract Parameters parameters();
+
+   public abstract Map<String, String> variables();
+
+   public abstract List<ResourceDefinition> resources();
+
+   @Nullable
+   public abstract List<?> outputs();
+
+   @SerializedNames({"$schema", "contentVersion", "parameters", "variables", "resources" , "outputs"})
+   public static DeploymentTemplate create(final String schema,
+                                           final String contentVersion,
+                                           final Parameters parameters,
+                                           final Map<String, String> variables,
+                                           final List<ResourceDefinition> resources,
+                                           final List<?> outputs) {
+
+      DeploymentTemplate.Builder builder = DeploymentTemplate.builder()
+              .schema(schema)
+              .contentVersion(contentVersion)
+              .parameters(parameters);
+
+      if (variables != null)
+         builder.variables(variables);
+
+      if (resources != null)
+         builder.resources(resources);
+
+      builder.outputs(outputs == null ? null : ImmutableList.copyOf(outputs));
+
+      return builder.build();
+   }
+
+   public static Builder builder() {
+      return new AutoValue_DeploymentTemplate.Builder();
+   }
+
+   @AutoValue.Builder
+   public abstract static class Builder {
+      public abstract Builder schema(String schema);
+
+      public abstract Builder contentVersion(String type);
+
+      public abstract Builder parameters(Parameters parameters);
+
+      public abstract Builder variables(Map<String, String> variables);
+
+      public abstract Builder resources(List<ResourceDefinition> resources);
+
+      public abstract Builder outputs(List<?> outputs);
+
+      abstract Map<String, String> variables();
+      abstract List<ResourceDefinition> resources();
+
+      abstract DeploymentTemplate autoBuild();
+
+      public DeploymentTemplate build() {
+         variables(variables() != null ? ImmutableMap.copyOf(variables()) : null);
+         resources(resources() != null ? ImmutableList.copyOf(resources()) : null);
+         return autoBuild();
+      }
+   }
+}
--- a/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/features/DeploymentApi.java
+++ b/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/features/DeploymentApi.java
@@ -1,0 +1,109 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jclouds.azurecompute.arm.features;
+
+import javax.inject.Named;
+import javax.ws.rs.DELETE;
+import javax.ws.rs.POST;
+import javax.ws.rs.Path;
+import javax.ws.rs.Consumes;
+import javax.ws.rs.PUT;
+import javax.ws.rs.GET;
+import javax.ws.rs.Produces;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.core.MediaType;
+
+import org.jclouds.Fallbacks;
+import org.jclouds.Fallbacks.EmptyListOnNotFoundOr404;
+import org.jclouds.Fallbacks.NullOnNotFoundOr404;
+
+import org.jclouds.azurecompute.arm.domain.Deployment;
+import org.jclouds.azurecompute.arm.functions.URIParser;
+import org.jclouds.oauth.v2.filters.OAuthFilter;
+import org.jclouds.rest.annotations.QueryParams;
+import org.jclouds.rest.annotations.RequestFilters;
+import org.jclouds.rest.annotations.Fallback;
+import org.jclouds.rest.annotations.Payload;
+import org.jclouds.rest.annotations.PayloadParam;
+import org.jclouds.rest.annotations.ResponseParser;
+import org.jclouds.rest.annotations.SelectJson;
+
+import java.net.URI;
+import java.util.List;
+
+/**
+ * - create deployment
+ * - delete deployment
+ * - get information about deployment
+ */
+@Path("/resourcegroups/{resourcegroup}/providers/microsoft.resources/deployments")
+@QueryParams(keys = "api-version", values = "2016-02-01")
+@RequestFilters(OAuthFilter.class)
+@Consumes(MediaType.APPLICATION_JSON)
+public interface DeploymentApi {
+
+   /**
+    * The Create Template Deployment operation starts the process of an ARM Template deployment.
+    * It then returns a Deployment object.
+    */
+   @Named("deployment:create")
+   @Path("/{deploymentname}")
+   @Payload("{properties}")
+   @PUT
+   @Produces(MediaType.APPLICATION_JSON)
+   Deployment create(@PathParam("deploymentname") String deploymentname,
+                               @PayloadParam("properties") String properties);
+
+   /**
+    * Get Deployment Information returns information about the specified deployment.
+    */
+   @Named("deployment:get")
+   @Path("/{deploymentname}")
+   @GET
+   @Fallback(NullOnNotFoundOr404.class)
+   Deployment get(@PathParam("deploymentname") String deploymentname);
+
+   /**
+    * Validate Deployment validates deployment template before deployment
+    */
+   @Named("deployment:validate")
+   @Path("/{deploymentname}/validate")
+   @Payload("{properties}")
+   @POST
+   @Produces(MediaType.APPLICATION_JSON)
+   Deployment validate(@PathParam("deploymentname") String deploymentname,
+                                 @PayloadParam("properties") String properties);
+
+   /**
+    * List all deployments in a resource group
+    */
+   @Named("deployment:list")
+   @GET
+   @SelectJson("value")
+   @Fallback(EmptyListOnNotFoundOr404.class)
+   List<Deployment> list();
+
+   /**
+    * The Delete Template Deployment operation starts the process of an ARM Template removal.
+    */
+   @Named("deployment:delete")
+   @DELETE
+   @ResponseParser(URIParser.class)
+   @Path("/{deploymentname}")
+   @Fallback(Fallbacks.NullOnNotFoundOr404.class)
+   URI delete(@PathParam("deploymentname") String deploymentname);
+}
--- a/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/features/OSImageApi.java
+++ b/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/features/OSImageApi.java
@@ -1,0 +1,84 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jclouds.azurecompute.arm.features;
+
+import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
+
+import java.util.List;
+
+import javax.inject.Named;
+import javax.ws.rs.Consumes;
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+
+import org.jclouds.Fallbacks.EmptyListOnNotFoundOr404;
+import org.jclouds.azurecompute.arm.domain.Offer;
+import org.jclouds.azurecompute.arm.domain.Publisher;
+import org.jclouds.azurecompute.arm.domain.SKU;
+import org.jclouds.azurecompute.arm.domain.Version;
+import org.jclouds.oauth.v2.filters.OAuthFilter;
+import org.jclouds.rest.annotations.Fallback;
+import org.jclouds.rest.annotations.QueryParams;
+import org.jclouds.rest.annotations.RequestFilters;
+
+/**
+ * The Azure Resource Management API includes operations for managing the OS images in your subscription.
+ */
+@Path("/providers/Microsoft.Compute/locations/{location}")
+@RequestFilters(OAuthFilter.class)
+@QueryParams(keys = "api-version", values = "2015-06-15")
+@Consumes(APPLICATION_JSON)
+public interface OSImageApi {
+
+   /**
+    * List Publishers in location
+    */
+   @Named("publisher:list")
+   @GET
+   @Path("/publishers")
+   @Fallback(EmptyListOnNotFoundOr404.class)
+   List<Publisher> listPublishers();
+
+   /**
+    * List Offers in publisher
+    */
+   @Named("offer:list")
+   @GET
+   @Path("/publishers/{publisher}/artifacttypes/vmimage/offers")
+   @Fallback(EmptyListOnNotFoundOr404.class)
+   List<Offer> listOffers(@PathParam("publisher") String publisher);
+
+   /**
+    * List SKUs in offer
+    */
+   @Named("sku:list")
+   @GET
+   @Path("/publishers/{publisher}/artifacttypes/vmimage/offers/{offer}/skus")
+   @Fallback(EmptyListOnNotFoundOr404.class)
+   List<SKU> listSKUs(@PathParam("publisher") String publisher, @PathParam("offer") String offer);
+
+   /**
+    * List Versions in SKU
+    */
+   @Named("version:list")
+   @GET
+   @Path("/publishers/{publisher}/artifacttypes/vmimage/offers/{offer}/skus/{sku}/versions")
+   @Fallback(EmptyListOnNotFoundOr404.class)
+   List<Version> listVersions(@PathParam("publisher") String publisher, @PathParam("offer") String offer,
+                          @PathParam("sku") String sku);
+}
--- a/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/features/VMSizeApi.java
+++ b/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/features/VMSizeApi.java
@@ -1,0 +1,45 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jclouds.azurecompute.arm.features;
+
+import org.jclouds.Fallbacks.EmptyListOnNotFoundOr404;
+import org.jclouds.azurecompute.arm.domain.VMSize;
+import org.jclouds.oauth.v2.filters.OAuthFilter;
+import org.jclouds.rest.annotations.Fallback;
+import org.jclouds.rest.annotations.QueryParams;
+import org.jclouds.rest.annotations.RequestFilters;
+import org.jclouds.rest.annotations.SelectJson;
+
+import javax.inject.Named;
+import javax.ws.rs.Consumes;
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.core.MediaType;
+import java.util.List;
+
+@Path("/providers/Microsoft.Compute/locations/{location}/vmSizes")
+@QueryParams(keys = "api-version", values = "2015-06-15")
+@RequestFilters(OAuthFilter.class)
+@Consumes(MediaType.APPLICATION_JSON)
+public interface VMSizeApi {
+
+   @Named("vmSizes:list")
+   @GET
+   @SelectJson("value")
+   @Fallback(EmptyListOnNotFoundOr404.class)
+   List<VMSize> list();
+}
--- a/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/functions/ParseJobStatus.java
+++ b/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/functions/ParseJobStatus.java
@@ -30,6 +30,7 @@ public class ParseJobStatus implements Function<HttpResponse, ParseJobStatus.Job
       DONE,
       IN_PROGRESS,
       FAILED,
+      NO_CONTENT,
       UNRECOGNIZED;
 
       public static JobStatus fromValue(final String text) {
@@ -39,6 +40,8 @@ public class ParseJobStatus implements Function<HttpResponse, ParseJobStatus.Job
    public JobStatus apply(final HttpResponse from) {
       if (from.getStatusCode() == 202 ){
          return JobStatus.IN_PROGRESS;
+      } else if (from.getStatusCode() == 204) {
+         return JobStatus.NO_CONTENT;
       } else if (from.getStatusCode() == 200 ){
          return JobStatus.DONE;
       } else {
--- a/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/util/DeploymentTemplateBuilder.java
+++ b/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/util/DeploymentTemplateBuilder.java
@@ -1,0 +1,375 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jclouds.azurecompute.arm.util;
+
+import com.google.common.collect.ImmutableMap;
+import com.google.inject.assistedinject.Assisted;
+
+import org.jclouds.azurecompute.arm.domain.DeploymentProperties;
+import org.jclouds.compute.options.TemplateOptions;
+import org.jclouds.azurecompute.arm.domain.DataDisk;
+import org.jclouds.azurecompute.arm.domain.DeploymentBody;
+import org.jclouds.azurecompute.arm.domain.DeploymentTemplate;
+import org.jclouds.azurecompute.arm.domain.DiagnosticsProfile;
+import org.jclouds.azurecompute.arm.domain.DnsSettings;
+import org.jclouds.azurecompute.arm.domain.HardwareProfile;
+import org.jclouds.azurecompute.arm.domain.IdReference;
+import org.jclouds.azurecompute.arm.domain.ImageReference;
+import org.jclouds.azurecompute.arm.domain.IpConfiguration;
+import org.jclouds.azurecompute.arm.domain.IpConfigurationProperties;
+import org.jclouds.azurecompute.arm.domain.NetworkInterfaceCardProperties;
+import org.jclouds.azurecompute.arm.domain.NetworkProfile;
+import org.jclouds.azurecompute.arm.domain.OSDisk;
+import org.jclouds.azurecompute.arm.domain.OSProfile;
+import org.jclouds.azurecompute.arm.domain.PublicIPAddressProperties;
+import org.jclouds.azurecompute.arm.domain.ResourceDefinition;
+import org.jclouds.azurecompute.arm.domain.StorageProfile;
+import org.jclouds.azurecompute.arm.domain.StorageService;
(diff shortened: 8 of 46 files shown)


ritazh:
@nacx The build seems to break on jclouds docker api. Let me know if I need to trigger a fresh build.

ritazh:
_bump_ @nacx I have split the previous PR #267 as two. This one only has the APIs. Thanks!

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/options/AzureComputeArmTemplateOptions.java
```diff
@@ -0,0 +1,419 @@
...
+ * ComputeService client = // get connection
+ * templateBuilder.options(inboundPorts(22, 80, 8080, 443));
+ * Set&lt;? extends NodeMetadata&gt; set = client.createNodesInGroup(tag, 2, templateBuilder.build());
+ * </pre>
+ *
+ */
+public class AzureComputeArmTemplateOptions extends TemplateOptions implements Cloneable {
+
+   protected String virtualNetworkName;
+   protected List<String> subnetNames = ImmutableList.of();
```
nacx: The template options already have a way to provide the IDs of the networks where the VMs have to be attached. Can that property be used instead of having these ones?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/options/AzureComputeArmTemplateOptions.java
```diff
@@ -0,0 +1,419 @@
...
+   protected String storageAccountName;
+   protected String storageAccountType;
```
nacx: What are these options used for?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/options/AzureComputeArmTemplateOptions.java
```diff
@@ -0,0 +1,419 @@
...
+   protected String networkSecurityGroupName;
```
nacx: The template options already provide methods to specify the security groups to be added to the VM. This one seems to be redundant.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/options/AzureComputeArmTemplateOptions.java
```diff
@@ -0,0 +1,419 @@
...
+      return options;
+   }
+
+   @Override
+   public void copyTo(final TemplateOptions to) {
+      super.copyTo(to);
+      if (to instanceof AzureComputeArmTemplateOptions) {
+         final AzureComputeArmTemplateOptions eTo = AzureComputeArmTemplateOptions.class.cast(to);
+         eTo.virtualNetworkName(virtualNetworkName);
+         if (!subnetNames.isEmpty()) {
```
nacx: A clone operation should be consistent. Why not cloning an empty list? If not, the result is not a clone, but an aggregate.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/options/AzureComputeArmTemplateOptions.java
```diff
@@ -0,0 +1,419 @@
...
+            eTo.subnetNames(subnetNames);
+         }
+         eTo.storageAccountName(storageAccountName);
+         eTo.storageAccountType(storageAccountType);
+         eTo.reservedIPName(reservedIPName);
```
nacx: Missing the security groups property (but don't add it if it is going to be removed as per my previous comment).

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/ComputeNode.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+package org.jclouds.azurecompute.arm.domain;
+
+public class ComputeNode {
+   public enum Status {
+      GOOD,
+      BAD,
+      UNRECOGNIZED;
+
+      public static Status fromString(final String text) {
```
nacx: Convention is to name the method `fromValue`, to take benefit from the [EnumTypeAdapterThatReturnsFromValue](https://github.com/jclouds/jclouds/blob/master/core/src/main/java/org/jclouds/json/internal/EnumTypeAdapterThatReturnsFromValue.java) if this has to be deserialised. Also apply this to all the enums.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/Deployment.java
```diff
@@ -0,0 +1,277 @@
...
+      ACCEPTED("Accepted"),
+      READY("Ready"),
+      CANCELED("Canceled"),
+      FAILED("Failed"),
+      DELETED("Deleted"),
+      SUCCEEDED("Succeeded"),
+      RUNNING("Running"),
+      UNRECOGNIZED("");
+
+      private final String key;
```
nacx: This looks redundant. You can omit it and just compare by name()?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/Deployment.java
```diff
@@ -0,0 +1,277 @@
...
+         return UNRECOGNIZED;
+      }
+   }
+
+   public enum DeploymentMode {
+      INCREMENTAL("Incremental"),
+      COMPLETE("Complete"),
+      UNRECOGNIZED("");
+
+      private final String key;
```
nacx: Same here.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/Deployment.java
```diff
@@ -0,0 +1,277 @@
...
+      public abstract String resourceType();
+
+      @Nullable
+      public abstract List<String> locations();
+
+      @Nullable
+      public abstract List<String> apiVersions();
+
+      @Nullable
+      public abstract Map<String, JsonBall> properties();
```
nacx: Is there a need for collections to be nullable? Is this the same case than the `tags` properties?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/Deployment.java
```diff
@@ -0,0 +1,277 @@
...
+      public abstract String id();
+
+      @Nullable
+      public abstract String namespace();
+
+      @Nullable
+      public abstract String registrationState();
+
+      @Nullable
+      public abstract List<ProviderResourceType> resourceTypes();
```
nacx: Same here about nullable collections.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/DeploymentTemplate.java
```diff
@@ -0,0 +1,103 @@
...
+   public abstract Parameters parameters();
+
+   public abstract Map<String, String> variables();
+
+   public abstract List<ResourceDefinition> resources();
+
+   @Nullable
+   public abstract List<?> outputs();
+
+   //@SerializedNames({"$schema", "contentVersion", "parameters", "variables", "resources" , "outputs"})
```
nacx: Why is this commented?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/DeploymentTemplate.java
```diff
@@ -0,0 +1,103 @@
...
+
+   //Empty placeholders as we want to generate the empty JSON object
+   @AutoValue
+   public abstract static class Parameters {
+      public static Parameters create() {
+         return new AutoValue_DeploymentTemplate_Parameters();
+      }
+   }
+
+   public abstract String $schema();
```
nacx: Can we remove the `$` character fro the variable?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/features/DeploymentApi.java
```diff
@@ -0,0 +1,110 @@
...
+   /**
+    * The Create Template Deployment operation starts the process of an ARM Template deployment.
+    * It then returns a Deployment object.
+    */
+   @Named("deployment:create")
+   @Path("/{deploymentname}")
+   @Payload("{properties}")
+   @PUT
+   @Produces(MediaType.APPLICATION_JSON)
+   @Fallback(NullOnNotFoundOr404.class)
```
nacx: Don't use 404 fallbacks in POST/PUT

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/features/DeploymentApi.java
```diff
@@ -0,0 +1,110 @@
...
+   Deployment get(@PathParam("deploymentname") String deploymentname);
+
+   /**
+    * Validate Deployment validates deployment template before deployment
+    */
+   @Named("deployment:validate")
+   @Path("/{deploymentname}/validate")
+   @Payload("{properties}")
+   @POST
+   @Fallback(NullOnNotFoundOr404.class)
```
nacx: Remove fallback

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/features/DeploymentApi.java
```diff
@@ -0,0 +1,110 @@
...
+   List<Deployment> list();
+
+   /**
+    * The Delete Template Deployment operation starts the process of an ARM Template removal.
+    */
+   @Named("deployment:delete")
+   @DELETE
+   @ResponseParser(URIParser.class)
+   @Path("/{deploymentname}")
+   @Fallback(Fallbacks.VoidOnNotFoundOr404.class)
```
nacx: `NullOnNotFoundOr404`

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/features/OSImageApi.java
```diff
@@ -0,0 +1,89 @@
...
+@Path("/providers/Microsoft.Compute/locations/{location}")
+@RequestFilters(OAuthFilter.class)
+@QueryParams(keys = "api-version", values = "2015-06-15")
+@Consumes(APPLICATION_JSON)
+public interface OSImageApi {
+
+   /**
+    * List Publishers in location
+    */
+   @Named("ListPublishers")
```
nacx: Follow the same naming pattern than in the other apis.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/features/OSImageApi.java
```diff
@@ -0,0 +1,89 @@
...
+   @GET
+   @Produces(APPLICATION_JSON)
```
nacx: GET methods don't produce a body

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/features/VMSizeApi.java
```diff
@@ -0,0 +1,47 @@
...
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+import java.util.List;
+
+@Path("/providers/Microsoft.Compute/locations/{location}/vmSizes")
+@QueryParams(keys = "api-version", values = "2015-06-15")
+@RequestFilters(OAuthFilter.class)
+@Produces(MediaType.APPLICATION_JSON)
```
nacx: There are no methods here that produce a body

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/util/DeploymentTemplateBuilder.java
```diff
@@ -0,0 +1,356 @@
...
+              )
+              .caching("ReadWrite")
+              .createOption("FromImage")
+              .build();
+
+
+      //Create Data Disk(s) and add to list
+      final String dataDiskName = name + "datadisk";
+      variables.put("dataDiskName", dataDiskName);
+      final String dataDiskSize = "100";
```
nacx: Is this meant to be hardcoded? Would this be extracted from the hardware profile or a template option?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/util/DeploymentTemplateBuilder.java
```diff
@@ -0,0 +1,356 @@
...
+              .build();
+
+      resources.add(storageAccount);
+   }
+
+   private void addVirtualNetworkResource() {
+      String virtualNetworkName = group + "virtualnetwork";
+      String vnAddresSpacePrefix = "10.0.0.0/16";
+      String subnetName = group + "subnet";
+      String subnetAddressPrefix = "10.0.0.0/24";
```
nacx: Is this range meant to be hardcoded?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/util/DeploymentTemplateBuilder.java
```diff
@@ -0,0 +1,356 @@
...
+   private void addStorageResource() {
+      String storageAccountName = name.replaceAll("[^A-Za-z0-9 ]", "") + "storage";
+
+      variables.put("storageAccountName", storageAccountName);
+
+      ResourceDefinition storageAccount = ResourceDefinition.builder()
+              .name("[variables('storageAccountName')]")
+              .type("Microsoft.Storage/storageAccounts")
+              .location(location)
+              .apiVersion("2015-06-15")
```
nacx: Move this to a constant so the value is shared by the APIs too? (consider this elsewhere).

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceContextLiveTest.java
```diff
@@ -0,0 +1,292 @@
+/*
```
nacx: Remove the class from this PR, as there is no compute service yet.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/DeploymentApiLiveTest.java
```diff
@@ -0,0 +1,127 @@
...
+        String parameters = "{\"newStorageAccountName\":{\"value\":\"" + resourceName + "\"},\"storageAccountType\":{\"value\":\"Standard_LRS\"},\"location\":{\"value\":\"West US\"}}";
+        String properties = getPutBody(template, "Incremental", parameters);
```
nacx: Shouldn't this use the DeploymentTemplateBuilder?

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/DeploymentApiLiveTest.java
```diff
@@ -0,0 +1,127 @@
...
+
+        Deployment deployment = api().create(deploymentName, properties);
```
nacx: Looking more closely at this method, instead of receiving a json string, shouldn't it receive a DeploymentTemplate parameter?

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/DeploymentApiLiveTest.java
```diff
@@ -0,0 +1,127 @@
...
+        assertNotNull(deployment);
+
+        ProvisioningState state = ProvisioningState.fromString(deployment.properties().provisioningState());
+        int testTime = 0;
+        while (testTime < maxTestDuration) {
+            if ((state == ProvisioningState.SUCCEEDED) ||
```
nacx: Use the jclouds style retries, and consider providing a generic predicate that could be injected. This was already commented on past pull requests.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/DeploymentApiLiveTest.java
```diff
@@ -0,0 +1,127 @@
...
+        Deployment deployment = api().get(deploymentName);
+        assertNotNull(deployment);
+        ProvisioningState state = ProvisioningState.fromString(deployment.properties().provisioningState());
+        assertTrue(state == ProvisioningState.SUCCEEDED);
+    }
+
+    @Test(groups = "live", dependsOnMethods = "testCreate")
+    public void testListDeployments() {
+        List<Deployment> deployments = api().list();
+        assertTrue(deployments.size() > 0);
```
nacx: Verify that the created deployment exists.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/DeploymentApiLiveTest.java
```diff
@@ -0,0 +1,127 @@
...
+                    assertTrue(uri.toString().contains("operationresults"));
+                }
+            }
+        }
+    }
+
+    private DeploymentApi api() {
+        return api.getDeploymentApi(resourceGroup);
+    }
+}
```
nacx: Also test the `validate` method

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/DeploymentApiMockTest.java
```diff
@@ -0,0 +1,105 @@
...
+    @Test
+    public void testNullGetDeployment() throws Exception {
+        final DeploymentApi deploymentApi = api.getDeploymentApi(resourceGroup);
+
+        server.enqueue(new MockResponse().setResponseCode(404));
+        assertNull(deploymentApi.get("randomName"));
+
+        assertSent(server, "GET", "/subscriptions/" + subscriptionId + "/resourcegroups/" + resourceGroup +
+                "/providers/microsoft.resources/deployments/randomName?api-version=2016-02-01");
+    }
```
nacx: Add all missing tests: fallbacks, delete...

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/DeploymentApiMockTest.java
```diff
@@ -0,0 +1,105 @@
...
+        String parameters = "{\"newStorageAccountName\":{\"value\":\"" + resourceName + "\"},\"storageAccountType\":{\"value\":\"Standard_LRS\"},\"location\":{\"value\":\"West US\"}}";
+        String properties = getPutBody(template, "Incremental", parameters);
```
nacx: Use the builders and java objects.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/functions/ParseJobStatus.java
```diff
@@ -30,6 +30,7 @@
       DONE,
       IN_PROGRESS,
       FAILED,
+      NO_CONTENT,
```
nacx: What does this status mean?

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/NetworkInterfaceCardApiLiveTest.java
```diff
@@ -66,8 +69,18 @@ public void setup() {
    public void deleteNetworkInterfaceCardResourceDoesNotExist() {
 
       final NetworkInterfaceCardApi nicApi = api.getNetworkInterfaceCardApi(resourcegroup);
-      boolean status = nicApi.delete(NETWORKINTERFACECARD_NAME);
-      assertFalse(status);
+      URI uri = nicApi.delete(NETWORKINTERFACECARD_NAME);
```
nacx: We expect this test to return null, so just assert that the URI is actually null and forget about the rest..

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/OSImageApiLiveTest.java
```diff
@@ -0,0 +1,77 @@
...
+
+import java.util.List;
+
+@Test(groups = "live", testName = "OSImageApiLiveTest")
+public class OSImageApiLiveTest extends AbstractAzureComputeApiLiveTest {
+
+   @Test
+   public void testPublisher() {
+      List<Publisher> publishers = api().listPublishers();
+      System.out.println(publishers.size());
```
nacx: Remove all system outs.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/OSImageApiMockTest.java
```diff
@@ -0,0 +1,80 @@
...
+
+   public void testVersions() throws InterruptedException {
+      server.enqueue(jsonResponse("/versions.json"));
+
+      List<Version> versions = api.getOSImageApi(location).listVersions(publisher, offer, sku);
+
+      assertEquals(size(versions), 2);
+
+      assertSent(server, "GET", requestUrl + "/" + publisher + "/artifacttypes/vmimage/offers/" + offer + "/skus/" + sku + "/versions" + apiversion);
+   }
```
nacx: Add missing fallback tests.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/TemplateToDeploymentTemplateLiveTest.java
```diff
@@ -0,0 +1,196 @@
...
+
+        String deploymentTemplate = json.toJson(properties);
+        deploymentTemplate = UrlEscapers.urlFormParameterEscaper().escape(deploymentTemplate);
+
+        //creates an actual VM using deployment template
+        Deployment deployment = api().create(deploymentName, deploymentTemplate);
+
+        Deployment.ProvisioningState state = Deployment.ProvisioningState.fromString(deployment.properties().provisioningState());
+        int testTime = 0;
+        while (testTime < maxTestDuration) {
```
nacx: Same comment about polling and reusable predicates bound to the Guice injector. This predicates are used in tests and potentially by the compute service, so please, make them injectable so they can be shared, and the polling parameters configured.

nacx:
Thanks @ritazh! Initial review done. Apart from the inline comments, please apply the jclouds formatting to all the classes.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/options/AzureComputeArmTemplateOptions.java
ritazh: It's used by `DeploymentTemplateBuilder` [here](https://github.com/ritazh/jclouds-labs/blob/74f8e9bf66e7427298c5c3c9567aa7d8284e0987/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/util/DeploymentTemplateBuilder.java#L74)

nacx: Oh, I didn't mean "where" are they used, but "for what purpose" the storageAccountName and storageAccountType options are used, just to understand how the provider is supposed to work :)
(This said, this concrete storageAccountType property seems not to be used in the class you referenced?)

ritazh: You are right. I only see reference of `authorizePublicKey`, `inboundPorts`, `getPublicKey`, `getLoginUser`, `getLoginPassword`. @jtjk can you please take a look?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/DeploymentTemplate.java
ritazh: `$schema` is actually a json keyword. It is used here to specify the schema of the template. Check out this [reference](https://msdn.microsoft.com/en-us/library/azure/dn790564.aspx). 

```
"properties": {
    "template": {
      "$schema": "http://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
      "contentVersion": "1.0.0.0",
...
```

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/Deployment.java
ritazh: Correct. It can be null.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/Deployment.java
ritazh: Correct. It can be null.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/features/DeploymentApi.java
ritazh: ok.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/util/DeploymentTemplateBuilder.java
ritazh: @jtjk @isalento can you guys please take a look at this?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/util/DeploymentTemplateBuilder.java
ritazh: @jtjk @isalento please take a look at this.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/util/DeploymentTemplateBuilder.java
ritazh: @jtjk @isalento please take a look at this.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/DeploymentApiLiveTest.java
ritazh: Azure ARM deployment utilizes ARM templates to describe resources to create for a given deployment. `DeploymentApi` is mapped to the [DeploymentApi on Azure](https://msdn.microsoft.com/en-us/library/azure/dn790549.aspx). `DeploymentTemplateBuilder` is the utility to construct the actual ARM template. @jtjk please add anything I may have missed.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/DeploymentTemplate.java
nacx: Yes, I know, but having special characters as part of variable names is not very Java. I suggested to name the property "in Java" just `schema` and configure the `@SerializedNames` to be `$schema`, so you can use the right json value while keeping a clean naming in the Java model.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/DeploymentApiLiveTest.java
nacx: Yes, I understand that. What I mean is that a deployment template is actually a json that goes directly in the request body. Despite the fact that it is an ARM template (conceptually), there is no practical differente with other API requests. You are just sending a request with a json body.

But the user experience, in this case is worse: Users have to build on their own that json. They can use the builders, but they have to manually serialize the template into a json, just because our `DeploymentApi` methods expect the template to be a json String.

What I wanted to say is that the DeploymentApi methods should be changed to receive a "DeploymentTemplate" (or whatever name it is appropriate) object, instead of a raw json, as it does now with the "properties" string. Let jclouds serialize the ARM template, but keep users just using high level objects. They shouldn't have to worry about generating raw json.

jmspring: @nacx - The plan was to basically reflect the raw API itself, thus the three JSON inputs.  For generating VMs, users should use the higher level functions, but the process within the actual code will essentially be declare as a JClouds template, translate from that representation to an ARM template (JSON) then use the DeploymentAPI to deploy said template.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/options/AzureComputeArmTemplateOptions.java
jtjk: You are correct. We should remove all obsolete fields from AzureComputeArmTemplateOptions. I guess some of the fields were left there as they existed in original Azure template options and we thought that there might be use for them later. In this phase it would be good to remove all unused fields.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/options/AzureComputeArmTemplateOptions.java
jtjk: Yep... I commented to first comment already. So we can remove all unused fields

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/util/DeploymentTemplateBuilder.java
jtjk: This is little bit tricky. If we have constant like STORAGE_API_VERSION and then we update it and update also StorageAccountAPI, we may face problems with this part of code. Maybe we should have that constant but in that case we should be very careful when updating the version to remember to update all code that is affected.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/util/DeploymentTemplateBuilder.java
isalento: Hi, Yes eventually data disk information (number of disks and their sizes) should be set using template option. In Azure the number of supported data disks depends on hardware in use, but the max size for singe disk is 1023 GB. Should we just drop data disk creation until Azure template option has support for it or create a singe data disk using 1023 as default size?

nacx: >Should we just drop data disk creation until Azure template option has support for it or create a singe data disk using 1023 as default size?

+1

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/DeploymentTemplate.java
isalento: DeploymentTemplate is only used to produce JSON string that is passed to the deployment API. It is never used to parse JSON. Adding @SerializedNames does not seem to have desired outcome, after serialization the output is still "{"properties":{"template":{"schema":"https.,".  I'll investigate more.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/DeploymentApiLiveTest.java
ritazh: > I don't think anyone would really use it directly

@nacx Would users ever call the DeploymentApi directly? As @jmspring mentioned, if users always leverage jclouds templates to provision everything, then the APIs should reflect the raw APIs. WDYT?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/DeploymentTemplate.java
jmspring: @nacx It looks like SerializedNames only goes one direction (deserialization).  In order to support this, especially with AutoValue, we need serialization.  Maybe I've missed it, but is there support for this?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/util/DeploymentTemplateBuilder.java
isalento: Fair point. Maybe not hardcoded, but we need to have default values for vn / subnet CIDR blocks as Azure does not assign default values internally. Basically we need to set some values, unless user overrides these by using Azure specific template options (which is not yet implemented).

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/functions/ParseJobStatus.java
ritazh: Returns 204-No Content if resource does not exist. https://msdn.microsoft.com/en-us/library/azure/mt163632.aspx Used by `PublicIPAddress`, 'SubnetApi`, and `VirtualNetworkApi`.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/DeploymentApiLiveTest.java
nacx: >The plan was to basically reflect the raw API itself, thus the three JSON inputs

Well, the three inputs only exist in the test. The method int he API just gets one raw JSON parameter, which turns to be the request body. Every single method in the raw API accepts a json payload. In this case, it happens to be a deployment template, but all other method in the API accept raw jsons too, be it a network, a disk or a deployment template. Why should we be providing high level and object oriented methods for some methods, and some other methods that require already serialized json? I understand that the template is a json object, but I think the user experience is not good here: users need to know serialization details. Given that we already have builder objects to be the template "users should be serializing", couldn't we let the api method accept what that builder produces, and let jclouds serialize it? Looks much less error-prone and a better user experience?

>Would users ever call the DeploymentApi directly

jclouds should allow users to call the DeploymentApi directly, yes. And they'll do, eventually, when they want to tweak a deployment in a way the portable abstraction does not let them to do so.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/util/DeploymentTemplateBuilder.java
nacx: I mean, if I get this right, this concerete version should match the version configured in the StorageAccountAPI. I'm just saying to create a constant and use it here and in the Storage API, and use one different constant for each API so it can be reused here, to make sure there are no version mismatches.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/util/DeploymentTemplateBuilder.java
nacx: +1 to have these values as defaults then!

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/DeploymentApiLiveTest.java
jmspring: I happen to disagree, because the actual API is being reflected.  There may be cases where an ARM template that isn't exposed via a builder -- maybe Azure Container Service, for instance -- may want to be used.  The user should have access to the raw deployment API do do that.

I agree builders are a better abstraction and those, to me, should be at a higher level and not necessarily imposed on the actual API implementation itself.

My biggest two concerns are the imposing of "object oriented" methods on top of a defined API that is basically pre-structured JSON (yes, I know it is error prone); and second, requiring builders reduces the generality of the API, if one were to want to use it in a manner that JClouds (and the arm provider) didn't expose directly.

ritazh: > But the user experience, in this case is worse: Users have to build on their own that json. They can use the builders, but they have to manually serialize the template into a json, just because our DeploymentApi methods expect the template to be a json String.

I agree the user experience is not great today, but this is how users are interfacing with the Azure ARM template deployment process today and this is the intended behavior of the original Deployment API. From the official API doc, the template "Specifies the JSON definition for the template. You use this element when you want to pass the template syntax directly in the request rather than link to an existing template."

If users can call the DeploymentApi directly, then in this case, users will expect to pass in a raw json template to the api much like how they interface with the original api anywhere whether it's thru a REST api call, or thru the Azure CLI, or thru an SDK. [Here is the reference](https://azure.microsoft.com/en-us/documentation/articles/resource-group-authoring-templates/)

Given that this is the design of this particular Deployment API and the Azure ARM template deployment process, we should probably follow the intent of the original api and ensure we give users the same experience as they expect it elsewhere. 

However if users use this thru the jclouds api, then we want to abstract all this raw template business by providing developers with the `DeploymentTemplateBuilder` so they do not have to deal with creating the raw json but instead pass in the parameters and we will build the json template for them.

@nacx WDYT? The current implementation allows users to use the DeploymentApi directly with the raw json template as intended by the original api, but it also supports abstracting the raw json template with a builder if users use jclouds apis.

nacx: Ok, let's keep the current API signature then. Thanks for taking your tine to discuss and dissgree! :)

>WDYT? The current implementation allows users to use the DeploymentApi directly with the raw json template as intended by the original api, but it also supports abstracting the raw json template with a builder if users use jclouds apis

That sounds good. Does the builder already generate the json?

ritazh: yes it does.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/util/DeploymentTemplateBuilder.java
ritazh: Added `public static final String STORAGE_API_VERSION = "2015-06-15";` to `AzureComputeProperties` and referencing it in both `StorageAccountApi` and `DeploymentTemplateBuilder`.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/NetworkInterfaceCardApiLiveTest.java
ritazh: Done.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/OSImageApiMockTest.java
ritazh: Done.

ritazh:
@nacx Ready for another round of review. The only outstanding question is how to deal with `$schema`.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/DeploymentTemplate.java
ritazh: @nacx We are using `org.jclouds.json.SerializedNames` and I do not think it supports mapping `schema` to `$schema` here. Do you suggest we use another implementation of `SerializedNames` or were you thinking of something else?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/util/DeploymentTemplateBuilder.java
```diff
@@ -0,0 +1,359 @@
...
+                              ))
+              )
+              .build();
+
+
+      ResourceDefinition virtualNetwork = ResourceDefinition.builder()
+              .name("[variables('virtualNetworkName')]")
+              .type("Microsoft.Network/virtualNetworks")
+              .location(location)
+              .apiVersion("2015-06-15")
```
nacx: Add a constant shared with the network api.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/util/DeploymentTemplateBuilder.java
```diff
@@ -0,0 +1,359 @@
...
+
+      properties.publicIPAllocationMethod("Dynamic");
+      variables.put("publicIPAddressName", publicIPAddressName);
+      variables.put("publicIPAddressReference", "[resourceId('Microsoft.Network/publicIPAddresses',variables('publicIPAddressName'))]");
+
+      ResourceDefinition publicIpAddress = ResourceDefinition.builder()
+              .name("[variables('publicIPAddressName')]")
+              .type("Microsoft.Network/publicIPAddresses")
+              .location(location)
+              .apiVersion("2015-06-15")
```
nacx: Add a constant shared with the PublicIpAddresses api

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/util/DeploymentTemplateBuilder.java
```diff
@@ -0,0 +1,359 @@
...
+
+      String networkInterfaceCardName = name + "nic";
+      variables.put("networkInterfaceCardName", networkInterfaceCardName);
+      variables.put("networkInterfaceCardReference", "[resourceId('Microsoft.Network/networkInterfaces',variables('networkInterfaceCardName'))]");
+
+      ResourceDefinition networkInterfaceCard = ResourceDefinition.builder()
+              .name("[variables('networkInterfaceCardName')]")
+              .type("Microsoft.Network/networkInterfaces")
+              .location(location)
+              .apiVersion("2015-06-15")
```
nacx: Add a constant here too

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/util/DeploymentTemplateBuilder.java
```diff
@@ -0,0 +1,359 @@
...
+              .networkProfile(networkProfile)
+              .diagnosticsProfile(diagnosticsProfile)
+              .build();
+
+      variables.put("virtualMachineName", name);
+      ResourceDefinition virtualMachine = ResourceDefinition.builder()
+              .name("[variables('virtualMachineName')]")
+              .type("Microsoft.Compute/virtualMachines")
+              .location(location)
+              .apiVersion("2015-06-15")
```
nacx: Use a shared a constant here too

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/DeploymentApiLiveTest.java
nacx: Hmmm I mean, in the live test, the actual body passed to the API method is the result of the `getPutBody ` method, but this method is only defined in the tests (which means that users will have to manually generate the json).

When I ask about the "abstracting the raw json template with a builder", I mean a builder that already produces the right body to be sent in the request (the `getPutBody`). I imagine something like `TempalteBuilder.foo().bar().mode(mode).parameters(parameters).build()` or `TempalteBuilder.foo().bar().build(mode, parameters)` that already produces what the `getPutBody` does in the tests, so users can actually forget about manually building json strings.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/DeploymentApiLiveTest.java
```diff
@@ -0,0 +1,129 @@
...
+      Deployment deployment = api().get(deploymentName);
+      assertNotNull(deployment);
+      ProvisioningState state = ProvisioningState.fromValue(deployment.properties().provisioningState());
+      assertTrue(state == ProvisioningState.SUCCEEDED);
+   }
+
+   @Test(groups = "live", dependsOnMethods = "testCreate")
+   public void testListDeployments() {
+      List<Deployment> deployments = api().list();
+      assertTrue(deployments.size() > 0 && deployments.get(0).name().equals(deploymentName));
```
nacx: If the live tests are run in an account that already has data, is it correct to expect the one used in the tests to be returned in the first position?

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/DeploymentApiLiveTest.java
```diff
@@ -0,0 +1,129 @@
...
+            URI uri = api().delete(d.name());
+            assertNotNull(uri);
+            assertTrue(uri.toString().contains("api-version"));
+            assertTrue(uri.toString().contains("operationresults"));
+
+            boolean jobDone = Predicates2.retry(new Predicate<URI>() {
+               @Override
+               public boolean apply(URI uri) {
+                  System.out.println(uri.toString());
+                  System.out.println(api.getJobApi().jobStatus(uri).name());
```
nacx: Remove this or use the anonymous logger

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/DeploymentApiLiveTest.java
```diff
@@ -0,0 +1,129 @@
...
+            }, 60 * maxTestDuration * 1000).apply(uri);
+            assertTrue(jobDone, "delete operation did not complete in the configured timeout");
+         }
+      }
+   }
+
+   private DeploymentApi api() {
+      return api.getDeploymentApi(resourceName);
+   }
+}
```
nacx: Add a test (or call it in an existing one) that exercises the `DeploymentApi.validate` method

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/DeploymentApiMockTest.java
```diff
@@ -0,0 +1,142 @@
...
+      String properties = getPutBody(template, "Incremental", parameters);
+      Deployment deployment = deploymentApi.create(deploymentName, properties);
+      assertTrue(deployment != null);
+      assertEquals(ProvisioningState.fromValue(deployment.properties().provisioningState()), ProvisioningState.ACCEPTED);
+
+      // check if deployment succeeded
+      server.enqueue(jsonResponse("/createdeploymentsucceeded.json"));
+      deployment = deploymentApi.create(deploymentName, properties);
+      assertTrue(deployment != null);
+      assertEquals(ProvisioningState.fromValue(deployment.properties().provisioningState()), ProvisioningState.SUCCEEDED);
```
nacx: Verify the recorded requests

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/DeploymentApiMockTest.java
```diff
@@ -0,0 +1,142 @@
...
+
+   @Test
+   public void testGetDeployment() throws Exception {
+      final DeploymentApi deploymentApi = api.getDeploymentApi(resourceGroup);
+
+      // check if deployment succeeded
+      server.enqueue(jsonResponse("/createdeploymentsucceeded.json"));
+      Deployment deployment = deploymentApi.get(deploymentName);
+      assertTrue(deployment != null);
+      assertEquals(ProvisioningState.fromValue(deployment.properties().provisioningState()), ProvisioningState.SUCCEEDED);
```
nacx: Verify the recorded request

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/DeploymentApiMockTest.java
```diff
@@ -0,0 +1,142 @@
...
+   }
+
+   @Test
+   public void testListDeployment() throws Exception {
+      final DeploymentApi deploymentApi = api.getDeploymentApi(resourceGroup);
+
+      // check if deployment succeeded
+      server.enqueue(jsonResponse("/listdeployments.json"));
+      List<Deployment> deployments = deploymentApi.list();
+      assertTrue(deployments.size() > 0);
```
nacx: Verify the recorded request

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/DeploymentApiMockTest.java
```diff
@@ -0,0 +1,142 @@
...
+   }
+
+   @Test
+   public void testListDeploymentEmpty() throws Exception {
+      final DeploymentApi deploymentApi = api.getDeploymentApi(resourceGroup);
+
+      server.enqueue(new MockResponse().setResponseCode(404));
+
+      List<Deployment> deployments = deploymentApi.list();
+      assertTrue(deployments.size() == 0);
```
nacx: Verify the recorded request

nacx:
>Ready for another round of review. The only outstanding question is how to deal with $schema.

Done reviewing. Regarding the `$schema` thing, AFAIK it works both ways. I've just isolated the case and put a very simple example to prove the behavior:

```java
@AutoValue
public abstract class DeploymentTemplate {
   public abstract String schema();
   public abstract String contentVersion();

   @SerializedNames({"$schema", "contentVersion"})
   public static DeploymentTemplate create(final String schema, final String contentVersion) {
      return new AutoValue_DeploymentTemplate(schema, contentVersion);
   }

   public static void main(final String[] args) throws Exception {
      Properties props = new Properties();
      props.setProperty("oauth.endpoint", "https://login.microsoftonline.com/<subscriptionId>/oauth2/token");
      
      // Just build the context to have access to the jclouds Utils class
      ApiContext<AzureComputeApi> ctx = ContextBuilder.newBuilder("azurecompute-arm")
         .credentials("<identity>", "<credential>")
         .overrides(props)
         .build();

      // Get the json instance with the Gson configuration used by jclouds to serialize/deserialize
      Json json = ctx.utils().json();

      // Verify serialization
      DeploymentTemplate t = DeploymentTemplate.create("http://localhost", "1.0.0");
      String toJson = json.toJson(t);
      assert toJson.equals("{\"$schema\":\"http://localhost\",\"contentVersion\":\"1.0.0\"}");

      // Verify deserialization
      DeploymentTemplate fromString = json.fromJson("{\"$schema\":\"http://localhost\",\"contentVersion\":\"1.0.0\"}", DeploymentTemplate.class);
      assert fromString.schema().equals("http://localhost");
   }
}
```

@isalento @jmspring Could you have a look?

ritazh:
@nacx I guess the issue with `$schema` (after testing it) is not with `SerializedNames`, but with builders. When I use the builder to pass in the `schema` parameter
```
DeploymentTemplate template = DeploymentTemplate.builder()
              .schema("https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#")
```
I get this error:
```
java.lang.IllegalArgumentException: {"error":{"code":"InvalidRequestContent","message":"The request content was invalid and could not be deserialized: 'Could not find member 'schema' on object of type 'Template'. Path 'properties.template.schema', line 1, position 36.'."}}
```

nacx:
Hmmmm The builders shouldn't affect, as the serialization will happen on the actual object (not the builder) and deserialization will use the static factory method. The same test case using the builder is working:

```java
@AutoValue
public abstract class DeploymentTemplate {
   public abstract String schema();
   public abstract String contentVersion();

   @SerializedNames({"$schema", "contentVersion"})
   public static DeploymentTemplate create(final String schema, final String contentVersion) {
      return DeploymentTemplate.builder().schema(schema).contentVersion(contentVersion).build();
   }
   
   public static Builder builder() {
      return new AutoValue_DeploymentTemplate.Builder();
   }
   
   @AutoValue.Builder
   public abstract static class Builder {
      public abstract Builder schema(String schema);
      public abstract Builder contentVersion(String type);
      
      public abstract DeploymentTemplate build();
   }

   public static void main(final String[] args) throws Exception {

      Properties props = new Properties();
      props.setProperty("oauth.endpoint", "https://login.microsoftonline.com/<subscriptionId>/oauth2/token");
      
      ApiContext<AzureComputeApi> ctx = ContextBuilder.newBuilder("azurecompute-arm")
            .credentials("<identity>", "<credential>")
            .overrides(props)
            .build();

      Json json = ctx.utils().json();

      DeploymentTemplate t = DeploymentTemplate.builder()
            .schema("http://localhost")
            .contentVersion("1.0.0")
            .build();
      String toJson = json.toJson(t);
      
      assert toJson.equals("{\"$schema\":\"http://localhost\",\"contentVersion\":\"1.0.0\"}");

      DeploymentTemplate fromString = json.fromJson("{\"$schema\":\"http://localhost\",\"contentVersion\":\"1.0.0\"}",
            DeploymentTemplate.class);

      assert fromString.schema().equals("http://localhost");
   }
}
```

Do you have an isolated test case I can use to reproduce the issue, in case I'm not getting what do you mean? (A link to a gist would be perfect).

ritazh:
@nacx Thank you for taking a look at this. It's probably something really obvious and I just missed it somehow. 
These are the implementation files:
https://gist.github.com/ritazh/0171333aaa69839763e88c50d59d28ed
https://gist.github.com/ritazh/1df20e75be3f604f4538b4fae217888c#file-deploymenttemplatebuilder-java-L109

This is the failing test:
https://gist.github.com/ritazh/3232ae0cbaa8cd2c7d932023cd04f2e9#file-templatetodeploymenttemplatelivetest-java-L127

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/DeploymentApiLiveTest.java
ritazh: I updated `DeploymentApiLiveTest` to use `DeploymentTemplateBuilder` now so it’s all hooked up.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/DeploymentApiLiveTest.java
ritazh: Done.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/DeploymentApiMockTest.java
ritazh: Done.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/DeploymentApiMockTest.java
ritazh: Done.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/DeploymentApiMockTest.java
ritazh: Done.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/DeploymentApiMockTest.java
ritazh: Done.

nacx:
Oh, I see it now. The issue is that you're building your custom Gson instance instead of using the jclouds one (configured in the [GsonModule](https://github.com/jclouds/jclouds/blob/master/core/src/main/java/org/jclouds/json/config/GsonModule.java) in jclouds-core). That instance you use does not have the jclouds serialization/deserialization strategies.

You need to use the Json instance provided by the injector, like in my example. Can you try changing the tests to use the jclouds Json instance?

ritazh:
@nacx That was exactly it! Thank you so much for your help!!! Got all the issues resolved now. Also squashed all the commits as one. Ready for another round of review.

nacx:
Do you have a link to just the last commit? I'll be able to review it quickly. Otherwise it will take a bit longer, as I need to allocate some time to check the whole PR taking special care to the changed bits :)

ritazh:
@nacx Sorry about that. Next time I will wait to squash. 

Here are the changes in the last commit:

Changes in `BaseAzureComputeApiLiveTest.java` to create context:

https://github.com/ritazh/jclouds-labs/blob/ed40df18648ddd8602ee3167bf491a05cb14df72/azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/internal/BaseAzureComputeApiLiveTest.java#L145-L155

https://github.com/ritazh/jclouds-labs/blob/ed40df18648ddd8602ee3167bf491a05cb14df72/azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/internal/BaseAzureComputeApiLiveTest.java#L72-L91

Changed all the schema fields in `DeplpoymentTemplate.java`:
https://github.com/ritazh/jclouds-labs/blob/ed40df18648ddd8602ee3167bf491a05cb14df72/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/DeploymentTemplate.java

https://github.com/ritazh/jclouds-labs/blob/ed40df18648ddd8602ee3167bf491a05cb14df72/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/util/DeploymentTemplateBuilder.java#L109

Calling `getContext()` in `TemplateToDeploymentTemplateLiveTest.java`:

https://github.com/ritazh/jclouds-labs/blob/ed40df18648ddd8602ee3167bf491a05cb14df72/azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/TemplateToDeploymentTemplateLiveTest.java#L58

Using `org.jclouds.json.Json` instead of `Gson` for serialization:

https://github.com/ritazh/jclouds-labs/blob/ed40df18648ddd8602ee3167bf491a05cb14df72/azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/TemplateToDeploymentTemplateLiveTest.java#L72

https://github.com/ritazh/jclouds-labs/blob/ed40df18648ddd8602ee3167bf491a05cb14df72/azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/TemplateToDeploymentTemplateLiveTest.java#L95

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/internal/BaseAzureComputeApiLiveTest.java
```diff
@@ -114,6 +142,17 @@ private void deleteResourceGroup(String name) {
...
+      if (context == null) {
+         Properties properties = new Properties();
+         properties.put("oauth.endpoint", "https://login.microsoftonline.com/tenant-id/oauth2/token");
+         context = ContextBuilder.newBuilder("azurecompute-arm")
+                 .credentials(getIdentity(), getCredential())
+                 .overrides(properties)
+                 .build();
+      }
+      return context;
+   }
```
nacx: Doing this would mean callers will be creating *another* context, just to access the injector. Instead, you'd better override the [BaseApiLiveTest#create](https://github.com/jclouds/jclouds/blob/master/core/src/test/java/org/jclouds/apis/BaseApiLiveTest.java#L141-L145) method, to create the context and assign it before returning the api.

nacx:
One last thing I think would be very helpful: with the current approach users need to manually get the `Json` instance from the context, but that is not always obious or easy. In stead, I propose to add a method to the `DeploymentTemplateBuilder`, something like `getDeploymentTemplateJson` that already returnes the json string. To do that, however, we need the Json instance injected in that class. Here it is how this can be achieved:

We'll use Guice Assisted Inject, to be able to have a constructor with both, injected and user provided parameters. It just requires us to define a factory interface and annotate the parameters. Something like:

```java
public class DeploymentTemplateBuilder {
   interface Factory {
      DeploymentTemplateBuilder create(String group, String name, Template template);
   }
   ...
   @Inject // Do not make the constructor public anymore. We will get this from the injector
   DeploymentTemplateBuilder(Json json, @Assisted String group, @Assisted String name, @Assisted Template template) {
   }
...
}
```

Then, we need to configure the Factory in the Guice context by adding the following to a Guice module:

```java
install(new FactoryModuleBuilder().build(DeploymentTemplateBuilder.Factory.class));
```

And finally, we need to expose the factory in some user-friendly way so users can get it easily. We could add the following to the AzureComputeApi class:

```java
@Provides
DeploymentTemplateBuilder.Factory deploymentTemplateFactory();
```

For methods annotated as `@Provides`, jclouds looks for the values in the Guice injector. Unfortunately if I'm not wrong, that is not yet supported in delegate APIs, so we can't put this method in the DeploymentApi.

Once everything is in place, users should be able to get the json for a deployment template like:

```java
DeploymentTemplateBuilder builder = azureApi.deploymentTemplateFactory().create(group, name, template);
String rawTemplate = builder.getDeploymentTemplateJson();
```

Which looks pretty clean and users don't have to deal with serialization and deserialization at all. WDYT?

ritazh:
@nacx Thank you very much for your help! DeploymentTemplateBuilder Factor has been added. Ready for another round of review. :) Almost there!

nacx:
Just two minors left. Mind squashing the commits directly while addressing them? Let's get this merged!

ritazh:
@nacx Fixed and sqashed the commits.

nacx:
Pushed to master as [fca7addf](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/fca7addf). Many thanks @ritazh!

PR #275, opened 2016-05-27
JCLOUDS-1124 oneandone-server-api

alibazlamit:
@baldwinSPC


alibazlamit:
fixing the issues reported

nacx:
Thanks for the contribution @alibazlamit!

Before we start a detailed review of the code, I think we should try to address some design. Having a quick look, it seems this PR is inspired by the ProfitBricks rest one, but I'm not sure that is the best approach given the nature of the one and one api.

Most of the code in this pull request are just request binders, but as far as I see the 1&1 api, its model classes are pretty straightforward and we could easily map all them to plain AutoValue model classes, and get the jclouds default json binders just serialize and deserialize everything without the need for custom binders. Is there an special need to justify all the amount of code (and maintenance overhead) that having a specific binder for each request would introduce?

It is really important to have the simpler implementation possible, and I'd strongly suggest you refactor the model to uso AutoValue to directly map the 1&1 json model, and let jclouds default binders take care of serialization.

alibazlamit:
@nacx Removed all of the unneeded binders, i had to keep a few binders where the Json/Objects are more complex, let me know if its looks ok now.

alibazlamit:
@nacx  Do i need to open another ticket or change something after pushing the code in order for the changes to be reviewed?

nacx:
Apologies for the lack of feedback; I've been quite busy these last days and haven't had cycles to review PRs. I'll review it later today!

alibazlamit:
@nacx no problem at all, I thought i needed to trigger some alert or somthing, thanks for the response.

nacx:
I really think we should get rid of *all* binders. Looking at the remaining ones, most of them seem to have mostly direct mappings from the Json objects to the Java ones. Instead of having binders to map that, we should have a model that matches the API. It is OK if some properties are not used; if the domain model properly reflects the API, then using it will be easier for users and it will 
be easier to contribute and maintain the provider too.

alibazlamit:
@nacx How can i keep the java standards for fields for example in the `CreateServerRequestBinder `we have `"firewallPolicyId"` that maps to `"firewall_policy_id"` from the API is there a way to make the AutoValue map it from Java standards to API property name?

nacx:
Yes. The order in which fields are declared in the class should match the order in which the parameters are declared in the static factory method that creates the auto-value class. That method can be annotated with the `@SerializedNames` annotation, to provide the list, in the same order than the method parameters, of the json property names. jclouds uses that to translate between java and json.

zack-shoylev:
> I really think we should get rid of all binders.

I second that. Binders are sometimes (very rarely) needed but should be always considered suspicious.

alibazlamit:
@nacx  [Example](https://github.com/StackPointCloud/jclouds-labs/blob/oneandone-server-api/oneandone/src/main/java/org/apache/jclouds/oneandone/rest/domain/Server.java#L182-L389) Just to make sure i understand you correctly you are saying that i need to get rid of all these Classes inside the Request class  in the example and make each one of them a separate class with the static factory method like this `@SerializedNames({"id", "country_code", "location"})
        public static DataCenter create(String id, String countryCode, String location) {
            return new AutoValue_Server_DataCenter(id, countryCode, location);
}` to provide Models for POST and PUT requests.

nacx:
They can be inner classes. Each one must have the annotated factory method, though.
Also consider if the create/update builder pattern makes sense here. If update operations are simple ones, such as changing the names and descriptions, then it makes sense to have the api method receive those two parameters instead of an update object, and define a `@Payload` annotation to generate the body directly.
Let's look for simplicity. Builders are great for complex objects, but let's try to have a simple model matching the provider api and simple api methods.

alibazlamit:
OK Thanks ill do the change.

alibazlamit:
@nacx Removed all binders and did a few changes to the Mock tests as well, please take a look and let me know.

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/config/Authentication.java
```diff
@@ -0,0 +1,29 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+import javax.inject.Qualifier;
+
+@Retention(value = RetentionPolicy.RUNTIME)
+@Target(value = {ElementType.TYPE, ElementType.FIELD, ElementType.PARAMETER, ElementType.METHOD})
+@Qualifier
+public @interface Authentication {
+}
```
nacx: Is this class actually used? If not, please remove it.

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/domain/Hardware.java
```diff
@@ -0,0 +1,71 @@
...
+
+    @SerializedNames({"vcore", "cores_per_processor", "ram", "hdds"})
+    public static Hardware create(double vcore, double coresPerProcessor, double ram, List<Hdd> hdds) {
+        return new AutoValue_Hardware(vcore, coresPerProcessor, ram, hdds);
+    }
+
+    @AutoValue
+    public abstract static class CreateHardware {
+
+        public abstract double vCore();
```
nacx: Just `vcore`, like in the `Hardware` object? Use a consistent naming.

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/domain/HardwareFlavour.java
```diff
@@ -0,0 +1,72 @@
...
+    @SerializedNames({"id", "name", "hardware"})
+    public static HardwareFlavour create(String id, String name, Hardware hardware) {
+        return new AutoValue_HardwareFlavour(id, name, hardware);
+    }
+
+    @AutoValue
+    public abstract static class Hardware {
+
+        @Nullable
+        public abstract String fixedInstanceSizeId();
```
nacx: If this is nullable, could we add it to the existing `Hardware` class and reuse it instead of having this inner one?

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/domain/HardwareFlavour.java
```diff
@@ -0,0 +1,72 @@
...
+        }
+
+        @AutoValue
+        public abstract static class Hdd {
+
+            public abstract String unit();
+
+            public abstract int size();
+
+            public abstract Boolean isMain();
```
nacx: Use a primitive value if this can never be null, otherwise annotate as `@Nullable`.

nacx: Apply the same criteria to all Boolean object values.

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/domain/Image.java
```diff
@@ -0,0 +1,85 @@
...
+
+    public abstract int numImages();
+
+    @Nullable
+
+    public abstract String creationDate();
+
+    @SerializedNames({"id", "name", "os_family", "os", "os_version", "availableSites", "architecture", "os_image_type", "type", "min_hdd_size", "licenses", "state", "description", "hdds", "server_id", "frequency", "numImages", "creation_date"})
+    public static Image create(String id, String name, OSFamliyType osFamily, OSType os, String osVersion, List<String> availableSites, int architecture, String osImageType, ImageType type, int minHddSize, List<Licenses> licenses, String state, String description, List<Hdd> hdds, String serverId, ImageFrequency frequency, int numImages, String creationDate) {
+        return new AutoValue_Image(id, name, osFamily, os, osVersion, availableSites, architecture, osImageType, type, minHddSize, licenses, state, description, hdds, serverId, frequency, numImages, creationDate);
```
nacx: Can we make lists non-nullable, by initialising them to an immutable empty list, if the input parameter is null? Or does the api expect that you do not send empty lists?

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/domain/Licenses.java
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+package org.apache.jclouds.oneandone.rest.domain;
+
+import com.google.auto.value.AutoValue;
+
+@AutoValue
+public abstract class Licenses {
+
+    public abstract String name();
```
nacx: Add the factory method.

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/domain/Server.java
```diff
@@ -0,0 +1,304 @@
...
+
+    @Nullable
+    public abstract ServerMonitoringPolicy monitoringPolicy();
+
+    @Nullable
+    public abstract List<ServerPrivateNetwork> privateNetworks();
+
+    @SerializedNames({"id", "name", "creation_date", "first_password", "description", "status", "hardware", "image", "dvd", "snapshot", "datacenter", "ips", "alerts", "monitoring_policy", "private_networks"})
+    public static Server create(String id, String name, Date creationDate, String firstPassword, String description, Status status, Hardware hardware, Image image, Dvd dvd, Snapshot snapshot, DataCenter datacenter, List<ServerIp> ips, List<Alert> alerts, ServerMonitoringPolicy policy, List<ServerPrivateNetwork> privateNetworks) {
+        return new AutoValue_Server(id, name, creationDate, firstPassword, description, status, hardware, image, dvd, snapshot, datacenter, ips, alerts, policy, privateNetworks);
```
nacx: Same comment about nullable lists.

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/domain/Snapshot.java
```diff
@@ -0,0 +1,35 @@
...
+import org.jclouds.json.SerializedNames;
+
+@AutoValue
+public abstract class Snapshot {
+
+    public abstract String id();
+
+    public abstract String creation_date();
+
+    public abstract String deletion_date();
```
nacx: Change to Date and camel case?

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/domain/Types.java
```diff
@@ -0,0 +1,220 @@
...
+        // the json node value with this enum
+        private final String value;
+
+        OSFamliyType(final String type) {
+            value = type;
+        }
+
+        @Override
+        public String toString() {
+            return value;
```
nacx: Do we need to actually "produce" the empty string as the enum value, when it is UNRECOGNIZED? If not, you can get rid of the value variable, the constructor and this method, as the default enum implementation will use the `name()` method.

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/domain/Types.java
```diff
@@ -0,0 +1,220 @@
...
+
+        private ArchitectureType(int id) {
+            this.id = id;
+        }
+
+        public int getId() {
+            return this.id;
+        }
+
+        public static Map<Integer, ArchitectureType> buildMap() {
```
nacx: Is this method actually used?

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/domain/Types.java
```diff
@@ -0,0 +1,220 @@
...
+        // the json node value with this enum
+        private final String value;
+
+        ImageType(final String type) {
+            value = type;
+        }
+
+        @Override
+        public String toString() {
+            return value;
```
nacx: Same comment than above.

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/domain/Types.java
```diff
@@ -0,0 +1,220 @@
...
+
+        IPType(final String type) {
+            value = type;
+        }
+
+        @Override
+        public String toString() {
+            return value;
+        }
+    }
```
nacx: Consider adding the `fromValue` to all enums and removing the unnecessary value when it is exactly the same as the enum name.

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/domain/WarningAlert.java
```diff
@@ -0,0 +1,35 @@
...
+import org.jclouds.json.SerializedNames;
+
+@AutoValue
+public abstract class WarningAlert {
+
+    public abstract String type();
+
+    public abstract String description();
+
+    public abstract String date();
```
nacx: Change type to Date?

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/features/ServerApi.java
```diff
@@ -0,0 +1,567 @@
...
+    @GET
+    @ResponseParser(ServerApi.ServerListParser.class)
+    @Fallback(Fallbacks.EmptyListOnNotFoundOr404.class)
+    List<Server> getList();
+
+    @Named("servers:list")
+    @GET
+    @ResponseParser(ServerApi.ServerListParser.class)
+    @Fallback(Fallbacks.EmptyListOnNotFoundOr404.class)
+    List<Server> getList(GenericQueryOptions options);
```
nacx: Rename these method to just `list`.

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/features/ServerApi.java
```diff
@@ -0,0 +1,567 @@
...
+    @Named("servers:create")
+    @POST
+    @ResponseParser(ServerApi.ServerParser.class)
+    Server createFixedInstanceServer(@BinderParam(BindToJsonPayload.class) CreateFixedInstanceServer server);
+
+    @Named("server:update")
+    @PUT
+    @Path("/{serverId}")
+    @ResponseParser(ServerApi.ServerParser.class)
+    @Produces("application/json")
```
nacx: Add this annotation to all methods that produce a body?

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/features/ServerApi.java
```diff
@@ -0,0 +1,567 @@
...
+import org.jclouds.rest.annotations.Fallback;
+import org.jclouds.rest.annotations.MapBinder;
+import org.jclouds.rest.annotations.PayloadParam;
+import org.jclouds.rest.annotations.RequestFilters;
+import org.jclouds.rest.annotations.ResponseParser;
+import org.jclouds.rest.binders.BindToJsonPayload;
+import org.jclouds.util.Strings2;
+
+@Path("servers")
+@RequestFilters(AuthenticateRequest.class)
```
nacx: Should a `@Consumes(APPLICATION_JSON)` annotation be added here, to send the Accept header by default?

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/features/ServerApi.java
```diff
@@ -0,0 +1,567 @@
...
+    @Path("/{serverId}/status/action")
+    @ResponseParser(ServerApi.ServerParser.class)
+    @Produces("application/json")
+    Server updateStatus(@PathParam("serverId") String serverId, @BinderParam(BindToJsonPayload.class) Server.UpdateStatus server);
+
+    @Named("server:delete")
+    @DELETE
+    @Path("/{serverId}")
+    @ResponseParser(ServerApi.ServerParser.class)
+    @MapBinder(BindToJsonPayload.class)
```
nacx: Remove this? There is no payload in this method.

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/features/ServerApi.java
```diff
@@ -0,0 +1,567 @@
...
+    @ResponseParser(ServerApi.HddParser.class)
+    @Fallback(Fallbacks.NullOnNotFoundOr404.class)
+    Hdd getHdd(@PathParam("serverId") String serverId, @PathParam("hddId") String hddId);
+
+    @Named("server:update")
+    @PUT
+    @Path("/{serverId}/hardware/hdds/{hddId}")
+    @MapBinder(BindToJsonPayload.class)
+    @ResponseParser(ServerApi.ServerParser.class)
+    @Fallback(Fallbacks.NullOnNotFoundOr404.class)
```
nacx: Don't add 404 fallbacks to post/put methods.

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/features/ServerApi.java
```diff
@@ -0,0 +1,567 @@
...
+    Server updateHdd(@PathParam("serverId") String serverId, @PathParam("hddId") String hddId, @PayloadParam("size") double size);
+
+    @Named("server:hdds/delete")
+    @DELETE
+    @Path("/{serverId}/hardware/hdds/{hddId}")
+    @ResponseParser(ServerApi.ServerParser.class)
+    @MapBinder(BindToJsonPayload.class)
```
nacx: There is no payload in this method. Remove this.

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/features/ServerApi.java
```diff
@@ -0,0 +1,567 @@
...
+    @Fallback(Fallbacks.EmptyListOnNotFoundOr404.class)
+    List<Server> getList();
+
+    @Named("servers:list")
+    @GET
+    @ResponseParser(ServerApi.ServerListParser.class)
+    @Fallback(Fallbacks.EmptyListOnNotFoundOr404.class)
+    List<Server> getList(GenericQueryOptions options);
+
+    @Named("servers:Flavours:list")
```
nacx: Why Flavours with a capital letter? Follow the same naming in all methods.

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/features/ServerApi.java
```diff
@@ -0,0 +1,567 @@
...
+
+    @Named("server:update")
+    @PUT
+    @Path("/{serverId}/hardware/hdds/{hddId}")
+    @MapBinder(BindToJsonPayload.class)
+    @ResponseParser(ServerApi.ServerParser.class)
+    @Fallback(Fallbacks.NullOnNotFoundOr404.class)
+    Server updateHdd(@PathParam("serverId") String serverId, @PathParam("hddId") String hddId, @PayloadParam("size") double size);
+
+    @Named("server:hdds/delete")
```
nacx: Use the same naming convention in all methods.

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/features/ServerApi.java
```diff
@@ -0,0 +1,567 @@
...
+    @Path("/{serverId}/image")
+    @ResponseParser(ServerApi.ImageParser.class)
+    @Fallback(Fallbacks.NullOnNotFoundOr404.class)
+    Image getImage(@PathParam("serverId") String serverId);
+
+    @Named("server:image:update")
+    @PUT
+    @Path("/{serverId}/image")
+    @ResponseParser(ServerApi.UpdateServerParser.class)
+    @Fallback(Fallbacks.NullOnNotFoundOr404.class)
```
nacx: Remove fallback.

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/features/ServerApi.java
```diff
@@ -0,0 +1,567 @@
...
+    @Path("/{serverId}/ips/{ipId}")
+    @ResponseParser(ServerApi.ServerIpParser.class)
+    @Fallback(Fallbacks.NullOnNotFoundOr404.class)
+    ServerIp getIp(@PathParam("serverId") String serverId, @PathParam("ipId") String ipId);
+
+    @Named("server:ip:delete")
+    @DELETE
+    @Path("/{serverId}/ips/{ipId}")
+    @ResponseParser(ServerApi.ServerParser.class)
+    @MapBinder(BindToJsonPayload.class)
```
nacx: Remove binder.

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/features/ServerApi.java
```diff
@@ -0,0 +1,567 @@
...
+    @Fallback(Fallbacks.NullOnNotFoundOr404.class)
+    Server deleteSnapshot(@PathParam("serverId") String serverId, @PathParam("snapshotId") String snapshotId);
+
+    @Named("servers:clone:create")
+    @POST
+    @Path("/{serverId}/clone")
+    @ResponseParser(ServerApi.ServerParser.class)
+    Server clone(@PathParam("serverId") String serverId, @BinderParam(BindToJsonPayload.class) Server.Clone clone);
+
+    static final class ServerListParser extends ParseJson<List<Server>> {
```
nacx: jclouds should be able to deserialize these objects out of the box. Can all the parsers in this class be removed, and the corresponding response parser annotations in the api methods?

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/filters/AuthenticateRequest.java
```diff
@@ -0,0 +1,53 @@
...
+import static com.google.common.base.Preconditions.checkNotNull;
+import org.jclouds.domain.Credentials;
+
+@Singleton
+public class AuthenticateRequest implements HttpRequestFilter {
+
+    private static Credentials authToken;
+
+    @Inject
+    public AuthenticateRequest(@Provider Supplier<Credentials> splr) {
```
nacx: Remove the public modifier. Injection constructors should be package private.

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/filters/AuthenticateRequest.java
```diff
@@ -0,0 +1,53 @@
...
+import org.jclouds.http.HttpRequest;
+import org.jclouds.http.HttpRequestFilter;
+import org.jclouds.location.Provider;
+import static com.google.common.base.Preconditions.checkNotNull;
+import org.jclouds.domain.Credentials;
+
+@Singleton
+public class AuthenticateRequest implements HttpRequestFilter {
+
+    private static Credentials authToken;
```
nacx: Remove the static modifier and make it final.

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/filters/AuthenticateRequest.java
```diff
@@ -0,0 +1,53 @@
...
+        authToken = splr.get();
+    }
+
+    public static String TokenBased() {
+        return authToken.identity;
+    }
+
+    public static String TokenBased(String auth_token) {
+        return auth_token;
+    }
```
nacx: Remove these methods

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/filters/AuthenticateRequest.java
```diff
@@ -0,0 +1,53 @@
...
+
+    @Override
+    public HttpRequest filter(HttpRequest request) throws HttpException {
+        checkNotNull(authToken.identity, "credential returned null");
```
nacx: Move this to the constructor.

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/handlers/OneAndOneHttpErrorHandler.java
```diff
@@ -0,0 +1,67 @@
...
+               break;
+            case 413:
+            case 503:
+               // if nothing (default message was OK) was parsed from command executor, assume it was an 503 (Maintenance) html response.
+               if (response.getMessage().equals("OK"))
+                  exception = new HttpResponseException("The OneAndOne team is currently carrying out maintenance.", command, response);
+               else
+                  exception = new InsufficientResourcesException(response.getMessage(), exception);
+               break;
+         }
```
nacx: An exception must be set y default, otherwise we could be setting a null one if none of the codes above is returned. Let's better cover that case.

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/util/ArrayAdapter.java
```diff
@@ -0,0 +1,149 @@
...
+import com.google.inject.TypeLiteral;
+import java.io.IOException;
+import java.lang.reflect.Type;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import org.apache.jclouds.oneandone.rest.domain.ServerFirewallPolicy;
+import org.apache.jclouds.oneandone.rest.domain.Snapshot;
+
+public class ArrayAdapter<T> extends TypeAdapter<List<T>> {
```
nacx: Caan you give an example of what this is trying to deserialize? Is this class really needed?

oneandone/src/test/java/org/apache/jclouds/oneandone/rest/features/ServerApiLiveTest.java
```diff
@@ -0,0 +1,269 @@
...
+            assertNodeAvailable(currentServer);
+            deleteServer(currentServer.id());
+        }
+    }
+
+    @Test(dependsOnMethods = "testListHardwareFlavours")
+    public void testCreateFixedInstanceServer() {
+        fixedInstanceServer = serverApi().createFixedInstanceServer(Server.CreateFixedInstanceServer.create(
+                "java test fixed instance 2", "testing with jclouds", FixedInstanceHardware.create(currentFlavour.id()), "7C5FA1D21B98DE39D7516333AAB7DA54",
+                null, "Test123!", null, Boolean.TRUE, null, null, null, null));
```
nacx: For constructors taking so many values, consider adding an auto-value builder to the model class.

oneandone/src/test/java/org/apache/jclouds/oneandone/rest/features/ServerApiLiveTest.java
```diff
@@ -0,0 +1,269 @@
...
+    }
+
+    @Test
+    public void testList() {
+
+        servers = serverApi().getList();
+
+        assertNotNull(servers);
+        assertFalse(servers.isEmpty());
+        Assert.assertTrue(servers.size() > 0);
```
nacx: This is redundant.

oneandone/src/test/java/org/apache/jclouds/oneandone/rest/features/ServerApiLiveTest.java
```diff
@@ -0,0 +1,269 @@
...
+
+    }
+
+    @Test
+    public void testList() {
+
+        servers = serverApi().getList();
+
+        assertNotNull(servers);
+        assertFalse(servers.isEmpty());
```
nacx: Better verify that the server the test created exists. Otherwise this could lead to false positives if running the tests on an account with existing servers.

oneandone/src/test/java/org/apache/jclouds/oneandone/rest/features/ServerApiLiveTest.java
```diff
@@ -0,0 +1,269 @@
...
+
+    @Test(dependsOnMethods = "testList")
+    public void testListWithOption() {
+        GenericQueryOptions options = new GenericQueryOptions();
+        options.options(0, 0, null, "test", null);
+        List<Server> serversWithQuery = serverApi().getList(options);
+
+        assertNotNull(serversWithQuery);
+        assertFalse(serversWithQuery.isEmpty());
+        Assert.assertTrue(serversWithQuery.size() > 0);
```
nacx: Remove this redundant assertion and all similar ones.

oneandone/src/test/java/org/apache/jclouds/oneandone/rest/features/ServerApiLiveTest.java
```diff
@@ -0,0 +1,269 @@
...
+        Server updateResult = serverApi().updateHardware(currentServer.id(), Hardware.UpdateHardware.create(4, 2, 6));
+
+        assertNotNull(updateResult);
+    }
+
+    @Test(dependsOnMethods = "testAddHdds")
+    public void testListHardwareHdds() throws InterruptedException {
+        assertNodeAvailable(currentServer);
+        //give time for harddisk to be added
+        Thread.sleep(60000);
```
nacx: Is there a smarter way to do this?

oneandone/src/test/java/org/apache/jclouds/oneandone/rest/features/ServerApiMockTest.java
```diff
@@ -0,0 +1,702 @@
...
+        GenericQueryOptions options = new GenericQueryOptions();
+        options.options(0, 0, null, "test", null);
+        List<Server> servers = serverApi().getList(options);
+
+        assertNotNull(servers);
+        assertEquals(servers.size(), 9);
+
+        assertEquals(server.getRequestCount(), 1);
+        assertSent(server, "GET", "/servers?q=test");
+    }
```
nacx: Added test for fallback in the list with options method.

oneandone/src/test/java/org/apache/jclouds/oneandone/rest/features/ServerApiMockTest.java
```diff
@@ -0,0 +1,702 @@
...
+    public void testDeleteServer() throws InterruptedException {
+        server.enqueue(
+                new MockResponse().setBody(stringFromResource("/server/delete.json"))
+        );
+        Server response = serverApi().delete("serverId");
+
+        assertNotNull(response);
+        assertEquals(server.getRequestCount(), 1);
+        assertSent(server, "DELETE", "/servers/serverId");
+    }
```
nacx: Add missing mock tests (specially the fallback ones).

oneandone/src/test/java/org/apache/jclouds/oneandone/rest/features/ServerNetworkApiLiveTest.java
```diff
@@ -0,0 +1,142 @@
...
+        assertNotNull(response);
+    }
+
+    @Test(dependsOnMethods = "testAddIp")
+    public void testListips() throws InterruptedException {
+        List<ServerIp> ips = serverApi().listIps(currentServer.id());
+        currentIP = ips.get(0);
+        assertNotNull(ips);
+        assertFalse(ips.isEmpty());
+        Assert.assertTrue(ips.size() > 0);
```
nacx: Remove redundant assertions.

oneandone/src/test/java/org/apache/jclouds/oneandone/rest/features/ServerOperationsApiLiveTest.java
```diff
@@ -0,0 +1,171 @@
...
+
+    @BeforeClass
+    public void setupTest() {
+        currentServer = createServer("jclouds operations test");
+    }
+
+    @AfterClass(alwaysRun = true)
+    public void teardownTest() throws InterruptedException {
+        //give time for operations to finish
+        Thread.sleep(10000);
```
nacx: Is it possible to get rid of this?

oneandone/src/test/java/org/apache/jclouds/oneandone/rest/internal/BaseOneAndOneApiMockTest.java
```diff
@@ -0,0 +1,122 @@
...
+            throws InterruptedException {
+        RecordedRequest request = assertSent(server, method, path);
+
+        String expectedContentType = "application/json";
+
+        assertEquals(request.getHeader("Content-Type"), expectedContentType);
+        assertEquals(parser.parse(new String(request.getBody(), Charsets.UTF_8)), parser.parse(json));
+        return request;
+    }
+}
```
nacx: Remove unused methods.

oneandone/src/test/java/org/apache/jclouds/oneandone/rest/internal/BaseOneAndOneLiveTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+                if ((server.status().state() != Types.ServerState.POWERED_OFF
+                        && server.status().state() != Types.ServerState.POWERED_ON)
+                        || server.status().percent() != 0) {
+                    return false;
+                } else {
+                    return true;
+                }
+            }
+        };
```
nacx: Configure this in a reusable predicate in the HttpApiModule. Take [this](https://github.com/jclouds/jclouds/blob/master/providers/digitalocean2/src/main/java/org/jclouds/digitalocean2/compute/config/DigitalOcean2ComputeServiceContextModule.java#L100-L106) as an example.

oneandone/src/test/java/org/apache/jclouds/oneandone/rest/internal/BaseOneAndOneLiveTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+    protected Server createServer(String serverName) {
+
+        List<Hdd.CreateHdd> hdds = new ArrayList<Hdd.CreateHdd>();
+        Hdd.CreateHdd hdd = Hdd.CreateHdd.create(30, Boolean.TRUE);
+        hdds.add(hdd);
+        Hardware.CreateHardware hardware = Hardware.CreateHardware.create(4.0, 1.0, 2.0, hdds);
+
+        return api.serverApi().create(Server.CreateServer.create(serverName, "testing with jclouds", hardware, "81504C620D98BCEBAA5202D145203B4B", "908DC2072407C94C8054610AD5A53B8C",
+                "Test123!", null, Boolean.TRUE, null, null, null, null)
```
nacx: Consider using a builder here.

nacx:
Thanks @alibazlamit! It is now a pretty clean api! This is a really nice start. Great job!

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/domain/HardwareFlavour.java
alibazlamit: The Hdd object is different too

```

public abstract static class Hdd {
            public abstract String unit();
            public abstract int size();
            public abstract Boolean isMain();
```

And at the Hardware class

```

public abstract class Hdd {

    public abstract String id();

    public abstract double size();

    public abstract Boolean isMain();
```

I can mix them into one but i think this is neater, what do you say?

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/domain/Image.java
alibazlamit: The API does not require empty lists do i need to make this change?

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/features/ServerApi.java
alibazlamit: I did not have the need for that my guess is that the Response parser does a similar job

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/features/ServerApi.java
alibazlamit: The API asks for a content-type:applicationjson for delete methods and this does the job

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/domain/HardwareFlavour.java
nacx: Oh, right. Let's keep it as-is then!

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/domain/Image.java
nacx: In general it is a best practice to avoid having null collections, and helps a lot avoiding NPEs.

There is no problem in returning empty ones by default if the api does not return a value for them. But when sending requests, some apis require the values to be skipped from the json instead of sending an empty array. If empty arrays are ok in this api, then we'd better avoid having null collections and always initialize them to empty in the factory methods when the value is null.

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/features/ServerApi.java
nacx: Then replace it by a `@Produces` annotation. It is the way to explicitly configure the value of the content-type header.

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/features/ServerApi.java
alibazlamit: I get this error when removing the response parser and run the tests
`You must specify a ResponseParser annotation`
any other annotation i can use that would do the job

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/util/ArrayAdapter.java
alibazlamit: The SnapshotListParser uses this, the API return a single object if there is only one item, if there are many snapshots it would return a list so this Adapter solved this confusion by the API

oneandone/src/test/java/org/apache/jclouds/oneandone/rest/internal/BaseOneAndOneLiveTest.java
alibazlamit: Is there an example for this in the HttpApiModule this is in the ComputeContextModule, maybe leave it here until we get to the compute and then added it to the ContextModule?

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/features/ServerApi.java
nacx: If you add this annotation, jclouds will automatically use its built-in json parser to parse responses.

oneandone/src/test/java/org/apache/jclouds/oneandone/rest/internal/BaseOneAndOneLiveTest.java
nacx: No worries, it is OK to have it in the HttpApi module instead of in the compute one.

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/features/ServerApi.java
nacx: And that would also allow you to remove all the response parsers, except the array one.

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/features/ServerApi.java
alibazlamit: This did work for the DELETE method it didn't generate he content-type header for me
My code example 
```
@Named("servers:privatenetwork:delete")
    @DELETE
    @Path("/{serverId}/private_networks/{privateNetworkId}")
    @Fallback(Fallbacks.NullOnNotFoundOr404.class)
    @Produces("application/json")
    Server deletePrivateNetwork(@PathParam("serverId") String serverId, @PathParam("privateNetworkId") String privateNetworkId);
```

this is what the header looks like :

`User-Agent: jclouds-okhttp/2.0.0-SNAPSHOT java/1.8.0_73
Accept: application/json
X-TOKEN: 4f34bcc24bf7bbf6af2fac5e35e600d8
Content-Length: 0
Host: cloudpanel-api.1and1.com
Connection: Keep-Alive
Accept-Encoding: gzip
`

alibazlamit:
I have pushed all the changes Could you please take a look at the OneAndOneHttpApiModule and let me know if that's how its supposed to be.

Thanks

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/features/ServerApi.java
nacx: Oh, ok, then leave delete methods this way.

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/config/OneAndOneHttpApiModule.java
```diff
@@ -0,0 +1,116 @@
...
+        @Override
+        public boolean apply(ServerPrivateNetworkRef networkRef) {
+            checkNotNull(networkRef, "ServerPrivateNetworkRef");
+            PrivateNetwork server = api.serverApi().getPrivateNetwork(networkRef.serverId(), networkRef.privateNetworkId());
+
+            if (server.state() != Types.GenericState.ACTIVE) {
+                return false;
+            } else {
+                return true;
+            }
```
nacx: Just `return server.state() == Types.GenericState.ACTIVE`? Apply this to the other similar conditionals.

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/filters/AuthenticateRequest.java
```diff
@@ -0,0 +1,46 @@
...
+import org.jclouds.http.HttpRequestFilter;
+import org.jclouds.location.Provider;
+
+@Singleton
+public class AuthenticateRequest implements HttpRequestFilter {
+
+    private final Credentials authToken;
+
+    @Inject
+    private AuthenticateRequest(@Provider Supplier<Credentials> splr) {
```
nacx: Package private visibility [is preferred](https://github.com/google/guice/wiki/KeepConstructorsHidden) for injection constructors. Change it to just:
```java
@Inject
AuthenticateRequest(@Provider Supplier<Credentials> splr) {
```

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/features/ServerApi.java
```diff
@@ -0,0 +1,370 @@
...
+
+        @Inject
+        FirewallPolicyListParser(Json json) {
+            super(json, list);
+        }
+
+        @Override
+        public <V> V apply(InputStream stream, Type type) throws IOException {
+            try {
+                Gson gson = new GsonBuilder().registerTypeAdapterFactory(new ArrayAdapterFactory()).create();
```
nacx: Since the ArrayAdapterFactory is now only used for these two concrete lists, would it make sense to just have two adapters, one specific for each list and remove the generic logic?

nacx:
This looks great @alibazlamit! Just a couple minors left and a comment about the adapter!

alibazlamit:
Changes Updated

nacx:
Thanks for the change @alibazlamit! One last thing: can you format the code to use 3 space indentation, according to our [style guide](https://cwiki.apache.org/confluence/display/JCLOUDS/Coding+Standards)? No need for another review round, if it works for you, just squash all the commits into a single one and apply the formatting change directly so we can merge it without delaying it more :)

A side note about the adapters (feel free to apply the change or not): would it make sense to create the Gson instance in the constructor, so it hasn't to be created every time an objects needs to be deserialised?

alibazlamit:
Squashed Thanks @nacx moving on to the next branch.

nacx:
Pushed to master as [7df28d25](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/7df28d25). Thanks for the continued effort @alibazlamit!

PR #277, opened 2016-06-01
Use container host ip not manager ip for populating node metadata.

duncangrant:
When providing node ip in metadata jclouds was using the
ip of the docker manager which was fine for docker machine
but fails when using a swarm cluster.  This uses the ip returned
from docker.

Currently the live tests for docker pass against a machine provisioning using docker/docker-machine. However they fail when run against a swarm cluster.  This is because the wrong ip address is used to test ssh access to the node.  This change will make the tests in SshToCustomPortLiveTest.java pass.

I haven't set up a docker overlay network so I don't know if the tests in NetworkApiLiveTest.java pass but all other live tests pass against swarm.


grkvlt:
@andreaturli this is another PR required for Docker Swarm compatibility. The change is pretty simple, and should also be easy to port to master, I think. Can you take a look, please?

andreaturli:
Lgtm. I wonder how many differences/gaps there are between docker engine API and dovker swarm API. Should we consider a jclouds-docker-swarm provider based on the jclouds-docker(-engine) api or simply extend the jclouds-docker provider? @gkrvlt and @nacx, any thoughts on this?

grkvlt:
@andreaturli It's not that Swarm has a _different_ API, it's just that using Swarm instead of Docker Engine as the target API endpoint surfaces different aspects of the JSON requests and responses, which we didn't need most of the time for plain Docker usage scenarios. Docker are committed to keeping the APIs compatible at the moment, although there is a proposal open to add a few extra calls here https://github.com/docker/swarm/issues/2134

andreaturli:
lgtm, merging

andreaturli:
thx @duncangrant, merged at [master](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/7e55ad79)

do you need to back-port this patch to 1.9.x branch as well? I think there are conflicts when trying to cherry-pick the same commit.

grkvlt:
@andreaturli I think the only conflict is that `DockerTestUtils.java` doesn't exist in 1.9.x yet, everything else should merge OK. I'm happy for the `DockerTestUtils` change to be ignored here.

andreaturli:
ok backported at [1.9.x](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/2033e639)

grkvlt:
Thanks @andreaturli @duncangrant

PR #279, opened 2016-06-03
JCLOUDS-664 Azurecompute-arm compute service

ritazh:
@nacx @andreaturli 
This PR is for Azure ARM Compute Service, part of the foundations needed for interacting with Azure Resource Manager (ARM) Rest APIs.

FYI
@jtjk @isalento @jmspring

### DIFF
--- a/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
+++ b/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
@@ -1,0 +1,393 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jclouds.azurecompute.arm.compute;
+
+import static java.lang.String.format;
+import static java.util.concurrent.TimeUnit.SECONDS;
+import static org.jclouds.util.Predicates2.retry;
+import java.util.ArrayList;
+
+import java.util.Collection;
+
+import java.util.List;
+import java.util.Set;
+
+import javax.annotation.Resource;
+import javax.inject.Inject;
+import javax.inject.Named;
+import javax.inject.Singleton;
+
+import com.google.common.collect.ArrayListMultimap;
+import com.google.common.collect.Multimap;
+import com.google.common.net.UrlEscapers;
+import org.jclouds.azurecompute.arm.AzureComputeApi;
+import org.jclouds.azurecompute.arm.compute.config.AzureComputeServiceContextModule.AzureComputeConstants;
+import org.jclouds.azurecompute.arm.compute.functions.VMImageToImage;
+import org.jclouds.azurecompute.arm.domain.Deployment;
+import org.jclouds.azurecompute.arm.domain.DeploymentBody;
+import org.jclouds.azurecompute.arm.domain.DeploymentProperties;
+import org.jclouds.azurecompute.arm.domain.VMImage;
+import org.jclouds.azurecompute.arm.domain.VMHardware;
+import org.jclouds.azurecompute.arm.domain.Location;
+import org.jclouds.azurecompute.arm.domain.Offer;
+import org.jclouds.azurecompute.arm.domain.PublicIPAddress;
+import org.jclouds.azurecompute.arm.domain.ResourceProviderMetaData;
+import org.jclouds.azurecompute.arm.domain.SKU;
+import org.jclouds.azurecompute.arm.domain.VMDeployment;
+import org.jclouds.azurecompute.arm.domain.VMSize;
+import org.jclouds.azurecompute.arm.domain.VirtualMachine;
+import org.jclouds.azurecompute.arm.features.DeploymentApi;
+import org.jclouds.azurecompute.arm.features.OSImageApi;
+import org.jclouds.azurecompute.arm.features.VirtualMachineApi;
+import org.jclouds.azurecompute.arm.util.DeploymentTemplateBuilder;
+import org.jclouds.compute.ComputeServiceAdapter;
+import org.jclouds.compute.domain.Template;
+import org.jclouds.compute.reference.ComputeServiceConstants;
+import org.jclouds.domain.LoginCredentials;
+import org.jclouds.logging.Logger;
+import org.jclouds.azurecompute.arm.functions.CleanupResources;
+
+import com.google.common.base.Predicate;
+import com.google.common.collect.Iterables;
+import com.google.common.collect.Lists;
+import com.google.common.collect.Sets;
+import com.google.common.base.Splitter;
+import static com.google.common.base.Preconditions.checkState;
+
+/**
+ * Defines the connection between the {@link AzureComputeApi} implementation and the jclouds
+ * {@link org.jclouds.compute.ComputeService}.
+ */
+@Singleton
+public class AzureComputeServiceAdapter implements ComputeServiceAdapter<VMDeployment, VMHardware, VMImage, Location> {
+
+   private String azureGroup;
+   protected final CleanupResources cleanupResources;
+
+   @Resource
+   @Named(ComputeServiceConstants.COMPUTE_LOGGER)
+   private Logger logger = Logger.NULL;
+
+   private final AzureComputeApi api;
+
+   private final AzureComputeConstants azureComputeConstants;
+
+   @Inject
+   AzureComputeServiceAdapter(final AzureComputeApi api, final AzureComputeConstants azureComputeConstants,
+                              CleanupResources cleanupResources) {
+
+      this.api = api;
+      this.azureComputeConstants = azureComputeConstants;
+
+      this.azureGroup = azureComputeConstants.azureResourceGroup();
+
+      logger.debug("AzureComputeServiceAdapter set azuregroup to: " + azureGroup);
+
+      this.cleanupResources = cleanupResources;
+   }
+
+   @Override
+   public NodeAndInitialCredentials<VMDeployment> createNodeWithGroupEncodedIntoName(
+           final String group, final String name, final Template template) {
+
+      DeploymentTemplateBuilder deploymentTemplateBuilder = api.deploymentTemplateFactory().create(group, name, template);
+
+      final String loginUser = DeploymentTemplateBuilder.getLoginUserUsername();
+      final String loginPassword = DeploymentTemplateBuilder.getLoginPassword();
+
+      DeploymentBody deploymentTemplateBody =  deploymentTemplateBuilder.getDeploymentTemplate();
+
+      DeploymentProperties properties = DeploymentProperties.create(deploymentTemplateBody);
+
+      final String deploymentTemplate = UrlEscapers.urlFormParameterEscaper().escape(deploymentTemplateBuilder.getDeploymentTemplateJson(properties));
+
+      logger.debug("Deployment created with name: %s group: %s", name, group);
+
+
+
+      final Set<VMDeployment> deployments = Sets.newHashSet();
+
+      final DeploymentApi deploymentApi = api.getDeploymentApi(group);
+
+      if (!retry(new Predicate<String>() {
+         @Override
+         public boolean apply(final String name) {
+
+            Deployment deployment = deploymentApi.create(name, deploymentTemplate);
+
+            if (deployment != null) {
+               VMDeployment vmDeployment = new VMDeployment();
+               vmDeployment.deployment = deployment;
+               deployments.add(vmDeployment);
+            } else {
+               logger.debug("Failed to create deployment!");
+            }
+            return !deployments.isEmpty();
+         }
+      }, azureComputeConstants.operationTimeout(), 1, SECONDS).apply(name)) {
+         final String illegalStateExceptionMessage = format("Deployment %s was not created within %sms so it will be destroyed.",
+                 name, azureComputeConstants.operationTimeout());
+         logger.warn(illegalStateExceptionMessage);
+         destroyNode(name);
+         throw new IllegalStateException(illegalStateExceptionMessage);
+      }
+
+      final VMDeployment deployment = deployments.iterator().next();
+
+
+      return new NodeAndInitialCredentials<VMDeployment>(deployment, name,
+              LoginCredentials.builder().user(loginUser).identity(loginUser).password(loginPassword).authenticateSudo(true).build());
+   }
+
+   @Override
+   public Iterable<VMHardware> listHardwareProfiles() {
+
+      final List<VMHardware> hwProfiles = Lists.newArrayList();
+      final List<String> locationIds = Lists.newArrayList();
+
+      Iterable<Location> locations = listLocations();
+      for (Location location : locations){
+         locationIds.add(location.name());
+
+         Iterable<VMSize> vmSizes = api.getVMSizeApi(location.name()).list();
+
+         for (VMSize vmSize : vmSizes){
+            VMHardware hwProfile = new VMHardware();
+            hwProfile.name = vmSize.name();
+            hwProfile.numberOfCores = vmSize.numberOfCores();
+            hwProfile.osDiskSizeInMB = vmSize.osDiskSizeInMB();
+            hwProfile.resourceDiskSizeInMB = vmSize.resourceDiskSizeInMB();
+            hwProfile.memoryInMB = vmSize.memoryInMB();
+            hwProfile.maxDataDiskCount = vmSize.maxDataDiskCount();
+            hwProfile.location = location.name();
+            hwProfiles.add(hwProfile);
+         }
+
+      }
+
+      checkAndSetHwAvailability(hwProfiles, Sets.newHashSet(locationIds));
+
+      return hwProfiles;
+   }
+   private void checkAndSetHwAvailability(List<VMHardware> hwProfiles, Collection<String> locations) {
+      Multimap<String, String> hwMap = ArrayListMultimap.create();
+      for (VMHardware hw : hwProfiles) {
+         hwMap.put(hw.name, hw.location);
+      }
+
+      for (VMHardware hw : hwProfiles) {
+         hw.globallyAvailable = hwMap.get(hw.name).containsAll(locations);
+      }
+   }
+
+   private void getImagesFromPublisher(String publisherName, List<VMImage> osImagesRef, String location) {
+
+      OSImageApi osImageApi = api.getOSImageApi(location);
+
+      Iterable<Offer> offerList = osImageApi.listOffers(publisherName);
+
+      for (Offer offer : offerList) {
+         Iterable<SKU> skuList = osImageApi.listSKUs(publisherName, offer.name());
+
+         for (SKU sku : skuList) {
+            VMImage vmImage = new VMImage();
+            vmImage.publisher = publisherName;
+            vmImage.offer = offer.name();
+            vmImage.sku = sku.name();
+            vmImage.location = location;
+            osImagesRef.add(vmImage);
+         }
+      }
+   }
+
+   private List<VMImage> listImagesByLocation(String location) {
+      final List<VMImage> osImages = Lists.newArrayList();
+      Iterable<String> publishers = Splitter.on(',').trimResults().omitEmptyStrings().split(this.azureComputeConstants.azureImagePublishers());
+      for (String publisher : publishers) {
+         getImagesFromPublisher(publisher, osImages, location);
+      }
+      return osImages;
+   }
+
+   @Override
+   public Iterable<VMImage> listImages() {
+
+      final List<VMImage> osImages = Lists.newArrayList();
+      final List<String> locationIds = Lists.newArrayList();
+
+      for (Location location : listLocations()){
+         locationIds.add(location.name());
+         osImages.addAll(listImagesByLocation(location.name()));
+      }
+      checkAndSetImageAvailability(osImages, Sets.newHashSet(locationIds));
+      return osImages;
+   }
+
+   private void checkAndSetImageAvailability(List<VMImage> images, Collection<String> locations) {
+      Multimap<String, String> map = ArrayListMultimap.create();
+
+      for (VMImage image : images) {
+         map.put( image.offer + "/" + image.sku, image.location);
+      }
+
+      for (VMImage image : images) {
+         image.globallyAvailable = map.get(image.offer + "/" + image.sku).containsAll(locations);
+      }
+   }
+
+   @Override
+   public VMImage getImage(final String id) {
+      String[] fields = VMImageToImage.decodeFieldsFromUniqueId(id);
+
+      Iterable<VMImage> images = listImages();
+
+      for (VMImage image : images) {
+         String imageId = VMImageToImage.encodeFieldsToUniqueId(image);
+         if (id.equals(imageId)){
+            return image;
+         }
+      }
+      return null;
+   }
+
+   @Override
+   public Iterable<Location> listLocations() {
+      List<Location> locations = api.getLocationApi().list();
+
+      List<ResourceProviderMetaData> resources = api.getResourceProviderApi().get("Microsoft.Compute");
+
+      final List<String> vmLocations = new ArrayList<String>();
+
+      for (ResourceProviderMetaData m : resources){
+         if (m.resourceType().equals("virtualMachines")){
+            vmLocations.addAll(m.locations());
+            break;
+         }
+      }
+
+      Iterable<Location> result = Iterables.filter(locations, new Predicate<Location>() {
+         @Override
+         public boolean apply(Location input) {
+            return vmLocations.contains(input.displayName());
+         }
+      });
+
+      return  result;
+   }
+
+   private String getResourceGroupFromId(String id) {
+      String searchStr = "/resourceGroups/";
+      int indexStart = id.lastIndexOf(searchStr) + searchStr.length();
+      searchStr = "/providers/";
+      int indexEnd = id.lastIndexOf(searchStr);
+
+      String resourceGroup = id.substring(indexStart, indexEnd);
+      return resourceGroup;
+   }
+
+   @Override
+   public VMDeployment getNode(final String id) {
+      Deployment deployment = api.getDeploymentApi(azureGroup).get(id);
+      if (deployment == null)
+         return null;
+      String resourceGroup = getResourceGroupFromId(deployment.id());
+      VMDeployment vmDeployment = new VMDeployment();
+      vmDeployment.deployment = deployment;
+      List<PublicIPAddress> list = getIPAddresses(deployment);
+      vmDeployment.ipAddressList = list;
+
+      VirtualMachine vm = api.getVirtualMachineApi(azureGroup).get(id);
+      vmDeployment.virtualMachine = vm;
+      return vmDeployment;
+   }
+
+   @Override
+   public void destroyNode(final String id) {
+      checkState(cleanupResources.apply(id), "server(%s) and its resources still there after deleting!?", id);
+   }
+
+   @Override
+   public void rebootNode(final String id) {
+      api.getVirtualMachineApi(azureGroup).restart(id);
+   }
+
+   @Override
+   public void resumeNode(final String id) {
+      api.getVirtualMachineApi(azureGroup).start(id);
+   }
+
+   @Override
+   public void suspendNode(final String id) {
+      api.getVirtualMachineApi(azureGroup).stop(id);
+   }
+
+   private List<PublicIPAddress> getIPAddresses(Deployment deployment) {
+      List<PublicIPAddress> list = new ArrayList<PublicIPAddress>();
+      String resourceGroup = getResourceGroupFromId(deployment.id());
+
+      if (deployment.properties() != null && deployment.properties().dependencies() != null) {
+         List<Deployment.Dependency> dependencies = deployment.properties().dependencies();
+         for (int d = 0; d < dependencies.size(); d++) {
+            if (dependencies.get(d).resourceType().equals("Microsoft.Network/networkInterfaces")) {
+               List<Deployment.Dependency> dependsOn = dependencies.get(d).dependsOn();
+               for (int e = 0; e < dependsOn.size(); e++) {
+                  if (dependsOn.get(e).resourceType().equals("Microsoft.Network/publicIPAddresses")) {
+                     String resourceName = dependsOn.get(e).resourceName();
+                     PublicIPAddress ip = api.getPublicIPAddressApi(resourceGroup).get(resourceName);
+                     list.add(ip);
+                     break;
+                  }
+               }
+            }
+         }
+      }
+      return list;
+   }
+
+   @Override
+   public Iterable<VMDeployment> listNodes() {
+      List<Deployment> deployments = api.getDeploymentApi(azureGroup).list();
+
+      List<VMDeployment> vmDeployments = new ArrayList<VMDeployment>();
+
+      for (Deployment d : deployments){
+         VMDeployment vmDeployment = new VMDeployment();
+         vmDeployment.deployment = d;
+         VirtualMachineApi vmApi = api.getVirtualMachineApi(azureGroup);
+         vmDeployment.vm = vmApi.getInstanceDetails(d.name());
+         List<PublicIPAddress> list = getIPAddresses(d);
+         vmDeployment.ipAddressList = list;
+
+         VirtualMachine virtualMachine = vmApi.get(d.name());
+         vmDeployment.virtualMachine = virtualMachine;
+
+         vmDeployments.add(vmDeployment);
+      }
+      return vmDeployments;
+   }
+
+   @Override
+   public Iterable<VMDeployment> listNodesByIds(final Iterable<String> ids) {
+      return Iterables.filter(listNodes(), new Predicate<VMDeployment>() {
+         @Override
+         public boolean apply(final VMDeployment input) {
+            return Iterables.contains(ids, input.deployment.name());
+         }
+      });
+   }
+
+}
--- a/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/config/AzureComputeServiceContextModule.java
+++ b/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/config/AzureComputeServiceContextModule.java
@@ -1,0 +1,191 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jclouds.azurecompute.arm.compute.config;
+
+import javax.inject.Named;
+import javax.inject.Singleton;
+import com.google.inject.Provides;
+
+import org.jclouds.azurecompute.arm.compute.AzureComputeServiceAdapter;
+import org.jclouds.azurecompute.arm.compute.functions.VMImageToImage;
+import org.jclouds.azurecompute.arm.compute.functions.DeploymentToNodeMetadata;
+import org.jclouds.azurecompute.arm.compute.functions.VMHardwareToHardware;
+import org.jclouds.azurecompute.arm.compute.functions.LocationToLocation;
+import org.jclouds.azurecompute.arm.domain.VMDeployment;
+import org.jclouds.azurecompute.arm.domain.VMHardware;
+import org.jclouds.azurecompute.arm.domain.VMImage;
+import org.jclouds.azurecompute.arm.domain.Location;
+import org.jclouds.azurecompute.arm.compute.strategy.CreateResourceGroupThenCreateNodes;
+import org.jclouds.azurecompute.arm.AzureComputeApi;
+import org.jclouds.azurecompute.arm.functions.ParseJobStatus;
+import org.jclouds.azurecompute.arm.compute.AzureComputeService;
+
+import org.jclouds.compute.ComputeServiceAdapter;
+import org.jclouds.compute.config.ComputeServiceAdapterContextModule;
+import org.jclouds.compute.domain.Hardware;
+import org.jclouds.compute.domain.NodeMetadata;
+import org.jclouds.compute.strategy.CreateNodesInGroupThenAddToSet;
+import org.jclouds.compute.reference.ComputeServiceConstants.Timeouts;
+import org.jclouds.compute.reference.ComputeServiceConstants.PollPeriod;
+import org.jclouds.compute.ComputeService;
+import static org.jclouds.util.Predicates2.retry;
+import static org.jclouds.compute.config.ComputeServiceProperties.TIMEOUT_NODE_TERMINATED;
+
+import com.google.common.base.Function;
+import com.google.inject.Inject;
+import com.google.inject.TypeLiteral;
+import com.google.common.base.Predicate;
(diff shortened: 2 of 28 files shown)


ritazh:
_bump_ @nacx we also have few more features that depend on this getting merged. (For example, capturing image, creating VM with custom images, and getting the jclouds-example app working against this provider) If there are missing tests, let us know and we will get those created in a followup PR. Thank you!

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
```diff
@@ -0,0 +1,429 @@
...
+               VMDeployment vmDeployment = new VMDeployment();
+               vmDeployment.deployment = deployment;
+               deployments.add(vmDeployment);
+            } else {
+               logger.debug("Failed to create deployment!");
+            }
+            return !deployments.isEmpty();
+         }
+      }, azureComputeConstants.operationTimeout(), 1, SECONDS).apply(name)) {
+         final String illegalStateExceptionMessage = format("Deployment %s was not created within %sms so it will be destroyed.",
```
andreaturli: Where are you destroying the deployment? Does the destroyDeployment takes care of deleting all the cloud resources created for this deployment?

ritazh: @andreaturli The `delete` operation of a template deployment does not take care of deletion of all the resources created as part of the deployment. When you click on "Delete" from the Azure portal, this is the message displayed:

> Delete deployment
Are you sure you want to delete this deployment? This will only delete the deployment history but will leave the resources and the resource group unaffected.

From MSDN document, here is more information on what the `delete` operation does:

> A template deployment that is currently running cannot be deleted. Deleting a template deployment removes the associated deployment operations. Deleting a template deployment does not affect the state of the resource group.

andreaturli: Thanks @ritazh I think this is interesting. In Openstack we have implemented a roll-back operation (see [link](https://github.com/jclouds/jclouds/blob/master/apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/NovaComputeServiceAdapter.java#L152) for more details, which can be an interesting pattern for ARM as well. wdyt?

ritazh: Sure thing! Let me give that a try! 👍

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
```diff
@@ -0,0 +1,429 @@
...
+import org.jclouds.util.Predicates2;
+
+/**
+ * Defines the connection between the {@link AzureComputeApi} implementation and the jclouds
+ * {@link org.jclouds.compute.ComputeService}.
+ */
+@Singleton
+public class AzureComputeServiceAdapter implements ComputeServiceAdapter<VMDeployment, VMHardware, VMImage, Location> {
+
+   private static int runningNumber = 1;
```
nacx: We need to get rid of this shared mutable state

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
```diff
@@ -0,0 +1,429 @@
...
+
+      logger.debug("Deployment created with name: %s", name);
+
+      final Set<VMDeployment> deployments = Sets.newHashSet();
+
+      ResourceGroupApi resourceGroupApi = api.getResourceGroupApi();
+      ResourceGroup resourceGroup = resourceGroupApi.get(azureGroup);
+      final String resourceGroupName;
+
+      if (resourceGroup == null){
```
nacx: Race conditions will appear here when deploying N nodes, as jclouds will call this method concurrently. All "shared" resources should be created *before* creating the actual nodes. This can be done by subclassing the `CreateNodesWithGroupEncodedIntoNameThenAddToSet` class and overriding its `execute` method. Look at how [DigitalOcean](https://github.com/jclouds/jclouds/blob/master/providers/digitalocean2/src/main/java/org/jclouds/digitalocean2/compute/strategy/CreateKeyPairsThenCreateNodes.java) creates the key pairs before creating the nodes and how [it binds that strategy](https://github.com/jclouds/jclouds/blob/master/providers/digitalocean2/src/main/java/org/jclouds/digitalocean2/compute/config/DigitalOcean2ComputeServiceContextModule.java#L92) to the Guice context.
Moving this logic there will eliminate the race condition and the code here will be able to assume the resource group always exists.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
```diff
@@ -0,0 +1,429 @@
...
+      } else {
+         resourceGroupName = resourceGroup.name();
+      }
+
+      if (!retry(new Predicate<String>() {
+         @Override
+         public boolean apply(final String name) {
+            runningNumber++;
+
+            Deployment deployment = api.getDeploymentApi(resourceGroupName).create(name, deploymentTemplate);
```
nacx: Get the deployment api outside the loop to save the class lookup in every iteration.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
```diff
@@ -0,0 +1,429 @@
...
+         }
+      }
+   }
+
+   private List<VMImage> listImagesByLocation(String location) {
+      final List<VMImage> osImages = Lists.newArrayList();
+
+      getImagesFromPublisher("Microsoft.WindowsAzure.Compute", osImages, location);
+      getImagesFromPublisher("MicrosoftWindowsServer", osImages, location);
+      getImagesFromPublisher("Canonical", osImages, location);
```
nacx: Avoid hardcoding this, and configure a comma-separated property instead, so users can customize the publishers? These three values should be preconfigured as a default in the azure api metadata.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
```diff
@@ -0,0 +1,429 @@
...
+      String group = id.substring(0, index);
+
+      // Delete VM
+      URI uri = api.getVirtualMachineApi(azureGroup).delete(id);
+      if (uri != null){
+         boolean jobDone = Predicates2.retry(new Predicate<URI>() {
+            @Override public boolean apply(URI uri) {
+               return ParseJobStatus.JobStatus.DONE == api.getJobApi().jobStatus(uri);
+            }
+         }, 60 * 10 * 1000 /* 5 minute timeout */).apply(uri);
```
nacx: Refactor this into a reusable predicate (see https://github.com/jclouds/jclouds-labs/pull/280#discussion_r65962712). This will also allow users to configure these timeouts and polling.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
```diff
@@ -0,0 +1,429 @@
...
+
+            // Delete NIC
+            uri = api.getNetworkInterfaceCardApi(azureGroup).delete(id + "nic");
+            if (uri != null){
+               jobDone = Predicates2.retry(new Predicate<URI>() {
+                  @Override public boolean apply(URI uri) {
+                     return ParseJobStatus.JobStatus.DONE == api.getJobApi().jobStatus(uri)
+                             || ParseJobStatus.JobStatus.NO_CONTENT == api.getJobApi().jobStatus(uri);
+                  }
+               }, 60 * 10 * 1000 /* 5 minute timeout */).apply(uri);
```
nacx: Same here, move to a reusable predicate and get it injected.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
```diff
@@ -0,0 +1,429 @@
...
+      if (deployment.properties() != null && deployment.properties().dependencies() != null) {
+         List<Deployment.Dependency> dependencies = deployment.properties().dependencies();
+         for (int d = 0; d < dependencies.size(); d++) {
+            if (dependencies.get(d).resourceType().equals("Microsoft.Network/networkInterfaces")) {
+               List<Deployment.Dependency> dependsOn = dependencies.get(d).dependsOn();
+               for (int e = 0; e < dependsOn.size(); e++) {
+                  if (dependsOn.get(e).resourceType().equals("Microsoft.Network/publicIPAddresses")) {
+                     String resourceName = dependsOn.get(e).resourceName();
+                     PublicIPAddress ip = api.getPublicIPAddressApi(resourceGroup).get(resourceName);
+                     list.add(ip);
```
nacx: Should we break here or continue iterating?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/config/AzureComputeServiceContextModule.java
```diff
@@ -0,0 +1,127 @@
...
+      bind(new TypeLiteral<Function<VMHardware, Hardware>>() {
+      }).to(VMHardwareToHardware.class);
+      bind(new TypeLiteral<Function<VMDeployment, NodeMetadata>>() {
+      }).to(DeploymentToNodeMetadata.class);
+
+      //bind(PrioritizeCredentialsFromTemplate.class).to(UseNodeCredentialsButOverrideFromTemplate.class);
+      bind(new TypeLiteral<Function<Location, org.jclouds.domain.Location>>() {
+      }).to(LocationToLocation.class);
+
+      //bind(CreateNodesInGroupThenAddToSet.class).to(GetOrCreateStorageServiceAndVirtualNetworkThenCreateNodes.class);
```
nacx: Remove commented lines

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/config/AzureComputeServiceContextModule.java
```diff
@@ -0,0 +1,127 @@
...
+   }
+
+   @Provides
+   @Named("azureGroupId")
+   protected String getGroupId() {
+      String group =  System.getProperty("test.azurecompute-arm.groupname");
+      if (group == null)
+         group = "jCloudsGroup";
+      return group;
+   }
```
nacx: Remove this test code from the provider code and configure the default value in the azure api metadata. The value can be overridden in the tests, if needed, just like any other property.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/functions/DeploymentToNodeMetadata.java
```diff
@@ -0,0 +1,162 @@
...
+                   build();
+
+   public static Deployment.ProvisioningState provisioningStateFromString(final String text) {
+      if (text != null) {
+         for (Deployment.ProvisioningState status : Deployment.ProvisioningState.values()) {
+            if (text.equalsIgnoreCase(status.name())) {
+               return status;
+            }
+         }
+      }
```
nacx: Use the existing GetEnumValue helper?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/functions/DeploymentToNodeMetadata.java
```diff
@@ -0,0 +1,162 @@
...
+   @Override
+   public NodeMetadata apply(final VMDeployment from) {
+      final NodeMetadataBuilder builder = new NodeMetadataBuilder();
+      final Deployment deployment = from.deployment;
+      builder.id(deployment.name());
+      builder.providerId(deployment.name());
+      builder.name(deployment.name());
+      String group = deployment.name();
+      int index = group.lastIndexOf("-");
+      builder.group(group.substring(0, index));
```
nacx: If this is parsing a jclouds generated name, then the group naming convention already has methods to get the group part. Use that, since the separator used by the convention is also configurable by properties.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/functions/DeploymentToNodeMetadata.java
```diff
@@ -0,0 +1,162 @@
...
+      }
+
+      builder.status(status);
+
+      Credentials credentials = credentialStore.get("node#" + from.deployment.name());
+      if (credentials != null && credentials.identity.equals(JCLOUDS_DEFAULT_USERNAME)) {
+         credentials = new Credentials(AZURE_DEFAULT_USERNAME, credentials.credential);
+      } else if (credentials == null) {
+         credentials = new Credentials(AZURE_DEFAULT_USERNAME, DEFAULT_LOGIN_PASSWORD);
+      }
```
nacx: Hmmm I don't quite get this. Why are you changing the username here?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/functions/DeploymentToNodeMetadata.java
```diff
@@ -0,0 +1,162 @@
...
+         for (int c = 0; c < from.ipAddressList.size(); c++) {
+            PublicIPAddress ip = from.ipAddressList.get(c);
+            if (ip != null && ip.properties() != null && ip.properties().ipAddress() != null)
+               publicIpAddresses.add(ip.properties().ipAddress());
+         }
+         if (publicIpAddresses.size() > 0)
+            builder.publicAddresses(publicIpAddresses);
+      }
+
+      return builder.build();
```
nacx: Also set the image, location and hardware

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/functions/VMHardwareToHardware.java
```diff
@@ -0,0 +1,79 @@
...
+   @Override
+   public Hardware apply(VMHardware from) {
+      final HardwareBuilder builder = new HardwareBuilder()
+              .name(from.name)
+              .id(from.name)
+              .processors(ImmutableList.of(new Processor(from.numberOfCores, 2)))
+              .ram(from.memoryInMB)
+              .location(from.globallyAvailable ? null : FluentIterable.from(locations.get())
+                      .firstMatch(LocationPredicates.idEquals(from.location))
+                      .orNull());
```
nacx: This `orNull` would indicate a global hardware profile. Given how the method in the adapter is built, this should never happen, so I think we'd better fail if none is found?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/functions/VMImageToImage.java
```diff
@@ -0,0 +1,126 @@
...
+
+   private final Supplier<Set<? extends org.jclouds.domain.Location>> locations;
+
+   public static String encodeFieldsToUniqueId(VMImage imageReference){
+      return (imageReference.globallyAvailable ? "global" : imageReference.location) + "/" + imageReference.publisher + "/" + imageReference.offer + "/" + imageReference.sku;
+   }
+
+   public static String[] decodeFieldsFromUniqueId(final String id) {
+      final String[] parts = checkNotNull(id, "id").split("/");
+      return parts;
```
nacx: Just `return checkNotNull(id, "id").split("/");`

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/functions/VMImageToImage.java
```diff
@@ -0,0 +1,126 @@
...
+   }
+
+   @Inject
+   VMImageToImage(@Memoized final Supplier<Set<? extends Location>> locations) {
+      this.locations = locations;
+   }
+
+   @Override
+   public Image apply(final VMImage image) {
+      FluentIterable.from(locations.get()).firstMatch(LocationPredicates.idEquals(image.location)).orNull();
```
nacx: Remove this?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/functions/VMImageToImage.java
```diff
@@ -0,0 +1,126 @@
...
+      final ImageBuilder builder = new ImageBuilder()
+              .name(image.offer)
+              .description(image.sku)
+              .status(Image.Status.AVAILABLE)
+              .version(image.sku)
+              .id(encodeFieldsToUniqueId(image))
+              .providerId(image.publisher)
+              .location(image.globallyAvailable ? null : FluentIterable.from(locations.get())
+                      .firstMatch(LocationPredicates.idEquals(image.location))
+                      .orNull());
```
nacx: Same comment about failing instead of returning null

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/ResourceProviderMetaData.java
```diff
@@ -0,0 +1,68 @@
...
+   public abstract List<String> locations();
+
+   public abstract List<String> apiVersions();
+
+   @SerializedNames({"resourceType", "locations", "apiVersions"})
+   public static ResourceProviderMetaData create(final String resourceType, final List<String> locations, final List<String> apiVersions) {
+      ResourceProviderMetaData.Builder builder = ResourceProviderMetaData.builder()
+              .resourceType(resourceType)
+              .locations(locations == null ? null : ImmutableList.copyOf(locations))
+              .apiVersions(apiVersions == null ? null : ImmutableList.copyOf(apiVersions));
```
nacx: If fields can be really nullable, then annotate the fields accordingly.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VMHardware.java
```diff
@@ -0,0 +1,66 @@
...
+ */
+package org.jclouds.azurecompute.arm.domain;
+
+/**
+ * A VM Size that is available in a region for a given subscription.
+ *
+ * @see <a href="https://msdn.microsoft.com/en-us/library/azure/mt269440.aspx" >api</a>
+ */
+
+public class VMHardware {
```
nacx: Use google auto for internal classes too

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VMImage.java
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jclouds.azurecompute.arm.domain;
+
+public class VMImage {
```
nacx: Use google auto for internal classes too

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceContextLiveTest.java
```diff
@@ -0,0 +1,292 @@
...
+            else
+            {
+               allRestarted = true;
+            }
+         }
+      }
+      assertTrue(allRestarted);
+
+      view.getComputeService().destroyNodesMatching(inGroup(groupName));
+   }
```
nacx: All the above tests are already covered by the `BaseComputeServiceLiveTest`. You should subclass that one and try to make all tests pass without overriding its methods (as that is our compute service implementation contract). Take the [DigitalOcean](https://github.com/jclouds/jclouds/blob/master/providers/digitalocean2/src/test/java/org/jclouds/digitalocean2/compute/DigitalOcean2ComputeServiceLiveTest.java) one as an example.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceContextLiveTest.java
```diff
@@ -0,0 +1,292 @@
...
+      final String groupName = String.format("ubu%s", System.getProperty("user.name").substring(0, 3));
+      final TemplateBuilder templateBuilder = view.getComputeService().templateBuilder();
+      templateBuilder.osFamily(OsFamily.UBUNTU);
+      templateBuilder.osVersionMatches("14.04");
+      templateBuilder.hardwareId("Standard_A0");
+      templateBuilder.locationId("westus");
+      final Template template = templateBuilder.build();
+
+      final TemplateOptions options = template.getOptions();
+      options.inboundPorts(5985);
```
nacx: Is this relevant for this test?

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceContextLiveTest.java
```diff
@@ -0,0 +1,292 @@
...
+   public void testWindowsNode() throws RunNodesException {
+      final String groupName = String.format("win%s", System.getProperty("user.name").substring(0, 3));
+      final TemplateBuilder templateBuilder = view.getComputeService().templateBuilder();
+      templateBuilder.imageId("westus/MicrosoftWindowsServer/WindowsServer/2016-Technical-Preview-with-Containers");
+      templateBuilder.hardwareId("Standard_A0");
+      templateBuilder.locationId("westus");
+      final Template template = templateBuilder.build();
+
+      final TemplateOptions options = template.getOptions().as(TemplateOptions.class);
+      options.inboundPorts(5985);
```
nacx: Same here, is this relevant for this test?

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceContextLiveTest.java
nacx: Also a subclass of the `BaseTemplateBuilderLiveTest` must be added. As usual, you can take the [DigitalOcean](https://github.com/jclouds/jclouds/blob/master/providers/digitalocean2/src/test/java/org/jclouds/digitalocean2/compute/DigitalOcean2TemplateBuilderLiveTest.java) one as an example. Providers should have those tests passing, and try to avoid overriding the base class tests (as much as possible) as that is the contract to guarantee the provider is aligned with the jclouds interfaces and really portable.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/ResourceProviderAPIMockTest.java
```diff
@@ -0,0 +1,53 @@
...
+      String path = String.format("/subscriptions/SUBSCRIPTIONID/providers/%s?api-version=%s", resource, apiVersion);
+
+      assertSent(server, "GET", path);
+      assertTrue(metaDatas.size() > 0);
+      ResourceProviderMetaData md = metaDatas.get(0);
+
+      assertEquals(md.resourceType(), "availabilitySets");
+      assertEquals(md.locations().get(0), "East US");
+      assertEquals(md.apiVersions().get(0), "2016-03-30");
+   }
```
nacx: Add a mock test that exercises the fallback

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/ResourceProviderApiLiveTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+      boolean foundVMResource = false;
+
+      for (ResourceProviderMetaData m : computeMetaData) {
+         if (m.resourceType().equals(VM_RESOURCE_TYPE)) {
+            foundVMResource = true;
+            break;
+         }
+      }
+      assertTrue(foundVMResource);
```
nacx: Use iterables.any?

nacx:
Thanks @ritazh! The structure is pretty good, and it looks like the implementation is already in the right direction. There is a great job in here!

>If there are missing tests, let us know and we will get those created in a followup PR. Thank you!

I already commented the two test base classes that have to be implemented and passing in order to consider the provider ready. I think the current implementation is pretty close to having them! Once you override those, you'll see some tests fail due to credentials, others due to options, etc. Don't get stuck and ping me if you need help! Once those tests pass, we can say the provider is ready to go, even if the extensions are not there yet.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/functions/DeploymentToNodeMetadata.java
ritazh: @jtjk Can you take a look at this?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
ritazh: @nacx This sounds really obvious, but let me just confirm my understanding. Are you thinking of something like:
`properties.put(IMAGE_PUBLISHERS, "Microsoft.WindowsAzure.Compute, MicrosoftWindowsServer, Canonical");` 

then in `getImagesFromPublisher`, loop thru `String[] publishers = properties.getProperty(IMAGE_PUBLISHERS).split(",");`

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
ritazh: Done.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
ritazh: Done.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/functions/VMHardwareToHardware.java
ritazh: @nacx By fail, do you mean an exception? If so, what type of exception here?

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceContextLiveTest.java
ritazh: @nacx Since we have few more PRs that depend on this to be merged, can we work on these two tests as another PR after this one gets merged as is?

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceContextLiveTest.java
ritazh: Removed.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/ResourceProviderAPIMockTest.java
ritazh: Added.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
nacx: Right!

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/functions/VMHardwareToHardware.java
isalento: Using get() instead of orNull should throw  IllegalStateException in case location is not in the list ( which shouldn't happen) 
https://google.github.io/guava/releases/19.0/api/docs/com/google/common/base/Optional.html#get()

andreaturli: I guess @nacx meant that if you have orNull in the second option then it will be equals to the first option.
I think it is better to analyse the location outside the builder and fail-fast, maybe something like [this](https://github.com/jclouds/jclouds/blob/master/providers/digitalocean2/src/main/java/org/jclouds/digitalocean2/compute/functions/DropletToNodeMetadata.java#L157) ?

andreaturli:
hi @ritazh, I'm trying to follow the [README](https://github.com/jclouds/jclouds-labs/blob/8e7d4c057640151ebefbfbd3ef24de8be7a9171f/azurecompute-arm/README.md#run-live-tests) to run the liveTests but I get this error:
```
java.lang.NullPointerException: test.jclouds.oauth.resource

	at com.google.common.base.Preconditions.checkNotNull(Preconditions.java:229)
	at org.jclouds.azurecompute.arm.compute.AzureComputeServiceContextLiveTest.setupProperties(AzureComputeServiceContextLiveTest.java:74)
	at org.jclouds.apis.BaseViewLiveTest.initializeContext(BaseViewLiveTest.java:37)
	at org.jclouds.apis.BaseContextLiveTest.setupContext(BaseContextLiveTest.java:74)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:606)
	at org.testng.internal.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:85)
	at org.testng.internal.Invoker.invokeConfigurationMethod(Invoker.java:552)
	at org.testng.internal.Invoker.invokeConfigurations(Invoker.java:215)
	at org.testng.internal.Invoker.invokeConfigurations(Invoker.java:140)
	at org.testng.internal.TestMethodWorker.invokeBeforeClassMethods(TestMethodWorker.java:170)
	at org.testng.internal.TestMethodWorker.run(TestMethodWorker.java:104)
	at org.testng.TestRunner.privateRun(TestRunner.java:767)
	at org.testng.TestRunner.run(TestRunner.java:617)
	at org.testng.SuiteRunner.runTest(SuiteRunner.java:348)
	at org.testng.SuiteRunner.runSequentially(SuiteRunner.java:343)
	at org.testng.SuiteRunner.privateRun(SuiteRunner.java:305)
	at org.testng.SuiteRunner.run(SuiteRunner.java:254)
	at org.testng.SuiteRunnerWorker.runSuite(SuiteRunnerWorker.java:52)
	at org.testng.SuiteRunnerWorker.run(SuiteRunnerWorker.java:86)
	at org.testng.TestNG.runSuitesSequentially(TestNG.java:1224)
	at org.testng.TestNG.runSuitesLocally(TestNG.java:1149)
	at org.testng.TestNG.run(TestNG.java:1057)
	at org.testng.IDEARemoteTestNG.run(IDEARemoteTestNG.java:74)
	at org.testng.RemoteTestNGStarter.main(RemoteTestNGStarter.java:121)
```

Can you please update the README as well? Thanks!

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/functions/DeploymentToNodeMetadata.java
jtjk: I can't remember all the details now but we needed to do this. At least jClouds default username&password are not suitable with Azure. For example Azure does not allow to use "root" username.

nacx: So, the question here is "what happens if a user sets an invalid username when configuring the TemplateOptions?"
* If the deployment fails, then it is OK.
* If the deployment succeeds, then which username does the VM get?

In any case, jclouds has a mechanism to provide "default" credentials for images. Those default image credentials could be populated to the node if the user does not explicitly set ones.
Instead of setting the credentials constants in the DeploymentTemplateBuilder, you could configurate jclouds to populate them as default image credentials, then they could be accessed from a template with just `tempalte.getDefaultCredentials()` and be read by the DeploymentTemplateBuilder to provide the fallback credentials.

Configuring the default credentials for images just requires to set the `ComputeServiceProperties.IMAGE_LOGIN_USER` property to a sane default value for the username, or to a colon-delimited form providing also a default password: `<username>:<password>`.

If you configure that property in the metadata and read it in the deployment template builder to set it if the user does not explicitly set the credentials, then you could remove this code and assume the node has proper credentials. (This will however rely on the answer to the second point above).

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
andreaturli: Thanks @ritazh! 
I think you can extract the `destroyNode` logic into a function like [CleanupServer](https://github.com/jclouds/jclouds/blob/a4a255fa4ae6c7a1a46aa2a8420b19cd2b7df463/apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/functions/CleanupServer.java)
which is actually a `CleanupAccount` :)

This can be used in the `destroyNode` and in `AzureComputeService` similarly to [this](https://github.com/jclouds/jclouds/blob/a4a255fa4ae6c7a1a46aa2a8420b19cd2b7df463/apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/NovaComputeService.java#L99-L99)

makes sense?

ritazh:
@andreaturli Updated the README

andreaturli:
Thanks @ritazh, just wondering, do you think this
`-Dtest.jclouds.oauth.resource="https://management.azure.com/` parameter is likely to change? or it will be always the same? in that case, maybe we should add it to the [ProviderMetadata](https://github.com/jclouds/jclouds-labs/blob/2b4d283ce54f67114a64a6dad11df01a8d5ec324/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/AzureComputeProviderMetadata.java#L55-L55)

ritazh:
Thanks @andreaturli You are right. It's already in ProviderMetadata 
`properties.put(RESOURCE, "https://management.azure.com/");`
Updated `AbstractAzureComputeApiLiveTest` to remove the check.

andreaturli:
cool!

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
ritazh: Thanks @andreaturli Have few questions about this. Are you available on IRC?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
```diff
@@ -207,11 +207,11 @@ private void getImagesFromPublisher(String publisherName, List<VMImage> osImages
 
    private List<VMImage> listImagesByLocation(String location) {
       final List<VMImage> osImages = Lists.newArrayList();
-
-      getImagesFromPublisher("Microsoft.WindowsAzure.Compute", osImages, location);
-      getImagesFromPublisher("MicrosoftWindowsServer", osImages, location);
-      getImagesFromPublisher("Canonical", osImages, location);
-
+      String[] publishers = this.azureComputeConstants.azureImagePublishers().split(",");
```
andreaturli: consider something like
``
Iterable<String> result = Splitter.on(',')
	       .trimResults()
	       .omitEmptyStrings()
	       .split(str);
``

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceContextLiveTest.java
ritazh: @nacx Please take a look at [this](https://github.com/jclouds/jclouds-labs/pull/279/commits/dd1245789aeeb364c0fcdcf50c8444448b359e1d) commit.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
ritazh: @nacx Can you please take a look at [this](https://github.com/jclouds/jclouds-labs/pull/279/commits/4256a43cd5cd8737e3769a5cfbbf28be7fac02fa) commit

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
```diff
@@ -207,7 +208,7 @@ private void getImagesFromPublisher(String publisherName, List<VMImage> osImages
 
    private List<VMImage> listImagesByLocation(String location) {
       final List<VMImage> osImages = Lists.newArrayList();
-      String[] publishers = this.azureComputeConstants.azureImagePublishers().split(",");
+      Iterable<String> publishers = Splitter.on(',').trimResults().omitEmptyStrings().split(this.azureComputeConstants.azureImagePublishers());
       for (String publisher : publishers) {
          publisher = publisher.trim();
```
andreaturli: I don't think you need to trim again

ritazh:
@andreaturli @nacx Occasionally I'm getting the following error now. Any idea or recommendations?

org.jclouds.rest.AuthorizationException: (jclouds:pw[0cef1fb10f60529028a71f58e54ed07b]@[hostip]:22) (jclouds:pw[0cef1fb10f60529028a71f58e54ed07b]@[hostip:22) error acquiring {hostAndPort=[hostip:22, loginUser=jclouds, ssh=null, connectTimeout=60000, sessionTimeout=60000} (not retryable): Exhausted available authentication methods

andreaturli:
@ritazh generally this means that jclouds couldn't ssh to the VM for the following reasons:
```
- timeout: slow provisioning
- wrong password
- port 22 closed
```
could you identify a test case that is likely to show the problem more often?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
ritazh: @andreaturli Just added this "feature". Please take a look at 6f55f45

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/functions/DeploymentToNodeMetadata.java
ritazh: @nacx I have added these. Please take a look at dec3b7f

ritazh:
@nacx @andreaturli Lots of changes in for this PR. Ready for another round of review. Thanks!

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/config/AzureComputeServiceContextModule.java
```diff
@@ -0,0 +1,189 @@
...
+
+public class AzureComputeServiceContextModule
+        extends ComputeServiceAdapterContextModule<VMDeployment, VMHardware, VMImage, Location> {
+
+   @Override
+   protected void configure() {
+      super.configure();
+      bind(new TypeLiteral<ComputeServiceAdapter<VMDeployment, VMHardware, VMImage, Location>>() {
+      }).to(AzureComputeServiceAdapter.class);
+      bind(new TypeLiteral<Function<VMImage, org.jclouds.compute.domain.Image>>() {
```
andreaturli: As you added `AzureComputeService` you need to bind it in the ContextModule with something like
```
      bind(ComputeService.class).to(AzureComputeService.class);
```

ritazh: @andreaturli Added it in commit c944c77

andreaturli: thanks @ritazh

nacx:
The structure looks good. There are several things such as node credentials and other stuff that has to be refactored, but I think it is in a good state to merge.
Subsequent pull requests should add the mentioned live tests, and those will force the implementation to change things to be done the right way :)

@ritazh Mind squashing the commits so we can merge the PR?
@andreaturli Are you OK with merging this to set the foundation, and add the tests and functional fixes to  get them passing in a subsequent PR?

andreaturli:
Absolutely @nacx, +1

nacx:
Pushed to master as [03f72d5e](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/03f72d5e). Fantastic work done here @ritazh!

PR #281, opened 2016-06-09
Node json should be optional

duncangrant:
The Node element of the container json is optional as well as nullable.

In PR #277 I tested against docker-engine provisioned through docker-machine and docker-swarm but not against vanilla docker engine.  I have now tested against all three.

### DIFF
--- a/docker/src/main/java/org/jclouds/docker/compute/functions/ContainerToNodeMetadata.java
+++ b/docker/src/main/java/org/jclouds/docker/compute/functions/ContainerToNodeMetadata.java
@@ -36,7 +36,6 @@ import org.jclouds.domain.Location;
 import org.jclouds.providers.ProviderMetadata;
 
 import com.google.common.base.Function;
-import com.google.common.base.Strings;
 import com.google.common.base.Supplier;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterables;
@@ -113,8 +112,8 @@ public class ContainerToNodeMetadata implements Function<Container, NodeMetadata
 
    private List<String> getPublicIpAddresses(Container container) {
       String dockerIpAddress;
-      if (container.node() != null && !Strings.isNullOrEmpty(container.node().ip())) {
-         dockerIpAddress = container.node().ip();
+      if (container.node().isPresent()) {
+         dockerIpAddress = container.node().get().ip();
       } else {
          dockerIpAddress = URI.create(providerMetadata.getEndpoint()).getHost();
       }
--- a/docker/src/main/java/org/jclouds/docker/domain/Container.java
+++ b/docker/src/main/java/org/jclouds/docker/domain/Container.java
@@ -26,6 +26,7 @@ import org.jclouds.javax.annotation.Nullable;
 import org.jclouds.json.SerializedNames;
 
 import com.google.auto.value.AutoValue;
+import com.google.common.base.Optional;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 
@@ -77,7 +78,7 @@ public abstract class Container {
 
    @Nullable public abstract String processLabel();
 
-   @Nullable public abstract Node node();
+   public abstract Optional<Node> node();
 
    Container() {
    }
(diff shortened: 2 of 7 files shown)


grkvlt:
@duncangrant Thanks, good catch! The `Optional<?>` type in autovalue is a nice feature!

grkvlt:
@andreaturli This is _another_ Swarm related patch (for master and 1.9.x) that fixes an issue we missed previously, and @duncangrant has now added some tests, too. Can you take a look?

docker/src/main/java/org/jclouds/docker/domain/Container.java
```diff
@@ -77,7 +78,7 @@
 
    @Nullable public abstract String processLabel();
 
-   @Nullable public abstract Node node();
+   @Nullable public abstract Optional<Node> node();
```
nacx: Optional fields should never be nullable.

docker/src/main/java/org/jclouds/docker/domain/Container.java
```diff
@@ -93,7 +94,7 @@ public static Container create(String id, Date created, String path, String name
                                   String resolvConfPath, Map<String, String> volumes, HostConfig hostConfig,
                                   String driver, String execDriver, Map<String, Boolean> volumesRW, String command,
                                   String status, List<Port> ports, String hostnamePath, String hostsPath,
-                                  String mountLabel, String processLabel, Node node) {
+                                  String mountLabel, String processLabel, Optional<Node> node) {
       return new AutoValue_Container(id, created, path, name, copyOf(args), config, state, image, networkSettings,
               sysInitPath, resolvConfPath, copyOf(volumes), hostConfig, driver, execDriver, copyOf(volumesRW), command,
               status, copyOf(ports), hostnamePath, hostsPath, mountLabel, processLabel, node);
```
nacx: Use `Optional.fromNullable(node)` to set the node.

docker/src/main/java/org/jclouds/docker/compute/functions/ContainerToNodeMetadata.java
```diff
@@ -113,8 +112,8 @@ private String cleanUpName(String name) {
 
    private List<String> getPublicIpAddresses(Container container) {
       String dockerIpAddress;
-      if (container.node() != null && !Strings.isNullOrEmpty(container.node().ip())) {
-         dockerIpAddress = container.node().ip();
+      if (container.node() != null && container.node().isPresent()) {
```
nacx: If using optionals we should get rid of the null check and assume it is never null.

duncangrant:
@grkvlt @nacx Thanks for you comments.  I think I've addressed them.

docker/src/main/java/org/jclouds/docker/domain/Container.java
```diff
@@ -266,7 +267,7 @@ public Builder fromContainer(Container in) {
                  .sysInitPath(in.sysInitPath()).resolvConfPath(in.resolvConfPath()).driver(in.driver())
                  .execDriver(in.execDriver()).volumes(in.volumes()).hostConfig(in.hostConfig()).volumesRW(in.volumesRW())
                  .command(in.command()).status(in.status()).ports(in.ports()).hostnamePath(in.hostnamePath())
-                 .hostsPath(in.hostsPath()).mountLabel(in.mountLabel()).processLabel(in.processLabel()).node(in.node());
+                 .hostsPath(in.hostsPath()).mountLabel(in.mountLabel()).processLabel(in.processLabel()).node(in.node().get());
```
grkvlt: This will fail when `node` is `Optional.absent()`

grkvlt:
@duncangrant LGTM @nacx can you take another look?

docker/src/main/java/org/jclouds/docker/compute/functions/ContainerToNodeMetadata.java
nacx: Remove the `!= null` check as it will never be `null`.

nacx:
Just one last minor comment. Let's address this and squash directly so we can cleanly merge the PR. Thanks @duncangrant @grkvlt!

duncangrant:
Addressed and merged.

thanks Duncan

nacx:
Pushed to master as [f47223b5](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/f47223b5). Thanks!

grkvlt:
@duncangrant can you also make a PR for the 1.9.x branch, please?

PR #284, opened 2016-06-13
Docker: getting container IPs of container with network(s)

aledsage:


docker/src/main/java/org/jclouds/docker/domain/NetworkSettings.java
```diff
@@ -59,6 +59,77 @@ public static Details create(String endpointId, String gateway, String ipAddress
          return new AutoValue_NetworkSettings_Details(endpointId, gateway, ipAddress, ipPrefixLen, ipv6Gateway, globalIPv6Address, globalIPv6PrefixLen,
                  macAddress);
       }
+      
+      public static Builder builder() {
```
andreaturli: as you are adding the builders, can you please use [this](https://github.com/google/auto/blob/master/value/userguide/builders.md) ?

andreaturli:
minor comment on the builders. thanks @aledsage

grkvlt:
Looks good @aledsage, just need a version for 1.9.x as well

aledsage:
Thanks @andreaturli for pointing out the `@AutoValue.Builder` usage - so much nicer!
We should re-write the other builders (in master, rather than 1.9.x branch I presume) with that pattern as well. But in a separate PR obviously.

grkvlt:
@aledsage yes, it's much nicer, and agree should update master docker DTOs like this

andreaturli:
thanks @aledsage, merging now!

andreaturli:
Merged at [master](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/6dcadf1f)

PR #286, opened 2016-06-16
Fix incorrect key format check in Docker

grkvlt:
Corrects error in the code that accepts PEM files as data for identity and credential


nacx:
Pushed to master as [38ce54b9](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/38ce54b9). Thanks @grkvlt!

PR #287, opened 2016-06-16
Fix incorrect key format check in Docker

grkvlt:
Port of https://github.com/jclouds/jclouds-labs/pull/286 to 1.9.x branch


nacx:
Pushed to 1.9.x as [73920578](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/73920578). Thanks @grkvlt!

PR #288, opened 2016-06-17
JCLOUDS-1005: Backblaze B2 portable abstraction

andrewgaul:


andrewgaul:
@nacx @zack-shoylev Please review.  This is the last big commit!

b2/src/main/java/org/jclouds/b2/blobstore/B2BlobStore.java
```diff
@@ -0,0 +1,480 @@
...
+         String contentSha1;
+         try {
+            contentSha1 = ByteStreams2.hashAndClose(blob.getPayload().openStream(), Hashing.sha1()).toString();
+         } catch (IOException ioe) {
+            throw Throwables.propagate(ioe);
+         }
+
+         String name = blob.getMetadata().getName();
+
+         // B2 versions all files so we store the original fileId to delete it after the upload succeeds
```
zack-shoylev: I see. Should we be doing this, though? I am not sure, from a user perspective, what the expected behavior would be.

b2/src/main/java/org/jclouds/b2/blobstore/B2BlobStore.java
```diff
@@ -0,0 +1,480 @@
...
+      }
+
+      Bucket bucket = getBucket(container);
+
+      ImmutableList.Builder<StorageMetadata> builder = ImmutableList.builder();
+      B2ObjectList list = api.getObjectApi().listFileNames(bucket.bucketId(), options.getMarker(), options.getMaxResults());
+      Set<String> commonPrefixes = Sets.newHashSet();
+      for (B2ObjectList.Entry entry : list.files()) {
+         // B2 does not support server-side filtering via prefix and delimiter so we emulate it on the client.
+         // TODO: cooperate with maxResults
```
zack-shoylev: What about the TODO?

b2/src/main/java/org/jclouds/b2/blobstore/B2BlobStore.java
andrewgaul: The implemented behavior matches other blobstores.  Otherwise if a user issues an overwrite followed by a delete, the original file reappears.

b2/src/main/java/org/jclouds/b2/blobstore/B2BlobStore.java
andrewgaul: Done.

b2/src/main/java/org/jclouds/b2/blobstore/B2BlobStore.java
zack-shoylev: I see, makes sense.

zack-shoylev:
Looks pretty good.

PR #289, opened 2016-06-18
JCLOUDS-1005: Retry on 500 and 503 errors

andrewgaul:
This requires rewriting the URL for `b2_upload_file` and `b2_upload_part`
requests.

### DIFF
--- a/b2/src/main/java/org/jclouds/b2/filters/B2RetryHandler.java
+++ b/b2/src/main/java/org/jclouds/b2/filters/B2RetryHandler.java
@@ -1,0 +1,90 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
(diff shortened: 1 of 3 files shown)


andrewgaul:
This reference explains retries:

https://www.backblaze.com/b2/docs/integration_checklist.html

A follow-on commit will reuse the authorization token as suggested.

b2/src/main/java/org/jclouds/b2/filters/B2RetryHandler.java
```diff
@@ -0,0 +1,91 @@
...
+
+      return builder.build();
+   }
+
+   @Override
+   public boolean shouldRetryRequest(HttpCommand command, HttpResponse response) {
+      boolean retry = false;
+      try {
+         byte[] data = closeClientButKeepContentStream(response);
+         String message = data != null ? new String(data) : null;
```
nacx: Message seems to be unused. Remove this line?

nacx:
Just one small comment. LGTM!

b2/src/main/java/org/jclouds/b2/filters/B2RetryHandler.java
andrewgaul: Done.

PR #291, opened 2016-06-21
JCLOUDS-664 Azurecompute-arm image capture userdata keyvault

ritazh:
@nacx @andreaturli 
This PR is for Azure ARM Compute Service, part of the foundations needed for interacting with Azure Resource Manager (ARM) Rest APIs. Specifically, this PR includes image capture, userdata, keyvault, shared vnet, and adding `AzureComputeServiceLiveTest` and `AzureTemplateBuilderLiveTest`.

In a future PR, we will replace "Sleep 50" with a statement to wait for lock files and we will handle rate limits.

FYI
@jtjk @isalento @jmspring


azurecompute-arm/pom.xml
```diff
@@ -135,6 +135,12 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-lang3</artifactId>
+      <version>3.4</version>
+    </dependency>
```
nacx: This is added only to join a string. Remove the dependency and use Guava's `Joiner`.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/AzureComputeProviderMetadata.java
```diff
@@ -70,7 +72,10 @@ public static Properties defaultProperties() {
       properties.put(RESOURCE, "https://management.azure.com/");
       properties.put(CREDENTIAL_TYPE, CLIENT_CREDENTIALS_SECRET.toString());
       properties.put(RESOURCE_GROUP_NAME, "jcloudsgroup");
-      properties.put(IMAGE_PUBLISHERS, "Microsoft.WindowsAzure.Compute, MicrosoftWindowsServer, Canonical");
+      properties.put(IMAGE_PUBLISHERS, "Canonical");
+      properties.put(DEFAULT_VNET_ADDRESS_SPACE_PREFIX, "10.0.0.0/16");
+      properties.put(DEFAULT_SUBNET_ADDRESS_PREFIX, "10.0.0.0/24");
+      properties.put(IMAGE_PUBLISHERS, "Canonical,RedHat");
```
nacx: This has already been set before.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
```diff
@@ -149,8 +148,18 @@ public boolean apply(final String name) {
...
+      NodeAndInitialCredentials<VMDeployment> credential = null;
+
+      if (template.getOptions().getPublicKey() != null){
+         String privateKey = template.getOptions().getPrivateKey();
+         credential = new NodeAndInitialCredentials<VMDeployment>(deployment, name,
+                 LoginCredentials.builder().user(loginUser).privateKey(privateKey).authenticateSudo(true).build());
+      } else {
+         credential = new NodeAndInitialCredentials<VMDeployment>(deployment, name,
+                 LoginCredentials.builder().user(loginUser).password(loginPassword).authenticateSudo(true).build());
+      }
```
nacx: All the VM credentials configuration should be changed according to the proposal: https://github.com/jtjk/jclouds-labs/pull/5
This can be done in subsequent PRs, but there are important fixes to be done and race conditions to be removed in the actual code that are resolved there.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
```diff
@@ -176,6 +185,16 @@ public boolean apply(final String name) {
...
+*/
+         VMHardware hwProfile = new VMHardware();
+         hwProfile.name = "Standard_A5";
+         hwProfile.numberOfCores = 2;
+         hwProfile.osDiskSizeInMB = 1047552;
+         hwProfile.resourceDiskSizeInMB = 138240;
+         hwProfile.memoryInMB = 14336;
+         hwProfile.maxDataDiskCount = 4;
+         hwProfile.location = "westus";
+         hwProfiles.add(hwProfile);
```
nacx: We need to remove the hardcoded hardware profiles before merging. We don't want to restrict users just to this configuration.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
```diff
@@ -212,10 +231,12 @@ private void getImagesFromPublisher(String publisherName, List<VMImage> osImages
             osImagesRef.add(vmImage);
          }
       }
+
    }
 
    private List<VMImage> listImagesByLocation(String location) {
       final List<VMImage> osImages = Lists.newArrayList();
+      final List<Publisher> publisherList = api.getOSImageApi(location).listPublishers();
```
nacx: This is not used.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
```diff
@@ -252,6 +273,12 @@ private void checkAndSetImageAvailability(List<VMImage> images, Collection<Strin
    @Override
    public VMImage getImage(final String id) {
       String[] fields = VMImageToImage.decodeFieldsFromUniqueId(id);
+      if (fields[2].substring(0, CUSTOM_IMAGE_PREFIX.length()).equals(CUSTOM_IMAGE_PREFIX)) {
```
nacx: Just `startsWith`?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
```diff
@@ -266,7 +293,11 @@ public VMImage getImage(final String id) {
 
    @Override
    public Iterable<Location> listLocations() {
-      List<Location> locations = api.getLocationApi().list();
+
+      List<Location> locations = new ArrayList<Location>();
+      Location loc = Location.create("westus", "westus", "westus", 20, 20);
+      locations.add(loc);
```
nacx: Hardcoded list of locations have to be removed before merging.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
```diff
@@ -372,9 +409,15 @@ public void suspendNode(final String id) {
...
-         vmDeployment.virtualMachine = virtualMachine;
+         VirtualMachine vm = vmApi.get(d.name());
+         vmDeployment.virtualMachine = vm;
 
+         if (vm != null && vm.tags() != null) {
+            vmDeployment.userMetaData = vm.tags();
+            String tagString = vmDeployment.userMetaData.get("tags");
+            List<String> tags = Arrays.asList(tagString.split(","));
+            vmDeployment.tags = tags;
+         }
```
nacx: Seeing now that a VM has tags. The `DeploymentToNodeMetadata` should set them.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/extensions/AzureComputeImageExtension.java
```diff
@@ -0,0 +1,135 @@
...
+                              VMImageToImage imageReferenceToImage) {
+      this.userExecutor = userExecutor;
+      this.group = azureComputeConstants.azureResourceGroup();
+      this.imageReferenceToImage = imageReferenceToImage;
+      this.api = api;
+   }
+
+   @Override
+   public ImageTemplate buildImageTemplateFromNode(String name, String id) {
+      name = name.toLowerCase();
```
nacx: Better assume immutable parameters. Don't modify references to an input variable.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/extensions/AzureComputeImageExtension.java
```diff
@@ -0,0 +1,135 @@
...
+   }
+
+   @Override
+   public ListenableFuture<Image> createImage(ImageTemplate template) {
+      final CloneImageTemplate cloneTemplate = (CloneImageTemplate) template;
+      String id = cloneTemplate.getSourceNodeId();
+      final String storageAccountName = id.replaceAll("[^A-Za-z0-9 ]", "") + "stor";
+
+      boolean generalized = false;
+      while (!generalized) {
```
nacx: Is this something we want to timeout? How much does that call take? Should we better wait a bit between calls during the loop?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/extensions/AzureComputeImageExtension.java
```diff
@@ -0,0 +1,135 @@
...
+                  }
+                  return false;
+               }
+               catch (Exception e) {
+                  return false;
+               }
+            }
+         }, 15 * 1000 /* 15 second timeout */).apply(uri);
+
+      }
```
nacx: With this approach the method is blocking all the time and returning a future that returns immediately. Instead, the heavy logic should be done by the future to avoid blocking users with this potentially expensive operation.

nacx: Also the timeout for an image to become available should be controlled by the `ComputeServiceProperties.TIMEOUT_IMAGE_AVAILABLE`. Take the [DigitalOcean2ImageExtension](https://github.com/jclouds/jclouds/blob/master/providers/digitalocean2/src/main/java/org/jclouds/digitalocean2/compute/extensions/DigitalOcean2ImageExtension.java) as an implementation example.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/extensions/AzureComputeImageExtension.java
```diff
@@ -0,0 +1,135 @@
...
+         public Image call() throws Exception {
+               return imageReferenceToImage.apply(ref);
+         }
+      });
+
+   }
+
+   @Override
+   public boolean deleteImage(String id) {
+      return false;
```
nacx: Can users upload images but not delete them?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
nacx: I've just seen this PR already changes it. Great! :)

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/functions/VMHardwareToHardware.java
```diff
@@ -48,6 +48,7 @@
    public Hardware apply(VMHardware from) {
       final HardwareBuilder builder = new HardwareBuilder()
               .name(from.name)
+              .providerId("azurecompute-arm")
```
nacx: This is not the id of the provider. This is the "real" id of the hardware profile in the provider; the "backend id". Some providers such as AWS encode the region information in the IDs, so they have things like "us-east-1/ami-0001", etc. But the +real* id in the provider is "ami-0001".

The `id` field is something free that providers can use to encode the information they need in order to be able to properly implement the "getById" methods in the adapter. The `providerId` is meant to hold the real id of the object in the backend provider.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/options/AzureTemplateOptions.java
```diff
@@ -0,0 +1,223 @@
...
+
+   private String customData;
+   private String virtualNetworkAddressPrefix;
+   private String subnetAddressPrefix;
+   private String DNSLabelPrefix;
+   private String keyVaultIdAndSecret;
+
+
+   /**
+    * Sets the CIDR block for virtual network
```
nacx: Invalid javadoc?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/strategy/CreateResourceGroupThenCreateNodes.java
```diff
@@ -54,6 +60,10 @@
    protected Logger logger = Logger.NULL;
 
    private final AzureComputeApi api;
+   private final AzureComputeServiceContextModule.AzureComputeConstants azureComputeConstants;
+   private final String azureGroupName;
+   private String vnetName;
+   private String subnetName;
```
nacx: Initialize in the constructor and make them final?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/strategy/CreateResourceGroupThenCreateNodes.java
```diff
@@ -62,35 +72,69 @@ protected CreateResourceGroupThenCreateNodes(
...
-      } else {
-         resourceGroupName = resourceGroup.name();
+         resourceGroupApi.create(azureGroupName, location, tags).name();
+      }
+
+      this.vnetName = this.azureGroupName + "virtualnetwork";
+      this.subnetName = this.azureGroupName + "subnet";
+
+      if (options.getVirtualNetworkName() != null) {
+         this.vnetName = options.getVirtualNetworkName();
```
nacx: Don't use member variables (shared mutable state!) for things that depend on particular method execution parameters.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/KeyVaultReference.java
```diff
@@ -0,0 +1,49 @@
...
+   public abstract Reference reference();
+
+   @SerializedNames({"reference"})
+   public static KeyVaultReference create(final Reference reference) {
+      return new AutoValue_KeyVaultReference(reference);
+   }
+}
+
+
+
```
nacx: Remove the blank lines at the end :)

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VMImage.java
```diff
@@ -21,6 +21,18 @@
 @AutoValue
 public class VMImage {
 
+   public VMImage(String publisher, String offer, String sku, String version, String location) {
```
nacx: Don't do this,as it is an auto-value anti-pattern. Leave the default constructor not visible, in default scope, and provide a static factory method to instantiate this. Also replace fields by abstract methods. This helper class should be trated just like any other domain class. Apply this also tot he other helper classes that do not properly implement autovalue.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/features/JobApi.java
```diff
@@ -37,5 +43,14 @@
    @GET
    @ResponseParser(ParseJobStatus.class)
    JobStatus jobStatus(@EndpointParam URI jobURI);
+
+   /**
+    * Get status of captured custom image after capture call
+    */
+   @GET
+   @Fallback(Fallbacks.NullOnNotFoundOr404.class)
```
nacx: Try to avoid returning null collections; it is a font of bugs and NPEs. Change the fallback to return an empty list.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/functions/CleanupResources.java
```diff
@@ -36,19 +43,23 @@
...
 
+   private final AzureComputeServiceContextModule.AzureComputeConstants azureComputeConstants;
    @Resource
    @Named(ComputeServiceConstants.COMPUTE_LOGGER)
    protected Logger logger = Logger.NULL;
 
    protected final AzureComputeApi api;
    private Predicate<URI> nodeTerminated;
    private Predicate<URI> resourceDeleted;
+   String azureGroup;
```
nacx: Make final

nacx: All variables

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/functions/CleanupResources.java
```diff
@@ -58,36 +69,67 @@ public CleanupResources(AzureComputeApi azureComputeApi,
...
+                        else {
+                           nsgDeleteStatus = true;
+                        }
+
+                        return deploymentDeleteStatus && storageAcctDeleteStatus && ipDeleteStatus && nsgDeleteStatus;
+                     } else {
+                        return false;
+                     }
                   } else {
                      return false;
```
nacx: Is the shared network that is created before deploying the nodes deleted at some point?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/features/VirtualMachineApi.java
```diff
@@ -131,5 +132,27 @@ VirtualMachine create(@PathParam("vmname") String vmname,
    @Path("/{name}/powerOff")
    void stop(@PathParam("name") String name);
 
+   /**
+    * Generalize the virtual machine
+    */
+   @Named("generalize")
+   @POST
+   @Path("/{name}/generalize")
+   @ResponseParser(StatusCodeParser.class)
```
nacx: What is the purpose of returning the status here? What information does it provide? Can't we just return a `boolean` or `void`?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/util/DeploymentTemplateBuilder.java
```diff
@@ -126,30 +134,49 @@ public Template getTemplate() {
...
+      addNetworkSecurityGroup();
       addNetworkInterfaceCard();
       addVirtualMachine();
 
+
+      DeploymentTemplate.TemplateParameters templateParameters = DeploymentTemplate.TemplateParameters.create(null);
+      DeploymentTemplate.Parameters parameters = DeploymentTemplate.Parameters.create(null);
+      if (keyVaultInUse()){
+         String[] keyVaultInfo = options.getKeyVaultIdAndSecret().split(":");
+         assert keyVaultInfo.length == 2;
```
nacx: Better use Guava's `Preconditions.checkArgument`.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/util/DeploymentTemplateBuilder.java
```diff
@@ -126,30 +134,49 @@ public Template getTemplate() {
...
       addStorageResource();
-      addVirtualNetworkResource();
       addPublicIpAddress();
+      addNetworkSecurityGroup();
       addNetworkInterfaceCard();
       addVirtualMachine();
 
+
+      DeploymentTemplate.TemplateParameters templateParameters = DeploymentTemplate.TemplateParameters.create(null);
+      DeploymentTemplate.Parameters parameters = DeploymentTemplate.Parameters.create(null);
```
nacx: Initialize in an else clause to avoid creating an object just to immediately override it?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/util/DeploymentTemplateBuilder.java
```diff
@@ -126,30 +134,49 @@ public Template getTemplate() {
...
+   public String getDeploymentTemplateJson(DeploymentProperties properties) {
       return json.toJson(properties);
    }
 
    private void addStorageResource() {
-      String storageAccountName = name.replaceAll("[^A-Za-z0-9 ]", "") + "storage";
+      String storageAccountName = name.replaceAll("[^A-Za-z0-9 ]", "") + "stor";
+
+      String storageName = template.getImage().getName();
+      if (storageName.substring(0, CUSTOM_IMAGE_PREFIX.length()).equals(CUSTOM_IMAGE_PREFIX)) {
```
nacx: Use `startsWith`

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/util/DeploymentTemplateBuilder.java
```diff
@@ -257,16 +267,58 @@ private void addNetworkInterfaceCard() {
...
+                    .destinationAddressPrefix("*")
+                    .priority(1234 + i)
+                    .direction(NetworkSecurityRuleProperties.Direction.Inbound)
+                    .build();
+
+            NetworkSecurityRule networkSecurityRule = NetworkSecurityRule.create(
+                    "default-allow-port-" + ports[i],
+                    null,
+                    null,
+                    ruleProperties);
```
nacx: Hmmm where do you set the port to be opened?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/util/DeploymentTemplateBuilder.java
```diff
@@ -275,60 +327,62 @@ private void addVirtualMachine() {
...
-              .createOption("FromImage")
-              .build();
+      boolean usingMarketplaceImage = true;
+      String cusotomImageUri = "";
+
+      // TODO: make new fields for group information
+      String publisher = template.getImage().getProviderId();
+      String storageName = template.getImage().getName();
+      String sku = template.getImage().getDescription(); // this is actual VHD
+      if (storageName.substring(0, CUSTOM_IMAGE_PREFIX.length()).equals(CUSTOM_IMAGE_PREFIX)) {
```
nacx: startsWith

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/util/DeploymentTemplateBuilder.java
```diff
@@ -275,60 +327,62 @@ private void addVirtualMachine() {
...
       //Create Data Disk(s) and add to list
       final String dataDiskName = name + "datadisk";
       variables.put("dataDiskName", dataDiskName);
+      final String dataDiskSize = "100";
 
       List<DataDisk> dataDisks = new ArrayList<DataDisk>();
       DataDisk dataDisk = DataDisk.builder()
               .name("[variables('dataDiskName')]")
-              .diskSizeGB(DEFAULT_DATA_DISK_SIZE)
+              .diskSizeGB(dataDiskSize)
```
nacx: Why not a default value in a constant?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
ritazh: @nacx Let's do that in a subsequent PR, along with the rate limit and the lock file updates. WDYT?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/util/DeploymentTemplateBuilder.java
```diff
@@ -370,19 +433,54 @@ private void addVirtualMachine() {
...
+
+   private OSDisk getOsDisk(String vhdUri, String imageUri) {
+      OSDisk.Builder builder = OSDisk.builder();
+      builder.name("[variables('osDiskName')]");
+      builder.caching("ReadWrite");
+      builder.createOption("FromImage");
+      builder.vhd(VHD.create(vhdUri));
+
+      if (!imageUri.isEmpty()) {
+         builder.osType("Linux");
```
nacx: Is this "Linux" valid for all images with an uri?

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceContextLiveTest.java
```diff
@@ -236,24 +245,7 @@ public void testLinuxNode() throws RunNodesException {
...
-
-   @Test(dependsOnMethods = "testLinuxNode")
-   public void testWindowsNode() throws RunNodesException {
-      final String groupName = this.azureGroup;
-      final TemplateBuilder templateBuilder = view.getComputeService().templateBuilder();
-      templateBuilder.imageId("global/MicrosoftWindowsServer/WindowsServer/Windows-Server-Technical-Preview");
-      templateBuilder.hardwareId("Standard_A0");
+      templateBuilder.hardwareId("Standard_A5");
       templateBuilder.locationId("westus");
       final Template template = templateBuilder.build();
```
nacx: Consider removing this class. Tests here should already be covered by the `AzureComputeServiceLiveTest`.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceLiveTest.java
```diff
@@ -56,18 +90,73 @@ protected ProviderMetadata createProviderMetadata() {
...
       properties.setProperty(TIMEOUT_NODE_RUNNING, scriptTimeout + "");
-      properties.put(RESOURCE_GROUP_NAME, azureGroup);
+      properties.setProperty(TIMEOUT_PORT_OPEN, scriptTimeout + "");
+      properties.setProperty(TIMEOUT_NODE_TERMINATED, scriptTimeout + "");
+      properties.setProperty(TIMEOUT_NODE_SUSPENDED, scriptTimeout + "");
+      properties.put(RESOURCE_GROUP_NAME, "a2");
+//      properties.put("jclouds.max-retries", 5);
+//      properties.put("jclouds.retries-delay-start", 5000L);
+
+      //properties.put(RESOURCE_GROUP_NAME, azureGroup);
```
nacx: Remove all commented code.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceLiveTest.java
```diff
@@ -56,18 +90,73 @@ protected ProviderMetadata createProviderMetadata() {
...
+
+   //   @Override
+   protected Template refreshTemplate() {
+      return this.template = addRunScriptToTemplateWithDelay(this.buildTemplate(this.client.templateBuilder()));
+   }
+
+   protected static Template addRunScriptToTemplateWithDelay(Template template) {
+      template.getOptions().runScript(Statements.newStatementList(new Statement[]{AdminAccess.standard(), Statements.exec("sleep 50"), InstallJDK.fromOpenJDK()}));
+      return template;
+   }
```
nacx: Do these methods need changes in the main jclouds repo?

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceLiveTest.java
```diff
@@ -56,18 +90,73 @@ protected ProviderMetadata createProviderMetadata() {
...
+      this.checkTagsInNodeEquals(node, tags);
+      Logger.getAnonymousLogger().info(String.format("<< available node(%s) os(%s) in %ss", new Object[]{node.getId(), node.getOperatingSystem(), Long.valueOf(createSeconds)}));
+      watch.reset().start();
+      this.client.runScriptOnNode(nodeId, new StatementList(Statements.exec("sleep 50"), JettyStatements.install()), nameTask("configure-jetty"));
+      long configureSeconds = watch.elapsed(TimeUnit.SECONDS);
+      Logger.getAnonymousLogger().info(String.format("<< configured node(%s) with %s and jetty %s in %ss", new Object[]{nodeId, this.exec(nodeId, "java -fullversion"), this.exec(nodeId, JettyStatements.version()), Long.valueOf(configureSeconds)}));
+      this.trackAvailabilityOfProcessOnNode(JettyStatements.start(), "start jetty", node);
+      this.client.runScriptOnNode(nodeId, JettyStatements.stop(), org.jclouds.compute.options.TemplateOptions.Builder.runAsRoot(false).wrapInInitScript(false));
+      this.trackAvailabilityOfProcessOnNode(JettyStatements.start(), "start jetty", node);
+   }
```
nacx: Can this test be removed so the one int he parent class is used instead?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/extensions/AzureComputeImageExtension.java
ritazh: @jtjk: Can you please take a look at this?

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/compute/extensions/AzureComputeImageExtensionLiveTest.java
```diff
@@ -0,0 +1,165 @@
...
+
+      try {
+         nodes = view.getComputeService().createNodesInGroup(groupName, 1, template2);
+      } finally {
+         assertThat(nodes).hasSize(1);
+         view.getComputeService().destroyNodesMatching(inGroup(groupName));
+      }
+      assertNotNull(img);
+
+   }
```
nacx: Remove this test. This should be already covered in the base class.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/VirtualMachineApiLiveTest.java
```diff
@@ -195,6 +195,15 @@ public boolean apply(String name) {
...
       VirtualMachine vm = api().get(getName());
+      assertTrue(list.size() > 0);
+      boolean found = false;
+      for (VirtualMachine virtualMachine : list) {
+         if (virtualMachine.name().equals(getName())){
+            found = true;
+            break;
+         }
+      }
+      assertTrue(found);
```
nacx: Use a more idiomatic `assertTrue(Iterables.any())`.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/extensions/AzureComputeImageExtension.java
ritazh: @jtjk: Please take a look.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/features/JobApi.java
```diff
@@ -37,5 +43,14 @@
...
+   @SelectJson("resources")
+   List<ResourceDefinition> captureStatus(@EndpointParam URI jobURI);
```
nacx: Add the Mock and Live tests for this method.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/features/VirtualMachineApi.java
```diff
@@ -131,5 +132,27 @@ VirtualMachine create(@PathParam("vmname") String vmname,
...
+   @Named("capture")
+   @POST
+   @Payload("%7B\"vhdPrefix\":\"{vhdPrefix}\",\"destinationContainerName\":\"{destinationContainerName}\",\"overwriteVhds\":\"true\"%7D")
+   @MapBinder(BindToJsonPayload.class)
+   @Path("/{name}/capture")
+   @ResponseParser(URIParser.class)
+   URI capture(@PathParam("name") String name,
+               @PayloadParam("vhdPrefix") String vhdPrefix,
+               @PayloadParam("destinationContainerName") String destinationContainerName);
+
```
nacx: Add the Mock and Live tests for the new methods.

nacx:
Thanks @ritazh !

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/features/VirtualMachineApi.java
ritazh: I think this is here because we need to generalize the VM before we can capture its image. @jtjk: what do you think about the recommended change?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/functions/CleanupResources.java
ritazh: @nacx In this case, the shared vnet is like the shared Resource group, they will need to be deleted by the user application since it can contain other resources or be used by other resources.

nacx: Hmmm can we have some way of determining if we can remove it? For example, when the last VM is being removed, etc? Otherwise jclouds will start leaking resources users will potentially be unaware of.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
nacx: Absolutely

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/util/DeploymentTemplateBuilder.java
ritazh: @isalento: why is `destinationPortRange` not set to one of the ports?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/util/DeploymentTemplateBuilder.java
ritazh: @jtjk

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceLiveTest.java
ritazh: @nacx After we add lock file changes, that SHOULD go into main jclouds repo right?

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceLiveTest.java
ritazh: @nacx These are out here because of issues with `weCanCancelTasks` and the `sleep 50` we added. Ideally they should be fixed in the main jclouds repo. WDYT?

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceLiveTest.java
nacx: This is specific to Azure, so not really. A PR should be opened to the main repo to allow subclasses to easily override the script, but the changes to the script itself must remain in this provider.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceLiveTest.java
nacx: I've just opened https://github.com/jclouds/jclouds/pull/977 so you can just override the methods you need, without having to copy entire tests.

nacx: PR is merged now. You should be able to rebase and cleanly override the cancel task and the `weCanCancelTasks` and `addRunScriptToTemplate` methods.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/extensions/AzureComputeImageExtension.java
jtjk: Hmm... if I correctly remember, there was some timing issue and first call was not always working, but second call has been working if first did not :) I guess we can leave it like this or then we should just remove whole while loop and trust that it works without that (as I cannot reproduce that issue now).

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/util/DeploymentTemplateBuilder.java
isalento: Good question. I don't know why, but the fix should be straightforward .destinationPortRange(Integer.toString(ports[i]))

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/extensions/AzureComputeImageExtension.java
jtjk: Hmm... if we do custom image from existing VM, it generalizes the existing VM and we cannot run it anymore. Also it uses the storage of that VM to create image there. Potentially we can do multiple custom images of that same VM there but it does not make any sense as all images are similar. That makes it possible to destroy whole original VM with all related custom images. That is actually only reasonable way to do this. If we want to destroy only created image and leave that "useless" VM there, we need to use Blob storage API (like we need in list images).

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/features/VirtualMachineApi.java
jtjk: We can change that to void

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/extensions/AzureComputeImageExtension.java
nacx: So, when the user creates an image from an existing VM, that VM can still be powered on/off, etc? Or it becomes completely unusable?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/extensions/AzureComputeImageExtension.java
nacx: If it can't be reproduced now I'd remove the loop.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/util/DeploymentTemplateBuilder.java
jtjk: I guess it is not always like that but depending on image it can be Windows as well.

isalento: Yes, imageUri is only set if captured image is used and at the moment only Linux is supported. So it should be a safe bet for now.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceContextLiveTest.java
jtjk: This is nice to have internally during development, but in the end we don't need this

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceLiveTest.java
jtjk: Yes. We need to open PR to make it possible to override refreshTemplate or then make addRunScriptToTemplateWithDelay nonstatic

nacx: Already done :) You should be able to rebase to the latest master and use it! https://github.com/jclouds/jclouds/pull/977

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/functions/CleanupResources.java
ritazh: @nacx One thing we thought about is to create all jclouds resources with a tag "jclouds". Then during delete, we will delete all resources that have the jclouds tag. If not, then we will leave them alone since those resources can be created outside of jclouds APIs.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceLiveTest.java
ritazh: @nacx I think we need to override the entire `createAndRunAServiceInGroup` block since we need to add the `sleep 50` here and later we still need to add `ScriptBuilder.waitForLockFile` here.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/functions/CleanupResources.java
andreaturli: Grear idea! I'd suggest to tag them with the name or id of the vm: in this way jclouds knows exactly who owns each resource and can delete them when appropriate. Wdyt?

ritazh: @nacx @andreaturli If you guys don't mind, let's get this PR done first. Then we will add this create with tag and cleanup by tag feature in a subsequent PR. 

As for adding a tag with the id of the VM, I recommend not because some of these resources are shared between nodes, for example vnet and subnet.

nacx: >Then we will add this create with tag and cleanup by tag feature in a subsequent PR.

Yes, let's better do that in another PR.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/extensions/AzureComputeImageExtension.java
ritazh: I think once the VM is generalized, it cannot be started again.

ritazh:
@nacx Ready for another round of review. It turns out `deleteImage` will require using the storage APIs. So let's hold off on that one in a subsequent PR.

ritazh:
@nacx _bump_ 😄

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/extensions/AzureComputeImageExtension.java
nacx: Ok, so the VM is unusable. Does the `listNodes` return it?

> That makes it possible to destroy whole original VM with all related custom images. That is actually only reasonable way to do this

Does this mean we don't need the blob store API to do this? It would be a good way to delete the image then.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VMHardware.java
```diff
@@ -16,14 +16,12 @@
...
 package org.jclouds.azurecompute.arm.domain;
 
-import com.google.auto.value.AutoValue;
-
 /**
  * A VM Size that is available in a region for a given subscription.
  *
  * @see <a href="https://msdn.microsoft.com/en-us/library/azure/mt269440.aspx" >api</a>
  */
-@AutoValue
```
nacx: Keep this and use a proper auto-value class, like in the `VMImage`.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/extensions/AzureComputeImageExtension.java
```diff
@@ -0,0 +1,149 @@
...
+            for (int c = 0; c < statuses.size(); c++) {
+               if (statuses.get(c).code().substring(0, 10).equals("PowerState")) {
+                  status = statuses.get(c).displayStatus();
+                  break;
+               }
+            }
+            if (status.equals("VM stopped")) {
+               return true;
+            }
+            return false;
```
nacx: Just `return status.equals("VM stopped")`?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/extensions/AzureComputeImageExtension.java
```diff
@@ -0,0 +1,149 @@
...
+         }
+      }, 60 * 4 * 1000).apply(id);
```
nacx: Refactor this in a reusable and injectable predicate the same way than has been done with the image available one. Also, the timeout to be used should be the [TIMEOUT_NODE_SUSPENDED](https://github.com/jclouds/jclouds/blob/master/compute/src/main/java/org/jclouds/compute/reference/ComputeServiceConstants.java#L103-L105).

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/extensions/AzureComputeImageExtension.java
```diff
@@ -0,0 +1,149 @@
...
+                           disks[0] = osProperties.get("name");
+                           disks[1] = datadiskObject.get("name");
+
+                           VirtualMachine vm = api.getVirtualMachineApi(group).get(id);
+                           String location = vm.location();
+                           final VMImage ref = VMImage.create(CUSTOM_IMAGE_PREFIX + group, CUSTOM_IMAGE_PREFIX + storageAccountName, disks[0], disks[1], location, false);
+                           return imageReferenceToImage.apply(ref);
+                        }
+                     }
+                  } catch (Exception e) {
```
nacx: What can fail here? Is there something we would want to log to give users proper feedback of what is going on?

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/VirtualMachineApiMockTest.java
```diff
@@ -182,12 +182,40 @@ public void testStop() throws Exception {
...
+
+   public void testCapture404() throws Exception {
+      server.enqueue(response404());
+
+      final VirtualMachineApi vmAPI = api.getVirtualMachineApi("groupname");
+      URI uri = vmAPI.capture("vm", "prefix", "container");
+      assertNull(uri);
+      assertSent(server, "POST", "/subscriptions/SUBSCRIPTIONID/resourceGroups/groupname/providers/Microsoft.Compute" +
+              "/virtualMachines/vm/capture?api-version=2015-06-15");
+   }
```
nacx: Please, assert the generated body in the capture methods.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/VirtualMachineApiLiveTest.java
```diff
@@ -194,11 +198,81 @@ public boolean apply(String name) {
...
+                  break;
+               }
+            }
+            if (status.equals("VM stopped")) {
+               api().generalize(getName()); // IllegalStateException if failed
+               return true;
+            }
+            return false;
+         }
+      }, 60 * 4 * 1000).apply(getName());
```
nacx: Once the previous comment is addressed, use the injectable predicate here.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/VirtualMachineApiLiveTest.java
```diff
@@ -194,11 +198,81 @@ public boolean apply(String name) {
...
+                        Object dataDisksObject = properties2.get("dataDisks");
+                        ArrayList<Object> dataProperties = (ArrayList<Object>) dataDisksObject;
+                        LinkedTreeMap<String, String> datadiskObject = (LinkedTreeMap<String, String>) dataProperties.get(0);
+
+                        Assert.assertNotNull(osProperties.get("name"));
+                        Assert.assertNotNull(datadiskObject.get("name"));
+                     }
+                     return true;
+                  }
+                  assert true;
```
nacx: Remove this

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/VirtualMachineApiLiveTest.java
```diff
@@ -194,11 +198,81 @@ public boolean apply(String name) {
...
+                  return false;
+               }
+               catch (Exception e) {
+                  assert true;
```
nacx: Remove this

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/VirtualMachineApiLiveTest.java
```diff
@@ -194,11 +198,81 @@ public boolean apply(String name) {
...
+                  return false;
+               }
+            }
+         }, 15 * 1000 /* 15 second timeout */).apply(uri);
+
```
nacx: Is this test failing if the capture operation fails?

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/compute/extensions/AzureComputeImageExtensionLiveTest.java
```diff
@@ -1,165 +0,0 @@
...
-import static org.jclouds.compute.config.ComputeServiceProperties.TIMEOUT_NODE_RUNNING;
-import static org.jclouds.compute.config.ComputeServiceProperties.TIMEOUT_SCRIPT_COMPLETE;
-import static org.jclouds.compute.predicates.NodePredicates.inGroup;
-import static org.testng.Assert.assertNotNull;
-
-/**
- * Live tests for the {@link org.jclouds.compute.extensions.ImageExtension} integration.
- */
-@Test(groups = "live", singleThreaded = true, testName = "DigitalOcean2ImageExtensionLiveTest")
-public class AzureComputeImageExtensionLiveTest extends BaseImageExtensionLiveTest {
```
nacx: I think I didn't express correctly what I meant. This test class needs to exist, but there is no need for additional tests. Just override the bits needed to set the azure provider and default properties, and let just the tests in the superclass run.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceLiveTest.java
nacx: There is no need to override it, as now just the `addRunScriptToTemplate` can be overridden. Just rebase to the latest version of master to get my latest changes to the base test class.

nacx:
Thanks @ritazh and apologies for the delay! Look forward to having this in!

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/extensions/AzureComputeImageExtension.java
ritazh: @nacx Adding a predicate for this in `AzureComputeServiceContrextModule` requires `api.getVirtualMachineApi(group)`, which needs the azure resource group value...what's the best way to pass in that value?

nacx: You can generate the predicate in a provider method. Arguments of provider methods are taken from the injector, so you should be able to pass the constants class as a parameter to get the group there.

ritazh:
@nacx Updated base on your feedback. Ready for another round of review. Still getting 1 test not passing for `AzureComputeServiceLiveTest`. See [log](https://gist.github.com/ritazh/0a91a8b3e8cef00ade8b59aa71443900)

ritazh:
Also, looks like `testCreateImage()` in `BaseImageExtensionLiveTest` has a hardcoded image name: `Assert.assertEquals("test-create-image", image.getName());`. In this provider, the name of the image is derived from: `VMImage.create(CUSTOM_IMAGE_PREFIX + group, CUSTOM_IMAGE_PREFIX + storageAccountName, disks[0], disks[1], location, false);`. Is there an example where I can override the image name to follow this pattern? Currently everything from image id to image name is all derived from the jclouds group name.

nacx:
That's because the contract is that users configure the image name when calling the [buildImageTemplateFromNode](https://github.com/jclouds/jclouds/blob/master/compute/src/main/java/org/jclouds/compute/extensions/ImageExtension.java#L30-L40) method. Is there any restriction to create an image with a provider-specific pattern, or can the code be changed to respect the provided name?

nacx:
> Updated base on your feedback. Ready for another round of review. Still getting 1 test not passing for AzureComputeServiceLiveTest.

I'll give a run later, but it won't block merging this PR. There's great progress here, and we can fix the failing test afterwards without blocking.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/extensions/AzureComputeImageExtension.java
```diff
@@ -0,0 +1,135 @@
...
+                        }
+                     }
+                  }
+               }
+               throw new UncheckedTimeoutException("Image was not created within the time limit: "
+                       + cloneTemplate.getName());
+            }
+         });
+      } else {
+         return null;
```
nacx: The returned future should not be null. Better fail with an IllegalStateException here.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/extensions/AzureComputeImageExtension.java
nacx: Any thoughts on this?

ritazh: @jtjk WDYT?

ritazh: @nacx I think `deleteImage` requires a bit more thoughts. For example, currently when we destroy the node, we also delete the resources attached to it, such as the storage account. When that happens, we essentially deleted the custom images stored in that storage account. We really need to think of a good way to persist that image while we can also clean up resources when destroying a node. If you don't mind, let's leave this out of this PR so we can get the rest of it merged first.

nacx: Makes sense. Mind opening a JIRA issue to track this delete thing?

ritazh: @nacx I have created an issue [here](https://issues.apache.org/jira/browse/JCLOUDS-1134)

nacx:
> That's because the contract is that users configure the image name when calling the buildImageTemplateFromNode method. Is there any restriction to create an image with a provider-specific pattern, or can the code be changed to respect the provided name?

@ritazh Thoughts on this?

ritazh:
@nacx ah yes...I was able to change this logic in [this commit](https://github.com/ritazh/jclouds-labs/commit/12d1ef226147857c8aaf78395974ba9d5978887b). Now `testCreateImage()` passes. But other tests in `BaseImageExtensionLiveTest` are still failing. As you can see here, I have removed the logic to delete storage account that contains the image when `destroyNode()` is called. Let's wait to merge this until the rest of image extension logic are figured out. 
@jtjk FYI

nacx:
Squashed and pushed to master as [2b36a75f](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/2b36a75f). Thanks @ritazh!

PR #294, opened 2016-07-01
[JCLOUDS-1135] improve performance of AzureComputeServiceAdapter.getNode

Ladicek:
The implementation is now O(1) in the number of existing cloud services
instead of O(N). This is achieved by assuming that deployment name
is the same as cloud service name, which is true for all nodes created
by JClouds Azure provider.


nacx:
Thanks @Ladicek!
To make this backwards compatible and not lose the ability to manage nodes not created by jclouds (this is something we should be able to do), could you default to the old behaviour if the deployment name is not the expected one? This way we would have good performance for nodes created by jclouds, and keep the old behaviour for the rest. WDYT?

Ladicek:
@nacx I pushed a change where the new behavior has to be enabled explicitly. Is that what you had in mind?

nacx:
Not exactly, instead of introducing a new configuration property I was thinking about something like:
```java
final Deployment deployment = api.getDeploymentApiForService(id).get(id);
if (deployment == null || deployment.roleInstanceList().isEmpty()) {
   // Instead of returning null, call the old logic that lists all cloud services
}
```
Does this make sense?

Ladicek:
Oh, OK, I'll take a stab at this.

Ladicek:
OK, I pushed another change.

azurecompute/src/main/java/org/jclouds/azurecompute/compute/AzureComputeServiceAdapter.java
```diff
@@ -267,6 +267,19 @@ public boolean apply(final OSImage input) {
 
    @Override
    public Deployment getNode(final String id) {
+      final Deployment deployment = api.getDeploymentApiForService(id).get(id);
+      if (deployment != null && !deployment.roleInstanceList().isEmpty()) {
```
nacx: This conditional is the same one than the used below. Refactor this a bit to reuse the code.

nacx:
The new approach lgtm. Thanks! @andreaturli do you want to have a look?

Ladicek:
I originally kept that small duplicity in place to make the diff smaller, but I factored the common bit out as you wish :-)

nacx:
Pushed to master as [7dd02a01](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/7dd02a01). Thanks @Ladicek! :)

PR #295, opened 2016-07-01
JCLOUDS-664 Azurecompute-arm Update image name logic

ritazh:
@nacx I have updated the logic for creating custom image. Now `testCreateImage()` passes in`BaseImageExtensionLiveTest`. 

@jtjk FYI.

### DIFF
--- a/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/functions/CleanupResources.java
+++ b/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/functions/CleanupResources.java
@@ -34,7 +34,6 @@ import org.jclouds.azurecompute.arm.domain.PublicIPAddress;
 import org.jclouds.azurecompute.arm.domain.VirtualMachine;
 import org.jclouds.compute.reference.ComputeServiceConstants;
 import org.jclouds.logging.Logger;
-import org.jclouds.azurecompute.arm.domain.StorageService;
 
 import com.google.common.base.Function;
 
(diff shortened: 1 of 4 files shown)


ritazh:
@nacx @andreaturli _bump_

nacx:
Pushed to master as [6a51667e](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/6a51667e). Many thanks @ritazh!

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/functions/CleanupResources.java
```diff
@@ -79,12 +78,6 @@ public Boolean apply(String id) {
             boolean deploymentDeleteStatus = false;
 
             if (jobDone) {
-               StorageService ss = api.getStorageAccountApi(group).get(storageAccountName);
-               if (ss != null) {
-                  storageAcctDeleteStatus = api.getStorageAccountApi(group).delete(storageAccountName);
-               } else {
-                  storageAcctDeleteStatus = true;
-               }
```
Ladicek: I wonder why are storage accounts no longer deleted? What the provider creates when creating the node, it should also destroy when destroying the node, no?

ritazh: @Ladicek Currently, Azure ImageExtension is still a WIP. @jtjk is currently working on this. 

Because the Azure capture image feature requires the new image to be stored in a storage account container, `createImage()` adds the image blob into the same storage account the node is created with. For `CleanupResources`, I temporarily removed the logic that deletes storage account when a node is destroyed because deleting the storage account will also delete all the custom images stored in that storage account. Once ImageExtension is completed, we will be able to query images stored in a storage account and delete a storage account only when no custom image exists.

Ladicek: @ritazh That totally makes sense, thanks!

PR #296, opened 2016-07-02
Connect container to list of networks in options

grkvlt:
Allows connecting a container to more than one virtual network. Note that networks must already exist.

### DIFF
--- a/docker/src/main/java/org/jclouds/docker/domain/Network.java
+++ b/docker/src/main/java/org/jclouds/docker/domain/Network.java
@@ -17,13 +17,15 @@
 package org.jclouds.docker.domain;
 
 import static org.jclouds.docker.internal.NullSafeCopies.copyOf;
+import static org.jclouds.docker.internal.NullSafeCopies.copyWithNullOf;
+
 import java.util.List;
 import java.util.Map;
 
+import com.google.auto.value.AutoValue;
+
 import org.jclouds.javax.annotation.Nullable;
 import org.jclouds.json.SerializedNames;
-
-import com.google.auto.value.AutoValue;
 
 @AutoValue
 public abstract class Network {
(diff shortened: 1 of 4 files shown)


grkvlt:
@andreaturli WDYT about `networkList` as the option name?

docker/src/main/java/org/jclouds/docker/domain/Network.java
```diff
@@ -104,4 +104,27 @@ public static Network create(String name, String id, String scope, String driver
       return new AutoValue_Network(name, id, scope, driver, ipam, copyOf(containers), copyOf(options));
```
andreaturli: I think you may want to change it to 
```
      return builder().name(name).id(id).scope(scope).driver(driver).ipam(ipam).containers(copyOf(containers)).options(copyOf(options)).build();
```

andreaturli:
@gkrvlt I think that TemplateOptions.networks https://github.com/jclouds/jclouds/blob/master/compute/src/main/java/org/jclouds/compute/options/TemplateOptions.java is the right mapping. No need for a new custom template option, in my opinion. Wdyt?

docker/src/main/java/org/jclouds/docker/domain/Network.java
```diff
@@ -101,7 +101,30 @@ public static Details create(String endpoint, String macAddress, String ipv4addr
...
+
+      public abstract Builder scope(@Nullable String scope);
+
+      public abstract Builder driver(@Nullable String driver);
+
+      public abstract Builder ipam(IPAM ipam);
+
+      public abstract Builder containers(Map<String, Details> containers);
+
+      public abstract Builder options(Map<String, String> options);
```
nacx: Enforce immutable collections in auto builders. You can follow [this pattern](https://github.com/jclouds/jclouds-labs/blob/master/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/Subnet.java#L67-L84). Basically it consists of providing a hidden builder method that will delegate to auto-value, and have the public one enforcing the immutable collections.

grkvlt:
@andreaturli that's what it uses now, the `TemplateOptions#getNetworks()` property, I will change to use @nacx suggestion for builders and we should be good to merge. For 1.9.x back-port I will need to cherry-pick the `NetworkApi` commits as well.

grkvlt:
@nacx I have made the `Network.Builder` changes you suggested, can you take a look again?

docker/src/main/java/org/jclouds/docker/domain/Network.java
```diff
@@ -17,56 +17,103 @@
...
+         @AutoValue.Builder
+         public abstract static class Builder {
+
+            public abstract Builder subnet(String subnet);
+
+            public abstract Builder ipRange(@Nullable String ipRange);
+
+            public abstract Builder gateway(@Nullable String gateway);
+
+            abstract Config autoBuild();
```
nacx: Hidden build methods are only needed when we want to enforce immutable collections. If no collections are used in the builders, then just provide a public abstract build() method.

grkvlt:
@nacx sorry I left this for so long! I have changed to only use the hidden `autoBuild()` method when it is needed due to collection copying, as you recommened

nacx:
Pushed to master. Thanks @grkvlt!

PR #300, opened 2016-07-13
[azure-arm] add supports to port ranges to addNetworkSecurityGroup

andreaturli:


azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/util/DeploymentTemplateBuilder.java
```diff
@@ -323,6 +326,43 @@ private void addNetworkSecurityGroup() {
 
    }
 
+   // Helper function for simplifying an array of ports to a list of ranges FirewallOptions expects.
+   public static List<String> simplifyPorts(int[] ports){
+      if ((ports == null) || (ports.length == 0)) {
+         return null;
```
nacx: This should never happen at this point, as it is already validated before. Consider it a precondition of this method and better throw an `IllegalArgumentException`. Otherwise you must add a check in the caller to verify that the returned value is not null, to avoid NPEs.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/util/DeploymentTemplateBuilder.java
```diff
@@ -323,6 +326,43 @@ private void addNetworkSecurityGroup() {
 
    }
 
+   // Helper function for simplifying an array of ports to a list of ranges FirewallOptions expects.
+   public static List<String> simplifyPorts(int[] ports){
```
nacx: Add the corresponding unit tests that verify this method works for the different types of input ports: isolated ones, contiguous, etc.

nacx:
+1!

andreaturli:
thanks @nacx, merging now

andreaturli:
Merged at [master](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/b975be5e)

PR #302, opened 2016-07-14
[JCLOUDS-1139] add an option to not deallocate a VM when suspending

Ladicek:


nacx:
Interesting! Thanks @Ladicek! Pushed to master as [787af1dc](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/787af1dc).

PR #305, opened 2016-08-02
JCLOUDS-664 Azurecompute-arm ImageExtension

jtjk:
Azurecompute-arm ImageExtension implementation

### DIFF
--- a/azurecompute-arm/pom.xml
+++ b/azurecompute-arm/pom.xml
@@ -114,6 +114,11 @@
       <scope>test</scope>
     </dependency>
     <dependency>
+      <groupId>org.apache.jclouds.provider</groupId>
+      <artifactId>azureblob</artifactId>
+      <version>${project.parent.version}</version>
+    </dependency>
+    <dependency>
      <groupId>ch.qos.logback</groupId>
      <artifactId>logback-classic</artifactId>
      <scope>test</scope>
@@ -134,6 +139,11 @@
           <artifactId>bcprov-jdk15on</artifactId>
         </exclusion>
       </exclusions>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.jclouds</groupId>
+      <artifactId>jclouds-blobstore</artifactId>
+      <version>${project.parent.version}</version>
     </dependency>
   </dependencies>
 
--- a/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
+++ b/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
@@ -18,7 +18,6 @@ package org.jclouds.azurecompute.arm.compute;
 
 import static java.lang.String.format;
 import static java.util.concurrent.TimeUnit.SECONDS;
-import static org.jclouds.azurecompute.arm.compute.extensions.AzureComputeImageExtension.CUSTOM_IMAGE_PREFIX;
 import static org.jclouds.util.Predicates2.retry;
 import java.util.ArrayList;
 
@@ -45,6 +44,7 @@ import org.jclouds.azurecompute.arm.domain.DeploymentBody;
 import org.jclouds.azurecompute.arm.domain.DeploymentProperties;
 import org.jclouds.azurecompute.arm.domain.NetworkInterfaceCard;
 import org.jclouds.azurecompute.arm.domain.ResourceProviderMetaData;
+import org.jclouds.azurecompute.arm.domain.StorageService;
 import org.jclouds.azurecompute.arm.domain.VMImage;
 import org.jclouds.azurecompute.arm.domain.VMHardware;
 import org.jclouds.azurecompute.arm.domain.Location;
@@ -56,6 +56,7 @@ import org.jclouds.azurecompute.arm.domain.VirtualMachine;
 import org.jclouds.azurecompute.arm.domain.VirtualMachineInstance;
 import org.jclouds.azurecompute.arm.features.DeploymentApi;
 import org.jclouds.azurecompute.arm.features.OSImageApi;
+import org.jclouds.azurecompute.arm.util.BlobHelper;
 import org.jclouds.azurecompute.arm.util.DeploymentTemplateBuilder;
 import org.jclouds.compute.ComputeServiceAdapter;
 import org.jclouds.compute.domain.Template;
@@ -97,7 +98,6 @@ public class AzureComputeServiceAdapter implements ComputeServiceAdapter<VMDeplo
 
       this.api = api;
       this.azureComputeConstants = azureComputeConstants;
-
       this.azureGroup = azureComputeConstants.azureResourceGroup();
 
       logger.debug("AzureComputeServiceAdapter set azuregroup to: " + azureGroup);
@@ -219,7 +219,7 @@ public class AzureComputeServiceAdapter implements ComputeServiceAdapter<VMDeplo
          for (SKU sku : skuList) {
             Iterable<Version> versionList = osImageApi.listVersions(publisherName, offer.name(), sku.name());
             for (Version version : versionList) {
-               VMImage vmImage = VMImage.create(publisherName, offer.name(), sku.name(), version.name(), location, false);
+               VMImage vmImage = VMImage.create(publisherName, offer.name(), sku.name(), version.name(), location);
                osImagesRef.add(vmImage);
             }
          }
@@ -247,6 +247,17 @@ public class AzureComputeServiceAdapter implements ComputeServiceAdapter<VMDeplo
          osImages.addAll(listImagesByLocation(location.name()));
       }
       checkAndSetImageAvailability(osImages, Sets.newHashSet(locationIds));
+
+      // list custom images
+      List<StorageService> storages = api.getStorageAccountApi(azureGroup).list();
+      for (StorageService storage : storages) {
+         String name = storage.name();
+         String key = api.getStorageAccountApi(azureGroup).getKeys(name).key1();
+            List<VMImage> images = BlobHelper.getImages("jclouds", azureGroup, storage.name(), key,
+                  "custom", storage.location());
+            osImages.addAll(images);
+      }
+
       return osImages;
    }
 
@@ -264,24 +275,25 @@ public class AzureComputeServiceAdapter implements ComputeServiceAdapter<VMDeplo
 
    @Override
    public VMImage getImage(final String id) {
-      String[] fields = VMImageToImage.decodeFieldsFromUniqueId(id);
-      if (fields[2].startsWith(CUSTOM_IMAGE_PREFIX)) {
-         String name = fields[2].substring(CUSTOM_IMAGE_PREFIX.length());
-         String sku = fields[3];
-         String version = "1";
-         VMImage ref = VMImage.create(CUSTOM_IMAGE_PREFIX + azureGroup, CUSTOM_IMAGE_PREFIX + name, sku, version, fields[0], false);
-         return ref;
+      VMImage image = VMImageToImage.decodeFieldsFromUniqueId(id);
+      if (image.custom()) {
+         String key = api.getStorageAccountApi(azureGroup).getKeys(image.storage()).key1();
+         if (BlobHelper.customImageExists(image.storage(), key))
+            return image;
+         else
+            return null;
+
       }
 
-      String location = fields[0];
-      String publisher = fields[1];
-      String offer = fields[2];
-      String sku = fields[3];
+      String location = image.location();
+      String publisher = image.publisher();
+      String offer = image.offer();
+      String sku = image.sku();
 
       OSImageApi osImageApi = api.getOSImageApi(location);
       List<Version> versions = osImageApi.listVersions(publisher, offer, sku);
       if (!versions.isEmpty()) {
-         return VMImage.create(publisher, offer, sku, versions.get(0).name(), location, false);
+         return VMImage.create(publisher, offer, sku, versions.get(0).name(), location);
       }
       return null;
    }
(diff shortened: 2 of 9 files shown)


azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/extensions/AzureComputeImageExtension.java
```diff
@@ -69,29 +73,28 @@
...
       this.nodeSuspendedPredicate = nodeSuspendedPredicate;
       this.azureComputeConstants = azureComputeConstants;
+      this.api = api;
    }
 
    @Override
    public ImageTemplate buildImageTemplateFromNode(String name, String id) {
-      String imageName = name.toLowerCase();
-      return new ImageTemplateBuilder.CloneImageTemplateBuilder().nodeId(id).name(imageName).build();
+      name = name.toLowerCase();
```
ritazh: We should probably use local variables and assume immutable parameters. Don't modify references to an input variable.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/extensions/AzureComputeImageExtension.java
```diff
@@ -118,26 +121,47 @@ public Image call() throws Exception {
...
-                           final VMImage ref = VMImage.create(CUSTOM_IMAGE_PREFIX + group, CUSTOM_IMAGE_PREFIX + name, disks[0], disks[1], location, false);
+                           final VMImage ref = VMImage.create(group, storageAccountName, disks[0], disks[1], name, "custom", vm.location());
                            return imageReferenceToImage.apply(ref);
                         }
                      }
                   }
                }
                throw new UncheckedTimeoutException("Image was not created within the time limit: "
-                       + cloneTemplate.getName());
+                     + cloneTemplate.getName());
```
ritazh: `cloneTemplate` and `template` are both used here. Let's stick to the local one to avoid modifying references to an input variable.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/extensions/AzureComputeImageExtension.java
```diff
@@ -118,26 +121,47 @@ public Image call() throws Exception {
...
+         4: imageReference.offer + "/" +
+         5: imageReference.location;
+         */
+         String group = fields[1];
+         String storage = fields[2];
+         StorageServiceKeys keys = api.getStorageAccountApi(group).getKeys(storage);
+
+         // This removes now all the images in this storage. At least in theory, there should be just one and if there is
+         // more, they should be copies of each other.
+         BlobHelper.deleteContainerIfExists(storage, keys.key1(), "system");
```
ritazh: Is it possible to have `deleteContainerIfExists` return some type of status here to verify that the container has been deleted successfully?

azurecompute-arm/pom.xml
```diff
@@ -135,6 +140,11 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
+      <groupId>org.apache.jclouds</groupId>
+      <artifactId>jclouds-blobstore</artifactId>
+      <version>2.0.0-SNAPSHOT</version>
```
nacx: Use `${project.parent.version}` too.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
```diff
@@ -247,6 +247,19 @@ private void getImagesFromPublisher(String publisherName, List<VMImage> osImages
          osImages.addAll(listImagesByLocation(location.name()));
       }
       checkAndSetImageAvailability(osImages, Sets.newHashSet(locationIds));
+
+      // list custom images
+      String offer = "";
```
nacx: No need for this variable, just pass "custom" directly to the BlobHelper call.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
```diff
@@ -265,12 +278,18 @@ private void checkAndSetImageAvailability(List<VMImage> images, Collection<Strin
    @Override
    public VMImage getImage(final String id) {
       String[] fields = VMImageToImage.decodeFieldsFromUniqueId(id);
-      if (fields[2].startsWith(CUSTOM_IMAGE_PREFIX)) {
-         String name = fields[2].substring(CUSTOM_IMAGE_PREFIX.length());
-         String sku = fields[3];
-         String version = "1";
-         VMImage ref = VMImage.create(CUSTOM_IMAGE_PREFIX + azureGroup, CUSTOM_IMAGE_PREFIX + name, sku, version, fields[0], false);
-         return ref;
+      boolean custom = fields.length == 6 ? true : false;
```
nacx: Just `boolean custom = fields.length == 6;`?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/extensions/AzureComputeImageExtension.java
```diff
@@ -118,26 +121,47 @@ public Image call() throws Exception {
...
-                 id, azureComputeConstants.operationTimeout());
+               id, azureComputeConstants.operationTimeout());
          throw new IllegalStateException(illegalStateExceptionMessage);
       }
    }
 
    @Override
    public boolean deleteImage(String id) {
+
+      String[] fields = VMImageToImage.decodeFieldsFromUniqueId(id);
```
nacx: This will at some point become hard to maintain. Could we refactor this method to return an high level object with the fields, instead of an array?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/util/BlobHelper.java
```diff
@@ -0,0 +1,102 @@
...
+              .credentials(storage, key)
+              .buildApi(AzureBlobClient.class);
+
+      azureBlob.deleteContainer(containerName);
+
+      try {
+         azureBlob.close();
+      } catch (IOException e) {
+         e.printStackTrace();
+      }
```
nacx: You could just use `Closeables2.closeQuietly(azureBlob);`

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/util/BlobHelper.java
```diff
@@ -0,0 +1,102 @@
...
+import java.util.List;
+
+public class BlobHelper {
+
+   public static void deleteContainerIfExists(String storage, String key, String containerName) {
+      final AzureBlobClient azureBlob = ContextBuilder.newBuilder("azureblob")
+              .credentials(storage, key)
+              .buildApi(AzureBlobClient.class);
+
+      azureBlob.deleteContainer(containerName);
```
nacx: This should be wrapped in the try/catch block.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/util/BlobHelper.java
```diff
@@ -0,0 +1,102 @@
...
+      }
+   }
+
+   public static boolean customImageExists(String storage, String key) {
+      boolean retval = false;
+      final AzureBlobClient azureBlob = ContextBuilder.newBuilder("azureblob")
+              .credentials(storage, key)
+              .buildApi(AzureBlobClient.class);
+
+      if (azureBlob.containerExists("system"))
```
nacx: Just ` return azureBlob.containerExists("system")`; also wrap it in a try/finally block to close the context.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/util/BlobHelper.java
```diff
@@ -0,0 +1,102 @@
...
+
+         BoundedSet<ContainerProperties> containerList = azureBlob.listContainers();
+         for (ContainerProperties props : containerList) {
+            if (props.getName().equals("system")) {
+               ListBlobsResponse blobList = azureBlob.listBlobs("system");
+               String osDisk = "";
+               String dataDisk = "";
+
+               for (BlobProperties blob : blobList) {
+                  System.out.println(blob.getName());
```
nacx: Don't do this. Consider initializing this helper class with the compute logger instead.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/util/BlobHelper.java
```diff
@@ -0,0 +1,102 @@
...
+               }
+               final VMImage ref = VMImage.create(group, storageAccountName, osDisk, dataDisk, "test-create-image", "custom", location);
+               list.add(ref);
+            }
+         }
+      } catch (Exception e) {
+         e.printStackTrace();
+      }
+      try {
+         azureBlob.close();
```
nacx: Move to a finally block for the previous try

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/util/BlobHelper.java
jtjk: I don't need to log it... it was unintentionally left there :)

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/extensions/AzureComputeImageExtension.java
jtjk: Fixed

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
jtjk: Fixed

jtjk:
All fixed

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
nacx: Remove the variable?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/extensions/AzureComputeImageExtension.java
```diff
@@ -118,26 +121,39 @@ public Image call() throws Exception {
...
+      if (image.custom()) {
+         StorageServiceKeys keys = api.getStorageAccountApi(image.group()).getKeys(image.storage());
+
+         // This removes now all the images in this storage. At least in theory, there should be just one and if there is
+         // more, they should be copies of each other.
+         BlobHelper.deleteContainerIfExists(image.storage(), keys.key1(), "system");
+         if (BlobHelper.customImageExists(image.storage(), keys.key1())) {
+            return false;
+         }
+         return true;
```
nacx: [minor] Just `return !BlobHelper.customImageExists(image.storage(), keys.key1());`

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/util/BlobHelper.java
```diff
@@ -0,0 +1,100 @@
...
+      final AzureBlobClient azureBlob = ContextBuilder.newBuilder("azureblob")
+              .credentials(storage, key)
+              .buildApi(AzureBlobClient.class);
+
+      try {
+         azureBlob.deleteContainer(containerName);
+      }
+      catch (Exception e) {
+         e.printStackTrace();
+      }
```
nacx: This will not fail if the container does not exist (see [here](https://github.com/jclouds/jclouds/blob/master/providers/azureblob/src/main/java/org/jclouds/azureblob/AzureBlobClient.java#L196])). Should we better not catch any exception, leave just the try/finally and let the method fail if the container can't be deleted?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/util/BlobHelper.java
```diff
@@ -0,0 +1,100 @@
...
+              .credentials(storage, key)
+              .buildApi(AzureBlobClient.class);
+
+      try {
+         return azureBlob.containerExists("system");
+      }
+      catch (Exception e) {
+         e.printStackTrace();
+         return false;
+      }
```
nacx: The container exists method should already return false [if the container exists call returns a 404](https://github.com/jclouds/jclouds/blob/master/providers/azureblob/src/main/java/org/jclouds/azureblob/AzureBlobClient.java#L156). Worth removing the catch, as it might be caused by other errors that should be taken into acount?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/util/BlobHelper.java
```diff
@@ -0,0 +1,100 @@
...
+                  if (dataDisk.length() == 0) dataDisk = name.substring(1 + name.lastIndexOf('/'));
+                  else if (osDisk.length() == 0) osDisk = name.substring(1 + name.lastIndexOf('/'));
+               }
+               final VMImage ref = VMImage.create(group, storageAccountName, osDisk, dataDisk, "test-create-image", "custom", location);
+               list.add(ref);
+            }
+         }
+      } catch (Exception e) {
+         e.printStackTrace();
+      }
```
nacx: Same here. Let the method fail on uncontrolled errors.

nacx:
Thanks @jtjk! Just some final comments on failure propagation.
Apart from that, now deleting an image deletes the container. I recall some discussion regarding the storage account deletion. Is this related? If not, should the storage account be now deleted when there are no images/vms in there? And is this something the adapter should take into account when deleting nodes, now that we have the blob utils class in place? (If needed this can be added in a subsequent PR).

jtjk:
Hi, I fixed remaining issues. The storage account deletion should now be created so that if BlobHelper.customImageExists() returns false, storage should be deleted normally. I agree that it should be now fixed as a separate PR

nacx:
Thanks @jtjk!

    Tests run: 5, Failures: 0, Errors: 0, Skipped: 0

All image extension live tests are in green :) Mind rebasing to the latest master and fixing the conflict so I can cleanly merge it?

nacx:
Pushed to master as [76d7aae6](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/76d7aae6). Many thanks @jtjk!

PR #308, opened 2016-08-11
add support for whitelisting locations

andreaturli:
- change location scope to ZONE vs REGION
- edit the README
- fix Region.byName
- add more Regions in Region class


azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
```diff
@@ -92,9 +95,11 @@
 
    private final AzureComputeConstants azureComputeConstants;
 
+   private final ProviderMetadata providerMetadata;
+
    @Inject
    AzureComputeServiceAdapter(final AzureComputeApi api, final AzureComputeConstants azureComputeConstants,
-                              CleanupResources cleanupResources) {
+                              CleanupResources cleanupResources, ProviderMetadata providerMetadata) {
```
nacx: Better inject directly the `@Nullable @Named(LocationConstants.PROPERTY_ZONES) String enabledZones` and split the property and assign the list in the constructor.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/functions/LocationToLocation.java
```diff
@@ -47,20 +47,14 @@
...
    public org.jclouds.domain.Location apply(final Location location) {
       final LocationBuilder builder = new LocationBuilder();
-      String id = location.id();
-      int index = id.lastIndexOf('/');
-      if (index > 0 && (index + 1) < id.length())
-         id = id.substring(index + 1);
-      builder.id(id);
+      builder.id(location.name());
       builder.description(location.displayName());
       builder.parent(getOnlyElement(justProvider.get()));
```
nacx: The parent location of the ZONE should be the REGION instead of the provider?

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceLiveTest.java
```diff
@@ -24,32 +31,25 @@
...
-
+import static org.jclouds.azurecompute.arm.config.AzureComputeProperties.RESOURCE_GROUP_NAME;
+import static org.jclouds.compute.config.ComputeServiceProperties.TIMEOUT_SCRIPT_COMPLETE;
+import static org.jclouds.compute.config.ComputeServiceProperties.TIMEOUT_NODE_RUNNING;
+import static org.jclouds.compute.config.ComputeServiceProperties.TIMEOUT_PORT_OPEN;
+import static org.jclouds.compute.config.ComputeServiceProperties.TIMEOUT_NODE_TERMINATED;
+import static org.jclouds.compute.config.ComputeServiceProperties.TIMEOUT_NODE_SUSPENDED;
 
 /**
  * Live tests for the {@link org.jclouds.compute.ComputeService} integration.
```
nacx: Remove this class from the change set

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/internal/AzureLiveTestUtils.java
```diff
@@ -28,6 +30,7 @@ public static Properties defaultProperties(Properties properties) {
        properties.put("oauth.credential", "password");
        properties.put("oauth.endpoint", "https://login.microsoftonline.com/oauth2/token");
        properties.put(CREDENTIAL_TYPE, CLIENT_CREDENTIALS_SECRET.toString());
+       properties.put(PROPERTY_ZONES, "northeurope,eastus");
```
nacx: How do locations look like in Azure? These looks more like Regions? I mean, zones in providers are usually in the form: northeurope-1, northeurope2, etc.

nacx:
Thanks @andreaturli!

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/internal/AzureLiveTestUtils.java
andreaturli: those are the region names according to http://azure.microsoft.com/en-us/regions

nacx: Yes. If they are regions, then keep the region naming unchanged.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
andreaturli: I think in this way the PROPERTY_ZONES becomes mandatory (otherwise Guice is not happy)
Is it what we want?

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/internal/AzureLiveTestUtils.java
andreaturli:
```
   CENTRAL_US("Central US", "US-IA"),
   EAST_US("East US", "US-VA"),
   EAST_US_2("East US 2", "US-VA"),
   US_GOV_IOWA("US Gov Iowa", "US-IA"),
   US_GOV_VIRGINIA("US Gov Virginia", "US-VA"),
   NORTH_CENTRAL_US("North Central US", "US-IL"),
   SOUTH_CENTRAL_US("South Central US", "US-TX"),
   WEST_US("West US", "US-CA"),
   NORTH_EUROPE("North Europe", "IE"),
   WEST_EUROPE("West Europe", "NL"),
   EAST_ASIA("East Asia", "HK"),
   SOUTH_EAST_ASIA("Southeast Asia", "SG"),
   JAPAN_EAST("Japan East", "JP-11"),
   JAPAN_WEST("Japan West", "JP-27"),
   BRAZIL_SOUTH("Brazil South", "BR"),
   AUSTRALIA_EAST("Australia East", "AU-NSW"),
   AUSTRALIA_SOUTH_EAST("Australia Southeast", "AU-VIC"),
   INDIA_CENTRAL("Central India", "IN-GA"),
   INDIA_SOUTH("South India", "IN-TN"),
   INDIA_WEST("West India", "IN-MH"),
   CHINA_EAST("China East", "CN-SH"),
   CHINA_NORTH("China North", "CN-BJ"),
   CANADA_CENTRAL("Canada Central", "CA-ON"),
   CANADA_EAST("Canada East", "CA-QC");
```

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
nacx: Does it complain too if you annotate it `@Nullable` as suggested?

andreaturli: yes

andreaturli: According to https://github.com/google/guice/wiki/UseNullable I've also tried javax.annotation.Nullable which requires
```
    <dependency>
      <groupId>com.google.code.findbugs</groupId>
      <artifactId>jsr305</artifactId>
    </dependency>
```
but no luck as well

nacx: Can you try removing it from the constructor and put the annotation in the class variable together with an `@Inject(optional = true)`?

andreaturli:
@nacx good to merge?

nacx:
I don't think so. The last comments have not been addressed. We agreed that the change form region to zone was not good, so all that stuff needs to be reverted.

andreaturli:
ok thanks, I guess this one confused me https://github.com/jclouds/jclouds-labs/pull/308#issuecomment-239163128 :)

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/config/AzureComputeHttpApiModule.java
```diff
@@ -46,9 +44,9 @@ protected void bindErrorHandlers() {
    @Override
    protected void installLocations() {
       super.installLocations();
-      bind(ImplicitLocationSupplier.class).
-              to(OnlyLocationOrFirstRegionOptionallyMatchingRegionId.class).
```
andreaturli: @nacx I'm having issues in restoring `OnlyLocationOrFirstRegionOptionallyMatchingRegionId` as it fall-back to `providerUri`which is always the same for each region. Wrong supplier?

andreaturli: I think `FirstRegion` as `ImplicitLocationSupplier` is a better choice here

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
nacx: Have you tried this?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
```diff
@@ -300,28 +288,43 @@ public VMImage getImage(final String id) {
 
    @Override
    public Iterable<Location> listLocations() {
+      final Iterable<String> whiteListZoneName = findWhiteListOfRegions();
```
nacx: Rename to `whiteListedRegionNames`?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
```diff
@@ -300,28 +288,43 @@ public VMImage getImage(final String id) {
...
-         }
-      });
-
-      return result;
+   private Iterable<String> findWhiteListOfRegions() {
+      if (providerMetadata.getDefaultProperties().get(LocationConstants.PROPERTY_REGIONS) == null)  return null;
+      return Splitter.on(",").trimResults().split((CharSequence) providerMetadata.getDefaultProperties().get(LocationConstants.PROPERTY_REGIONS));
    }
 
    private String getResourceGroupFromId(String id) {
```
nacx: If we have a global property that filters locations, we should also filter the `listNodes` operation.

How should the `getImage` and `getNode` behave when invoked with a resource that is from another location?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/functions/LocationToLocation.java
```diff
@@ -47,20 +47,14 @@
...
-
       builder.scope(LocationScope.REGION);
-      final Region region = Region.byName(location.name());
+      final Region region = Region.byName(location.displayName());
       if (region != null) {
          builder.iso3166Codes(ImmutableSet.of(region.iso3166Code()));
       }
-
       return builder.build();
    }
```
nacx: Now that we agree we shouldn't change REGION to ZONE, let's discard the changes in this class.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/config/AzureComputeHttpApiModule.java
nacx: This was working before. If you discard the changes in the location transformation function, can't you just discard the changes in this class too?

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceLiveTest.java
nacx: Just remove this from the change-set

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/internal/AzureLiveTestUtils.java
```diff
@@ -28,6 +30,7 @@ public static Properties defaultProperties(Properties properties) {
        properties.put("oauth.credential", "password");
        properties.put("oauth.endpoint", "https://login.microsoftonline.com/oauth2/token");
        properties.put(CREDENTIAL_TYPE, CLIENT_CREDENTIALS_SECRET.toString());
+       properties.put(PROPERTY_REGIONS, "northeurope,eastus");
```
nacx: When rolling back the location transformation function you'll probably have to change these values. They also appear in the README as an example, so bear in mind to change that too.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
andreaturli: yes, but it doesn't work as well

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
nacx: Also, better process this in the constructor and assign as a final variable of this class.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/config/AzureComputeHttpApiModule.java
nacx: I've had a deeper look at this and I agree with you.

andreaturli:
merged at (master)[http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/89cb6f0f]

PR #309, opened 2016-08-11
JCLOUDS-1153 Fix empty Docker repoTags field in ImageToImage function

kwart:
JIRA: https://issues.apache.org/jira/browse/JCLOUDS-1153

This commit allows to work with empty `repoTags` field in Docker images.

### DIFF
--- a/docker/src/main/java/org/jclouds/docker/compute/functions/ImageToImage.java
+++ b/docker/src/main/java/org/jclouds/docker/compute/functions/ImageToImage.java
@@ -16,9 +16,10 @@
  */
 package org.jclouds.docker.compute.functions;
 
-import com.google.common.base.Function;
-import com.google.common.base.Splitter;
-import com.google.common.collect.Iterables;
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import javax.annotation.Resource;
+import javax.inject.Named;
 
 import org.jclouds.compute.domain.Image;
 import org.jclouds.compute.domain.ImageBuilder;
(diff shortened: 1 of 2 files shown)


docker/src/main/java/org/jclouds/docker/compute/functions/ImageToImage.java
```diff
@@ -91,9 +91,15 @@ public OsFamily apply(final String description) {
...
 
-   private String parseVersion(String description) {
-      String version = get(Splitter.on(":").split(description), 1);
-      logger.debug("os version for item: %s is %s", description, version);
+   private String parseVersion(String repoTag) {
+      String version;
+      if (repoTag.contains(":")) {
+         version = get(Splitter.on(":").split(repoTag), 1);
+      } else {
+         version = "<none>";
```
nacx: Would it be better to use `version = repoTag`? This way we don't couple this method to the placeholder value and make the method itself more flexible.

kwart: I will get rid of the whole `parseVersion` method. It's still not correct. There can be `repoTag` with port provided for repo-host (e.g. `internal-repository.company.example:8888/a/b/c/d:test`)

kwart:
New version was pushed. I've removed the `parseVersion` method and made the relation between the `name` and `osVersion` more clear. A new test was also added for the repoTag which contains host and port.

nacx:
Thanks @kwart! Pushed to master as [05a5ae2e](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/05a5ae2e).

PR #313, opened 2016-08-22
Docker live test fix - SshToCustomPortLiveTest

kwart:
The SshToCustomPortLiveTest didn't work for remote docker endpoints due to wrong port number used. 

This fix allows to decide which `loginPort` to use based on system property configured from the tests.


nacx:
Using a system property gives the impression that there is something that can actually be configured to influence the default behaviour of jclouds. WDYT about changing the approach and let the predicate just pick one port or another based on the container name? We could easily hardcode meaningful names for the containers so the predicate and the tests make sense.

kwart:
@nacx Thx for review.

I've changed the predicate to work with container name instead of the system property.

nacx:
Thanks for the quick fixes! Pushed to master as [66c7617d](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/66c7617d).

PR #314, opened 2016-08-22
JCLOUDS-1158 fix Docker find image Predicate

kwart:
JIRA: https://issues.apache.org/jira/browse/JCLOUDS-1158

This commit contains fix for predicate used to find docker image. It resolves behavior on RHEL (Atomic) which uses "docker.io/" prefixed repoTag entries.

The commit also contains a new unit test to cover the change.

### DIFF
--- a/docker/src/main/java/org/jclouds/docker/compute/strategy/DockerComputeServiceAdapter.java
+++ b/docker/src/main/java/org/jclouds/docker/compute/strategy/DockerComputeServiceAdapter.java
@@ -22,6 +22,7 @@ import static com.google.common.collect.Iterables.find;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.regex.Pattern;
 
 import javax.annotation.Resource;
 import javax.inject.Inject;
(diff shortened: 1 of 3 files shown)


docker/src/main/java/org/jclouds/docker/compute/strategy/DockerComputeServiceAdapter.java
```diff
@@ -329,4 +319,19 @@ public void suspendNode(String id) {
       api.getContainerApi().pause(id);
    }
 
+   protected static Predicate<Image> createPredicateMatchingRepoTags(final String imageIdOrName) {
+      return new Predicate<Image>() {
+         @Override
+         public boolean apply(Image input) {
+            final String imageNameWithHostPrefix = "docker.io/" + imageIdOrName;
```
nacx: Move "docker.io" to a constant that can be referenced from the tests too? IIRC previous versions of Docker prefixed official images with `library/`. I agree this is unlikely to change, but I think it is worth having that prefix isolated in a single constant.

nacx:
Thanks @kwart! Just a minor comment. Apart from that LGTM!

kwart:
I've changed the logic, so it uses regular expression now.

Optional prefix and suffix were moved into constants.

nacx:
Thanks @kwart, you rock! :) Merged to master as [44946467]( http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/44946467).

PR #316, opened 2016-09-02
JCLOUDS-1170 The azurecompute provider should remove disk blobs

kwart:
JIRA: https://issues.apache.org/jira/browse/JCLOUDS-1170

This fixes removing VM disks on Azure. There was missing query parameter in the API call, which caused increasing storage consumption because the data blobs were not deleted for the removed disks.


nacx:
Pushed to master as [d31a85af](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/d31a85af). Thanks!

PR #318, opened 2016-09-16
Fix package export for OSGi

nacx:


demobox:
+1 pending a successful test - looks good to me. Were you able to try this out locally, @nacx?

nacx:
Yes, I've tested all PRs together and ARM works fine in the CLI :)

demobox:
> Yes, I've tested all PRs together and ARM works fine in the CLI :)

Nice ;-) Good to go for me, in that case!

PR #319, opened 2016-09-19
JCLOUDS-1181 oneandone-vpn-api

alibazlamit:
@nacx Ready for review, Thanks.


oneandone/src/main/java/org/apache/jclouds/oneandone/rest/features/VpnApi.java
```diff
@@ -0,0 +1,84 @@
...
+@Path("/vpns")
+@Produces("application/json")
+@Consumes("application/json")
+@RequestFilters(AuthenticateRequest.class)
+public interface VpnApi {
+
+   @Named("vpn:list")
+   @GET
+   @Fallback(Fallbacks.EmptyListOnNotFoundOr404.class)
+   List<Vpn> getList();
```
nacx: Rename to `list`.

oneandone/src/test/java/org/apache/jclouds/oneandone/rest/features/VpnApiLiveTest.java
```diff
@@ -0,0 +1,97 @@
...
+   public void testGet() {
+      Vpn result = vpnApi().get(currentVpn.id());
+
+      assertNotNull(result);
+      assertEquals(result.id(), currentVpn.id());
+   }
+
+   @Test
+   public void testGetConfiguration() throws InterruptedException {
+      Thread.sleep(15000);
```
nacx: Is this really needed? Is there a more deterministic way of waiting?

pom.xml
```diff
@@ -37,7 +37,7 @@
   </scm>
 
   <properties>
-    <jclouds.version>2.0.0-SNAPSHOT</jclouds.version>
+    <jclouds.version>2.0.0-SNAPSHOT</jclouds.version>
```
nacx: [nit] Discard this change.

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/features/VpnApi.java
```diff
@@ -0,0 +1,84 @@
...
+   @Path("/{vpnId}")
+   @Fallback(Fallbacks.NullOnNotFoundOr404.class)
+   Vpn get(@PathParam("vpnId") String vpnId);
+
+   @Named("vpn:configurations:get")
+   @GET
+   @Path("/{vpnId}/configuration_file")
+   @SelectJson("config_zip_file")
+   @Fallback(Fallbacks.NullOnNotFoundOr404.class)
+   String getConfiguration(@PathParam("vpnId") String vpnId);
```
nacx: Interesting, so this method returns a base64 encoded zip file containing the VPN files. Would it make sense to return a `ZipInputStream`, so users can easily consume and store the returned files?

You could create a `Function<String, ZipInputStream>` that takes that base64 string and returns a proper ZipInputStream, and then annotate this method with `@Transform(YourFunction.class)`, so jclouds automatically transforms the response.

WDYT?

alibazlamit: @nacx I think its an unnecessary step,users will have to still convert the ZipInputStream to a file and choose a path,i would leave the whole to who ever wants to use that method,on the other hand if you think its must be there ill implement that.

nacx: Well, I know it is unnecessary, but if we already know we are returning a zip file, why not returning a proper stream?

I mean, we know it is not *a regular string*, the same way that we return objects instead of raw json. Responses are not simple strings and we try to return useful objects. Also many users that consume apis look first at the method signatures and in the worst case they read the docs :) Returning a ZipInputStream here would serve two purposes: better document what this method does, and provide a better interface for users.

Adding that should be trivial, as the function should only ned to decode the string and create a stream for it, not a big deal we'll struggle to maintain.

alibazlamit: OK ill add it

alibazlamit: @nacx Added the function could you please check the PR, Thanks.

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/features/VpnApi.java
```diff
@@ -62,9 +65,10 @@
    @Named("vpn:configurations:get")
    @GET
    @Path("/{vpnId}/configuration_file")
-   @SelectJson("config_zip_file")
+   @ResponseParser(VPNConfigParser.class)
```
nacx: Is this parser class needed? You could leave the `@SelectJson` annotation and just create the transformation function as a top level class.

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/util/VPNConfigParser.java
```diff
@@ -0,0 +1,59 @@
...
+import javax.inject.Singleton;
+import org.apache.jclouds.oneandone.rest.domain.VPNConfig;
+import org.jclouds.http.functions.ParseJson;
+import org.jclouds.json.Json;
+
+@Singleton
+public class VPNConfigParser extends ParseJson<VPNConfig> {
+
+   @Inject
+   public VPNConfigParser(Json json) {
```
nacx: Remove the `public`modifier if this parser class remains.

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/util/VPNConfigParser.java
```diff
@@ -0,0 +1,59 @@
...
+      super(json, TypeLiteral.get(VPNConfig.class));
+   }
+
+   public static class ToZipStream implements Function<VPNConfig, ZipInputStream> {
+
+      @Inject
+      public ToZipStream() {
+         super();
+      }
```
nacx: Remove this constructor?

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/util/VPNConfigParser.java
```diff
@@ -0,0 +1,59 @@
...
+
+      @Override
+      public ZipInputStream apply(VPNConfig input) {
+         try {
+            byte[] decoded = base64().decode(input.content());
+            ZipInputStream zipStream = new ZipInputStream(new ByteArrayInputStream(decoded));
+            return zipStream;
+         } catch (Exception ex) {
+            Logger.getLogger(VPNConfigParser.class.getName()).log(Level.SEVERE, null, ex);
```
nacx: Don't use java logging. jclouds provides a logging abstraction that allows users to configure their preferred logging framework. In order to get the right log, just declare the following class variable of type `org.jclouds.logging.Logger`:

```java
@Resource
protected Logger logger = Logger.NULL;
```

This will configure a null logger by default, but use the one configured by users providing a logging module when creating the jclouds context.

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/util/VPNConfigParser.java
```diff
@@ -0,0 +1,59 @@
...
+            return null;
```
nacx: Instead of returning `null` on *any* failure, are there situations where it makes sense returning null? Can we return null just in those cases, and propagate the exception otherwise?

alibazlamit: Removed the try catch, if the API fails to return the config the fallback with work as needed and return a null object, no need for a try catch in there, if any other error should be returned from the API the error handler will do its job.

Changes pushed

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/util/VPNConfigParser.java
nacx: The visibility of this constructor still needs to be reduced by removing the public modifier (see [this](https://github.com/google/guice/wiki/KeepConstructorsHidden)). Or you can completely remove the parser class and bring back the `@SelectJson` annotation in the api method.

alibazlamit: I tried removing the parser class that create an error it could not parse the JSON correctly for a reason so i decided to leave it like this, i have removed the public modifier and pushed the change. Thanks

nacx: Perfect. Thanks!

nacx:
Mind squashing the commits so I can cleanly merge it?

alibazlamit:
Squashed. Thanks.

nacx:
Pushed to master as [4e397202](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/4e397202). Thanks @alibazlamit!

PR #321, opened 2016-09-21
Use the existing region suppliers to filter

nacx:
@andreaturli Can you give this a try?


andreaturli:
looks so clean, thx @nacx ! Merge it right away!

PR #323, opened 2016-09-28
Fix AzureTemplateBuilderLiveTest and AzureComputeServiceLiveTest

andreaturli:
AzureTemplateBuilderLiveTest is green


nacx:
@andreaturli Looks good. Once thing though: you might need to set [these properties](https://github.com/jtjk/jclouds-labs/pull/5/files#diff-40dee6156218fbe467914a6576b89080R73) too so jclouds configures those credentials in case users don't provide any credential. Sounds good?

Also the changes to the `CreateResourceGroupThenCreateNodes.java` made in https://github.com/jtjk/jclouds-labs/pull/5 would be nice to have.

andreaturli:
Thx @nacx! Not got to the bottom of it, but if I specify the default creds for the images, the test that tries to override them fail!

andreaturli:
@nacx any suggestions to fix https://jclouds.ci.cloudbees.com/job/jclouds-labs-pull-requests/1526/org.apache.jclouds.labs$azurecompute-arm/console ?

andreaturli:
I think this is superseded by #324 so closing it now

PR #326, opened 2016-10-14
Create one resource group in each region

nacx:
This changes creates one resource group for each location, instead of having one fixed resource group. Without this change it is not possible to use one single jclouds context to deploy nodes in different locations, because it will attempt to create the same resource group in each region.

Resource groups are created in the form `<prefix>-<locationname>`, where the prefix can be configured when creating the context with the usual jclouds property `ComputeServiceProperties.RESOURCENAME_PREFIX`. This way, the creation of the resource group will follow the same pattern that other resources that are created using the group naming convention (although resource groups will have always the location name as a suffix).

Still pending to run the complete live test suite, so please, do not merge yet.

/cc @andreaturli


nacx:
Squashed and pushed to master as [f58f96b1](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/f58f96b1).

PR #328, opened 2016-10-18
Configuration to deploy VMs from the Marketplace

nacx:
This also introduces a small change to store the jclouds group in the virtual machine tags so we can properly identify vms with custom names in the corresponding group.

With this commit we also have the version details (there is an extra api call needed to determine if the vm is from the Marketplace) in each image, and we can better fallback to the generic operating system (Linux/Windows) if we can't parse the specific version.

**NOTE**: *To deploy images from the Marketplace, one must first go to the Marketplace, select the image then "enable it for programmatic deployment" by clicking the link at the bottom and enabling it for the subscription.*

/cc @andreaturli


nacx:
Compute live tests are passing. Merged to master.

PR #329, opened 2016-10-20
Added rate limit module

nacx:
Add an optional rate limit module so users can configure jclouds to automatically throttle requests when the limits are reached.


andreaturli:
Didnt know azure has that rate limit, thx @nacx

PR #332, opened 2016-10-28
Properly track the status of write operations

nacx:
Write operations in ProfitBricks return the Location header with the URL where the request status can be tracked: https://devops.profitbricks.com/api/cloud/v2/#creation-modification-and-deletion-responses

Use this to poll the status of the operation instead of polling the existence of a resource. This way if the operation fails we notice earlier (don't wait for the polling timeout) and can give proper feedback on the error.


nacx:
Squashed and pushed to master as [c4e36e72](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/c4e36e72).

PR #335, opened 2016-11-07
Remove duplicated sharedstorage test inputs

andrewgaul:


demobox:
Thanks for catching this, @andrewgaul! Do we know which of the sets of resources is newer? Or are they identical?

Also, on a separate note, would it be worth sending an email about this to the dev list and/or adding this to the wiki or documentation?

/cc @nacx

andrewgaul:
`diff -r oneandone/src/test/resources/shared*` confirms that these files are identical.

demobox:
> diff -r oneandone/src/test/resources/shared* confirms that these files are identical

Thanks, @andrewgaul!

nacx:
Pushed to master as [cb06b681](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/cb06b681).

PR #336, opened 2016-12-12
JCLOUDS-1216 - oneandone-user-api

alibazlamit:
@nacx the Users PR is ready for review, i had one issue with this specific PR since the API does not allow to use many of the endpoints that affect the user info unless its the same user that is being used to make the request. all these tests are commented out now.


oneandone/src/main/java/org/apache/jclouds/oneandone/rest/domain/User.java
```diff
@@ -0,0 +1,584 @@
...
+   public abstract String name();
+
+   @Nullable
+   public abstract String description();
+
+   @Nullable
+   public abstract String email();
+
+   @Nullable
+   public abstract String state();
```
nacx: Should this be of type `UserState`?

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/domain/User.java
```diff
@@ -0,0 +1,584 @@
...
+   }
+
+   @AutoValue
+   public abstract static class AddUserIp {
+
+      public abstract List<String> ips();
+
+      @SerializedNames({"ips"})
+      public static AddUserIp create(List<String> ips) {
+         return new AutoValue_User_AddUserIp(ips == null ? ImmutableList.<String>of() : ips);
```
nacx: Enforce an immutable list if the parameter is not null.

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/domain/User.java
```diff
@@ -0,0 +1,584 @@
...
+      }
+   }
+
+   @AutoValue
+   public abstract static class Api {
+
+      public abstract Boolean active();
```
nacx: Prefer primitive types for non-nullable fields.

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/domain/User.java
```diff
@@ -0,0 +1,584 @@
...
+
+      @Nullable
+      public abstract String key();
+
+      @Nullable
+      public abstract List<String> allowedIps();
+
+      @SerializedNames({"active", "key", "allowed_ips"})
+      public static Api create(Boolean active, String key, List<String> allowedIps) {
+         return new AutoValue_User_Api(active, key, allowedIps == null ? ImmutableList.<String>of() : allowedIps);
```
nacx: Immutable when not null.

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/domain/User.java
```diff
@@ -0,0 +1,584 @@
...
+      }
+
+      @AutoValue
+      public abstract static class UpdateApi {
+
+         public abstract Boolean active();
```
nacx: Same about primitives.

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/features/UserApi.java
```diff
@@ -0,0 +1,127 @@
...
+import org.jclouds.Fallbacks;
+import org.jclouds.rest.annotations.BinderParam;
+import org.jclouds.rest.annotations.Fallback;
+import org.jclouds.rest.annotations.MapBinder;
+import org.jclouds.rest.annotations.RequestFilters;
+import org.jclouds.rest.annotations.SelectJson;
+import org.jclouds.rest.binders.BindToJsonPayload;
+
+@Path("/users")
+@Produces("application/json")
```
nacx: Remove this and leave only the method specific ones

oneandone/src/test/java/org/apache/jclouds/oneandone/rest/features/UserApiLiveTest.java
```diff
@@ -0,0 +1,162 @@
...
+   @Test(dependsOnMethods = "testGet")
+   public void testUpdate() throws InterruptedException {
+      //this test will fail too since its not allowed to update other users from the API
+//      String updatedDescription = "updatejclouds";
+//
+//      User updateResult = userApi().update(currentUser.id(), User.UpdateUser.builder()
+//              .description(updatedDescription)
+//              .build());
+//
+//      assertEquals(updateResult.description(), updatedDescription);
```
nacx: Can't we configure these tests to update the logged user? Could create a new API context for the just created user and use that to implement these tests?

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/features/UserApi.java
alibazlamit: `@Produces("application/json")` we have been using this approach so far not a method specific tag, you recommended this approach in the beginning.

nacx: Hmmm I think there was some misunderstanding then. We should only add the "content-type" headers in the requests that generate a body, and that's what that annotation does. If we put it at class level, all requests will have it, even GET requests without body. I don't know if the API needs it, but we'd better put it just where needed.

oneandone/src/test/java/org/apache/jclouds/oneandone/rest/features/UserApiLiveTest.java
nacx: Hmmm I don't really know. The questions to answer are:

* Why would we have tests we won't be able to run?
* Under which circumstances could we run a complete test suite for this API? What preconditions should be met?
* Can we dynamically configure those preconditions so we can run the test suite?

alibazlamit: @nacx 
I think that this PR and the Users API does not make a lot of sense to jClouds, most of the operations are only allowed to the current user, and changing the API key or deleting the users might break any flow, if you agree i think this feature could be skipped for jClouds, WDYT?

nacx: I agree this should not be our highest priority.
Moving forward, do you think the existing APIs cover the funcionality to implement the ComputeService abstraction?

alibazlamit:
I do believe so, we have covered all the important server operations, I will delete this PR and start a new one for the compute api.

oneandone/src/test/java/org/apache/jclouds/oneandone/rest/features/UserApiLiveTest.java
alibazlamit: I do believe so, we have covered all the important server operations, I will delete this PR and start a new one for the compute api.

nacx: That sounds great. Thanks! :)

PR #340, opened 2017-01-10
Add skeleton for ApiMetadata and ProviderMetadata

andreaturli:
- add ApiMetadata and ProviderMetadata
- add skeleton for PacketApi with ProjectApi only
- add XAuthTokenToRequest filter
- add HttpApiModule and ParserModule
- add ProjectApi feature with Mock and Live Tests


packet/src/main/java/org/jclouds/packet/PacketApiMetadata.java
```diff
@@ -0,0 +1,85 @@
...
+      this(new Builder());
+   }
+
+   protected PacketApiMetadata(Builder builder) {
+      super(builder);
+   }
+
+   public static Properties defaultProperties() {
+      Properties properties = BaseHttpApiMetadata.defaultProperties();
+      properties.put(TEMPLATE, "osFamily=UBUNTU,os64Bit=true,osVersionMatches=14.*");
```
nacx: Ccan we default to a more modern version?

packet/src/main/java/org/jclouds/packet/PacketApiMetadata.java
```diff
@@ -0,0 +1,85 @@
...
+      properties.put(TIMEOUT_NODE_RUNNING, 300000); // 5 mins
+      return properties;
+   }
+
+   public static class Builder extends BaseHttpApiMetadata.Builder<PacketApi, Builder> {
+
+      protected Builder() {
+         id("packet")
+                 .name("Packet API")
+                 .identityName("We can use project Id")
```
nacx: Let's say what needs to be put as the identity. Users will read this to know what to put, so let's be concrete and pick one option.

packet/src/main/java/org/jclouds/packet/PacketApiMetadata.java
```diff
@@ -0,0 +1,85 @@
...
+                 .credentialName("Must be Token")
+                 .documentation(URI.create("https://www.packet.net/help/api/#"))
+                 .defaultEndpoint("https://api.packet.net")
+                 .defaultProperties(PacketApiMetadata.defaultProperties())
+                 .view(typeToken(ComputeServiceContext.class))
```
nacx: Comment this out until the ComputeService interface is implemented.

packet/src/main/java/org/jclouds/packet/PacketProviderMetadata.java
```diff
@@ -0,0 +1,77 @@
...
+    public static class Builder extends BaseProviderMetadata.Builder {
+
+        protected Builder() {
+            id("packet")
+                    .name("Packet Compute Services")
+                    .apiMetadata(new PacketApiMetadata())
+                    .homepage(URI.create("https://www.packet.net/"))
+                    .console(URI.create("https://app.packet.net/portal"))
+                    .endpoint("https://api.packet.net")
+                    .defaultProperties(PacketProviderMetadata.defaultProperties());
```
nacx: Do we know which regions are available so we can put the iso codes here?

packet/src/main/java/org/jclouds/packet/config/PacketHttpApiModule.java
```diff
@@ -0,0 +1,61 @@
...
+import com.google.inject.Scopes;
+
+@ConfiguresHttpApi
+public class PacketHttpApiModule extends HttpApiModule<PacketApi> {
+
+   @Override
+   protected void configure() {
+      install(new PacketComputeParserModule());
+      super.configure();
+   }
```
nacx: We could remove this if we add the parser module to the list of modules in the api metadata. Better to have the complete list there?

packet/src/main/java/org/jclouds/packet/config/PacketHttpApiModule.java
```diff
@@ -0,0 +1,61 @@
...
+   @Override
+   protected void bindErrorHandlers() {
+      bind(HttpErrorHandler.class).annotatedWith(Redirection.class).to(PacketErrorHandler.class);
+      bind(HttpErrorHandler.class).annotatedWith(ClientError.class).to(PacketErrorHandler.class);
+      bind(HttpErrorHandler.class).annotatedWith(ServerError.class).to(PacketErrorHandler.class);
+   }
+
+   @Override
+   protected void bindRetryHandlers() {
+      bind(HttpRetryHandler.class).annotatedWith(ClientError.class).to(BackoffLimitedRetryHandler.class);
```
nacx: Hmmmm, this means we are going to retry 4xx errors. Is that what we want? Those usually are bad requests that will always fail. Retrying should be done by default just for server errors.

packet/src/main/java/org/jclouds/packet/config/PacketHttpApiModule.java
```diff
@@ -0,0 +1,61 @@
...
+   }
+
+   @Override
+   protected void installLocations() {
+      install(new LocationModule());
+      bind(ImplicitLocationSupplier.class).to(FirstRegion.class).in(Scopes.SINGLETON);
```
nacx: Move the binding to the configure method and remove this override?

packet/src/main/java/org/jclouds/packet/features/ProjectApi.java
```diff
@@ -0,0 +1,47 @@
...
+public interface ProjectApi {
+
+    /**
+     * List all projects
+     */
+    @Named("project:list")
+    @GET
+    @SelectJson("projects")
+    @Fallback(Fallbacks.EmptyListOnNotFoundOr404.class)
+    List<Project> list();
```
nacx: Since Packet supports pagination, we should add it as soon as possible, to set the foundation for all the APIs. I'd configure this list to support pagination right in this PR.
Looking at the docs, it works very very similar to the DigitalOcean2 one, so you could take it as an example (for example the [ActionApi](https://github.com/jclouds/jclouds/blob/master/providers/digitalocean2/src/main/java/org/jclouds/digitalocean2/features/ActionApi.java) class, which is a small one) and get it working soon.

packet/src/main/java/org/jclouds/packet/filters/AddXAuthTokenToRequest.java
```diff
@@ -0,0 +1,47 @@
...
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+@Singleton
+public class AddXAuthTokenToRequest implements HttpRequestFilter {
+
+    private final Supplier<Credentials> creds;
+
+    @Inject
+    public AddXAuthTokenToRequest(@Provider Supplier<Credentials> creds) {
```
nacx: Remove the public modifier.

packet/src/main/java/org/jclouds/packet/filters/AddXAuthTokenToRequest.java
```diff
@@ -0,0 +1,47 @@
...
+        this.creds = checkNotNull(creds, "creds");
```
nacx: The check is redundant. Injection will fail if it is not present.

packet/src/main/java/org/jclouds/packet/handlers/PacketErrorHandler.java
```diff
@@ -0,0 +1,63 @@
...
+      byte[] data = closeClientButKeepContentStream(response);
+      String message = data != null ? new String(data) : null;
+
+      Exception exception = message != null ? new HttpResponseException(command, response, message)
+              : new HttpResponseException(command, response);
+      message = message != null ? message : String.format("%s -> %s", command.getCurrentRequest().getRequestLine(),
+              response.getStatusLine());
+      switch (response.getStatusCode()) {
+         case 400:
+            break;
```
nacx: 400 errors are Bad Request errors which are usually populated as IllegalArgumentException errors.

packet/src/main/java/org/jclouds/packet/PacketApiMetadata.java
andreaturli: yes they offer ubuntu 16.04 LTS, will update

packet/src/main/java/org/jclouds/packet/PacketApiMetadata.java
andreaturli: it should simple say `project id`, agreed

packet/src/main/java/org/jclouds/packet/PacketApiMetadata.java
andreaturli: ok

packet/src/main/java/org/jclouds/packet/PacketProviderMetadata.java
andreaturli: yes, good idea, they have NJ, CA, NL and JP

packet/src/main/java/org/jclouds/packet/config/PacketHttpApiModule.java
andreaturli: ok, I've seen both patterns in jclouds, I will move that to ApiMetadata

packet/src/main/java/org/jclouds/packet/config/PacketHttpApiModule.java
andreaturli: ok will change it

packet/src/main/java/org/jclouds/packet/config/PacketHttpApiModule.java
andreaturli: what about the LocationModule then?

nacx: That's what the method of the parent class installs. If you don't override it, it will already get installed.

andreaturli:
@nacx I'm adding pagination as well, agreed better sooner than later

re PublicKey I think it was a typo from my first PR, this provider is not using them in the domain objects.

packet/pom.xml
```diff
@@ -107,6 +107,11 @@
                 </exclusion>
             </exclusions>
         </dependency>
+        <dependency>
+            <groupId>org.apache.jclouds.labs</groupId>
+            <artifactId>cloudsigma2</artifactId>
+            <version>2.1.0-SNAPSHOT</version>
+        </dependency>
```
nacx: Remove this

packet/src/main/java/org/jclouds/packet/config/PacketHttpApiModule.java
```diff
@@ -50,12 +56,7 @@ protected void bindErrorHandlers() {
 
    @Override
    protected void bindRetryHandlers() {
-      bind(HttpRetryHandler.class).annotatedWith(ClientError.class).to(BackoffLimitedRetryHandler.class);
+      bind(HttpRetryHandler.class).annotatedWith(ServerError.class).to(BackoffLimitedRetryHandler.class);
```
nacx: This is [the default handler](https://github.com/jclouds/jclouds/blob/master/core/src/main/java/org/jclouds/http/handlers/DelegatingRetryHandler.java#L56) bound for server errors, so there is no need to override this method to set it.

packet/src/main/java/org/jclouds/packet/handlers/PacketErrorHandler.java
```diff
@@ -43,8 +43,6 @@ public void handleError(HttpCommand command, HttpResponse response) {
       message = message != null ? message : String.format("%s -> %s", command.getCurrentRequest().getRequestLine(),
               response.getStatusLine());
       switch (response.getStatusCode()) {
-         case 400:
-            break;
```
nacx: If we don't handle this, we will propagate a generic `HttpResponseException` for bad requests. The common pattern is to handle it and encapsulate it in an `IllealArgumentException`.

packet/src/test/java/org/jclouds/packet/features/ProjectApiMockTest.java
```diff
@@ -33,17 +31,17 @@
...
       server.enqueue(jsonResponse("/projects.json"));
 
-      List<Project> projects = api.projectApi().list();
+      Iterable<Project> projects = api.projectApi().list().concat();
 
-      assertEquals(size(projects), 1); 
+      assertEquals(size(projects), 1);
       assertEquals(server.getRequestCount(), 1);
       assertSent(server, "GET", "/projects");
    }
```
nacx: Change the test to something like [this](https://github.com/jclouds/jclouds/blob/master/providers/digitalocean2/src/test/java/org/jclouds/digitalocean2/features/ActionApiMockTest.java#L37-L48), to also verify that pagination works.

andreaturli:
rebuild please

nacx:
Nice! +1!

andreaturli:
ok thanks, merging!

andreaturli:
merged at [master](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/070fe653)

PR #343, opened 2017-01-12
add AddApiVersionToRequest filter

andreaturli:


packet/src/main/java/org/jclouds/packet/filters/AddApiVersionToRequest.java
```diff
@@ -0,0 +1,47 @@
...
+import static com.google.common.base.Preconditions.checkNotNull;
+
+@Singleton
+public class AddApiVersionToRequest implements HttpRequestFilter {
+
+    private final String apiVersion;
+
+    @Inject
+    public AddApiVersionToRequest(@ApiVersion String apiVersion) {
+        this.apiVersion = checkNotNull(apiVersion, "apiVersion");
```
nacx: Remove the public modifier from the constructor and remove the null check.

packet/src/main/java/org/jclouds/packet/filters/AddApiVersionToRequest.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    @Override
+    public HttpRequest filter(final HttpRequest request) throws HttpException {
+        return request.toBuilder()
+                .replaceHeader(HttpHeaders.ACCEPT, String.format("application/json; version=%s", apiVersion))
```
nacx: Do we want to override the Accept header, whatever it is? Or just append the version if missing? Both approaches LGTM.
If we go for the former, then we could remove the `@Consumes` annotations from the Api classes, since the header will be always put by this filter.

andreaturli: I've slightly modified the logic, it now gets the previous `accept` header, which must be there because of @Consume, and append the `version=<value>` to it

packet/src/test/java/org/jclouds/packet/compute/internal/BasePacketApiMockTest.java
```diff
@@ -130,15 +131,15 @@ protected RecordedRequest assertSent(MockWebServer server, String method, String
...
+      assertEquals(request.getHeader("Accept"), "application/json; version=" + ctx.getMetadata().get("apiVersion"));
       assertEquals(request.getHeader("X-Auth-Token"), X_AUTHORIZATION_TOKEN);
       return request;
    }
 
    protected RecordedRequest assertSent(MockWebServer server, String method, String path, String json)
          throws InterruptedException {
       RecordedRequest request = assertSent(server, method, path);
-      assertEquals(request.getHeader("Content-Type"), "application/json");
+      assertEquals(request.getHeader("Accept"), "application/json; version=" + ctx.getMetadata().get("apiVersion"));
```
nacx: This is already checked in the method called in the line above. This one is intended to verify requests with a body and thus it checks they have the right content-type header.

andreaturli: addressed and rebased, I think it is ready for being merged.

andreaturli:
@nacx happy with that version?

andreaturli:
merged at [master](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/dfd6804c)

PR #344, opened 2017-01-18
JCLOUDS-1208 Implement the TemplateOptions.inboundPorts in ProfitBricks REST

alibazlamit:
@nacx An implementation for [this](https://issues.apache.org/jira/browse/JCLOUDS-1208?filter=-3).

Test results [here](https://gist.github.com/alibazlamit/c57841f81af07f5275ca63cf8716e649).

Thanks


nacx:
Thanks @alibazlamit! I'll review the PR in a while. Take care and edit the gist to remove your credentials, and change them in the provider now that they have been exposed!

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/compute/ProfitBricksComputeServiceAdapter.java
```diff
@@ -276,20 +274,42 @@ public boolean apply(Lan input) {
...
+                 .nicId(nic.id())
+                 .name(server.properties().name() + " jclouds-firewall")
+                 .protocol(FirewallRule.Protocol.TCP)
+                 .portRangeStart(inboundPort)
+                 .portRangeEnd(inboundPort)
+                 .build()
+         );
+         trackables.waitUntilRequestCompleted(rule);
+
+      }
```
nacx: Better use the [ComputeServiceUtils. getPortRangesFromList](https://github.com/jclouds/jclouds/blob/master/compute/src/main/java/org/jclouds/compute/util/ComputeServiceUtils.java#L265-L281) to compute the ranges and create the minimum set of rules.

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/compute/ProfitBricksComputeServiceAdapter.java
```diff
@@ -276,20 +274,42 @@ public boolean apply(Lan input) {
          }
       }
 
+      List<FirewallRule> firewallrules = new ArrayList<FirewallRule>();
```
nacx: Unused?

profitbricks-rest/src/main/java/org/apache/jclouds/profitbricks/rest/compute/ProfitBricksComputeServiceAdapter.java
```diff
@@ -276,20 +274,42 @@ public boolean apply(Lan input) {
...
+         firewallActive = true;
+      }
+
       Nic nic = api.nicApi().create(Nic.Request.creatingBuilder()
               .dataCenterId(dataCenterId)
               .name("jclouds" + name)
               .dhcp(Boolean.TRUE)
               .lan(lanId)
-              .firewallActive(Boolean.FALSE)
+              .firewallActive(firewallActive)
```
nacx: Just `.firewallActive(inboundPorts.length > 0)` and remove the previous conditional.

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/compute/ProfitBricksComputeServiceLiveTest.java
```diff
@@ -49,12 +56,11 @@ protected Module getSshModule() {
    protected LoggingModule getLoggingModule() {
       return new SLF4JLoggingModule();
    }
-   
+
    @Override
    protected Iterable<Module> setupModules() {
       ImmutableSet.Builder<Module> modules = ImmutableSet.builder();
       modules.addAll(super.setupModules());
-      modules.add(new ProfitBricksRateLimitModule());
```
nacx: Why is this removed?

profitbricks-rest/src/test/java/org/apache/jclouds/profitbricks/rest/compute/ProfitBricksComputeServiceLiveTest.java
```diff
@@ -113,4 +119,28 @@ public void testCreateNodeWithCustomHardwareUsingMins() throws Exception {
...
+         for (FirewallRule rule : server.entities().nics().items().get(0).entities().firewallrules().items()) {
+            if (rule.properties().portRangeStart() == 80 || rule.properties().portRangeStart() == 22 || rule.properties().portRangeStart() == 443) {
+               matches++;
+            }
+         }
+         Assert.assertEquals(3, matches);
+      } finally {
+         client.destroyNodesMatching(inGroup(group + "inbound"));
+      }
+   }
```
nacx: The `createAndRunAServiceInGroup`method of the parent class already configures a couple inbound ports for some tests. You could override that method, call super and just add the firewall rule validations after. Then you could remove this additional test.

alibazlamit: Do you mean to call the super `testCreateAndRunAService` or the `createAndRunAServiceInGroup`, if i override the `createAndRunAServiceInGroup` i still have to create an additional test since its not a test its just a method?

nacx: I'd just override the `createAndRunAServiceInGroup`, call super and add the validations after that. In the end, it is not a test itself, but the method configures a node with ports 22 and 8080 as inbound ports. Adding the validations after that makes sense, and a failure will produce a proper stacktrace and context.

In the end, we want to test as much as possible in the less time, but having some decent feedback. I think this approach meets both objectives.

alibazlamit:
@nacx updated with the changes

nacx:
Squashed and pushed to master as [b083f4c2](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/b083f4c2). Thanks @alibazlamit!

nacx:
Backported to [2.0.x](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/53439188) too, since it does not introduce backwards incompatible changes.

PR #345, opened 2017-01-18
add remaining features to support the abstraction

andreaturli:
- add DeviceApi with Mock and Live Test
- add FacilityApi with Mock and Live Test
- add OperatingSystemApi with Mock and Live Test
- add SshKeyApi with Mock and Live Test

---

Notice this PR adds many files but because we carefully reviewed `ProjectApi` in #340  and I copied the same structure for these APIs I expect that to be quite easy to review

### DIFF
--- a/packet/src/main/java/org/jclouds/packet/domain/Device.java
+++ b/packet/src/main/java/org/jclouds/packet/domain/Device.java
@@ -18,6 +18,8 @@ package org.jclouds.packet.domain;
 
 import java.util.Date;
 import java.util.List;
+import java.util.Map;
+import java.util.Set;
 
 import org.jclouds.javax.annotation.Nullable;
 import org.jclouds.json.SerializedNames;
@@ -27,6 +29,8 @@ import com.google.common.base.Enums;
 import com.google.common.base.Joiner;
 import com.google.common.base.Optional;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
 
 import static com.google.common.base.Preconditions.checkArgument;
 
@@ -60,15 +64,19 @@ public abstract class Device {
     public abstract OperatingSystem operatingSystem();
     public abstract Facility facility();
     public abstract Href project();
+    public abstract List<Href> sshKeys();
     public abstract Href projectLite();
     public abstract List<Object> volumes();
     public abstract List<IpAddress> ipAddresses();
+    public abstract List<ProvisioningEvent> provisioningEvents();
     public abstract Plan plan();
     public abstract String rootPassword();
     public abstract String userdata();
     public abstract String href();
 
-    @SerializedNames({"id", "short_id", "hostname", "description", "state", "tags", "billing_cycle", "user", "iqn", "locked", "bonding_mode", "created_at", "updated_at", "operating_system", "facility", "project", "project_lite", "volumes", "ip_addresses", "plan", "root_password", "userdata", "href"})
+    @SerializedNames({"id", "short_id", "hostname", "description", "state", "tags", "billing_cycle", "user", "iqn",
+            "locked", "bonding_mode", "created_at", "updated_at", "operating_system", "facility", "project", "ssh_keys",
+            "project_lite", "volumes", "ip_addresses", "provisioning_events", "plan", "root_password", "userdata", "href"})
     public static Device create(String id,
                                 String shortId,
                                 String hostname,
@@ -85,9 +93,11 @@ public abstract class Device {
                                 OperatingSystem operatingSystem,
                                 Facility facility,
                                 Href project,
+                                List<Href> sshKeys,
                                 Href projectLite,
                                 List<Object> volumes,
                                 List<IpAddress> ipAddresses,
+                                List<ProvisioningEvent> provisioningEvents,
                                 Plan plan,
                                 String rootPassword,
                                 String userdata,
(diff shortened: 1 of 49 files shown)


packet/src/main/java/org/jclouds/packet/compute/utils/URIs.java
```diff
@@ -0,0 +1,30 @@
...
+import java.net.URI;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+public class URIs {
+
+    public static String toId(URI uri) {
+        checkNotNull(uri, "uri");
+        String path = uri.getPath();
+        return path.substring(path.lastIndexOf('/') + 1);
```
nacx: What about URIs Just like `http://packet.net`? Let's better validate the input to propagate an appropriate error and add a unit test class for this.

packet/src/main/java/org/jclouds/packet/features/DeviceApi.java
```diff
@@ -0,0 +1,145 @@
...
+
+    @Named("device:create")
+    @POST
+    @Produces(MediaType.APPLICATION_JSON)
+    @MapBinder(BindToJsonPayload.class)
+    @ResponseParser(URIParser.class)
+    URI create(@PayloadParam("hostname") String hostname, @PayloadParam("plan") String plan,
+               @PayloadParam("billing_cycle") String billingCycle, @PayloadParam("facility") String facility,
+               @PayloadParam("features") Map<String, String> features, @PayloadParam("operating_system") String operatingSystem,
+               @PayloadParam("locked") boolean locked, @PayloadParam("userdata") String userdata, @PayloadParam("tags") Set<String> tags);
```
nacx: For methods with such an amount of parameters, consider creating an object instead.

packet/src/main/java/org/jclouds/packet/features/DeviceApi.java
```diff
@@ -0,0 +1,145 @@
...
+        private static class ToPagedIterable extends BaseToPagedIterable<Device, ListOptions> {
+
+            @Inject ToPagedIterable(PacketApi api, Function<Href, ListOptions> linkToOptions) {
+                super(api, linkToOptions);
+            }
+
+           @Override
+           protected List<Object> getArgs(GeneratedHttpRequest request) {
+              return request.getCaller().get().getArgs();
+           }
```
nacx: Do we need to override this?

packet/src/main/java/org/jclouds/packet/features/SshKeyApi.java
```diff
@@ -0,0 +1,125 @@
...
+import com.google.common.base.Function;
+import com.google.common.base.Optional;
+import com.google.inject.TypeLiteral;
+
+@Path("/ssh-keys")
+@Consumes(MediaType.APPLICATION_JSON)
+@RequestFilters({ AddXAuthTokenToRequest.class, AddApiVersionToRequest.class} )
+public interface SshKeyApi {
+
+    @Named("sshKey:list")
```
nacx: Use lowercase naming like in the other apis.

packet/src/main/java/org/jclouds/packet/features/DeviceApi.java
```diff
@@ -0,0 +1,145 @@
...
+                return api.deviceApi((String) arg0.get()).list(options);
+            }
+        }
+    }
+
+    @Named("device:create")
+    @POST
+    @Produces(MediaType.APPLICATION_JSON)
+    @MapBinder(BindToJsonPayload.class)
+    @ResponseParser(URIParser.class)
```
nacx: Use the existing [ParseURIFromListOrLocationHeaderIf20x](https://github.com/jclouds/jclouds/blob/master/core/src/main/java/org/jclouds/http/functions/ParseURIFromListOrLocationHeaderIf20x.java) instead?

packet/src/test/java/org/jclouds/packet/features/DeviceApiMockTest.java
```diff
@@ -0,0 +1,78 @@
...
+   public void testListDevicesWithOptionsReturns404() throws InterruptedException {
+      server.enqueue(response404());
+
+      Iterable<Device> actions = api.deviceApi("93907f48-adfe-43ed-ad89-0e6e83721a54").list(page(1).perPage(5));
+
+      assertTrue(isEmpty(actions));
+
+      assertEquals(server.getRequestCount(), 1);
+      assertSent(server, "GET", "/projects/93907f48-adfe-43ed-ad89-0e6e83721a54/devices?page=1&per_page=5");
+   }
```
nacx: Add the mock tests for the remaining methods in the DeviceApi.

packet/src/test/java/org/jclouds/packet/features/SshKeyApiLiveTest.java
```diff
@@ -0,0 +1,49 @@
...
+      final AtomicInteger found = new AtomicInteger(0);
+      assertTrue(Iterables.all(api().list().concat(), new Predicate<SshKey>() {
+         @Override
+         public boolean apply(SshKey input) {
+            found.incrementAndGet();
+            return !isNullOrEmpty(input.id());
+         }
+      }), "All ssh keys must have the 'id' field populated");
+      assertTrue(found.get() > 0, "Expected some ssh keys to be returned");
+   }
```
nacx: Add the live tests for the create, get and delete operations.

packet/src/test/java/org/jclouds/packet/features/SshKeyApiMockTest.java
```diff
@@ -0,0 +1,78 @@
...
+   public void testListSshKeysWithOptionsReturns404() throws InterruptedException {
+      server.enqueue(response404());
+
+      Iterable<SshKey> actions = api.sshKeyApi().list(page(1).perPage(5));
+
+      assertTrue(isEmpty(actions));
+
+      assertEquals(server.getRequestCount(), 1);
+      assertSent(server, "GET", "/ssh-keys?page=1&per_page=5");
+   }
```
nacx: Add the tests for the remaining methods in the api.

packet/src/main/java/org/jclouds/packet/features/DeviceApi.java
andreaturli: I think so, but happy to accept a more elegant solution

packet/src/main/java/org/jclouds/packet/features/DeviceApi.java
andreaturli: much better, will test that

packet/src/main/java/org/jclouds/packet/features/DeviceApi.java
nacx: Why it is not needed in the other similar classes? I think we can remove the override.

andreaturli: because deviceapi is the the only that requires an arg `@Path("/projects/{projectId}/devices")` without override arg0 is null

nacx: Oh, I see. It is a nested API that needs the args from the caller, not from their own methods. In that case, this function should better extend [Arg0ToPagedIterable.FromCaller](https://github.com/jclouds/jclouds/blob/master/core/src/main/java/org/jclouds/collect/internal/Arg0ToPagedIterable.java#L53), which is intended for nested APIs.

packet/src/main/java/org/jclouds/packet/domain/Device.java
```diff
@@ -105,4 +109,67 @@ public static Device create(String id,
...
+        public abstract String facility();
+        public abstract Map<String, String> features();
+        public abstract String operatingSystem();
+        public abstract Boolean locked();
+        public abstract String userdata();
+        public abstract Set<String> tags();
+
+        @SerializedNames({"hostname", "plan", "billing_cycle", "facility", "features", "operating_system",
+                "locked", "userdata", "tags" })
+        public static CreateDevice create(final String hostname, final String plan, final String billingCycle,
```
nacx: [minor] Make this private to enforce the use of the builder?

packet/src/test/java/org/jclouds/packet/features/SshKeyApiMockTest.java
nacx: Still missing?

packet/src/test/java/org/jclouds/packet/features/DeviceApiMockTest.java
nacx: Still missing?

andreaturli: ah sorry added only liveTests, will do asap

andreaturli:
I've found that they have changed the createDevice and createSsh API that now returns domain objects rather than URI, so even easier!

packet/src/main/java/org/jclouds/packet/features/DeviceApi.java
```diff
@@ -64,91 +60,91 @@
...
+      }
+
+      public static class ToPagedIterable extends Arg0ToPagedIterable.FromCaller<Device, ToPagedIterable> {
+
+         private final PacketApi api;
+         private final Function<Href, ListOptions> hrefToOptions;
+
+         @Inject
+         protected ToPagedIterable(PacketApi api, Function<Href, ListOptions> hrefToOptions) {
+            this.api = checkNotNull(api, "api");
```
nacx: Remove the redundant check and the protected modifier from the constructor

packet/src/test/java/org/jclouds/packet/features/DeviceApiLiveTest.java
```diff
@@ -73,7 +72,7 @@ public void testCreate() {
                       .tags(ImmutableSet.<String> of())
                       .build()
       );
-      deviceId = URIs.toId(deviceCreated);
+      deviceId = deviceCreated.id();
       assertNodeRunning(deviceId);
       Device device = api().get(deviceId);
       assertNotNull(device, "Device must not be null");
```
nacx: Add a test for the `actions` method too?

packet/src/test/java/org/jclouds/packet/features/DeviceApiMockTest.java
```diff
@@ -75,4 +81,68 @@ public void testListDevicesWithOptionsReturns404() throws InterruptedException {
...
+
+   public void testDeleteDeviceReturns404() throws InterruptedException {
+      server.enqueue(response404());
+
+      api.deviceApi("93907f48-adfe-43ed-ad89-0e6e83721a54").delete("1");
+
+      assertEquals(server.getRequestCount(), 1);
+      assertSent(server, "DELETE", "/projects/93907f48-adfe-43ed-ad89-0e6e83721a54/devices/1");
+   }
+
```
nacx: Still missing mock tests for the `actions` method?

nacx:
Thanks @andreaturli! Just some final comments. Almost there! :)

packet/src/main/java/org/jclouds/packet/features/DeviceApi.java
```diff
@@ -143,7 +143,7 @@ protected ToPagedIterable(PacketApi api, Function<Href, ListOptions> hrefToOptio
 
    @Named("device:actions")
    @POST
-   @Path("/devices/{id}/actions")
+   @Path("/{id}/actions")
```
nacx: The importance of having all tests! One bug spotted :)

nacx:
There is one checkstyle violation to fix before merging:
https://jclouds.ci.cloudbees.com/job/jclouds-labs-pull-requests/1622/org.apache.jclouds.labs$packet/violations/file/src/main/java/org/jclouds/packet/features/DeviceApi.java/

andreaturli:
squashing and merging

andreaturli:
Merged at [master](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/f749c88e)

PR #346, opened 2017-01-24
add compute functions

andreaturli:
- fix Device and PaginatedCollection

### DIFF
--- a/packet/src/main/java/org/jclouds/packet/compute/functions/DeviceToNodeMetadata.java
+++ b/packet/src/main/java/org/jclouds/packet/compute/functions/DeviceToNodeMetadata.java
@@ -1,0 +1,110 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jclouds.packet.compute.functions;
+
+import java.util.List;
+
+import javax.annotation.Resource;
+import javax.inject.Inject;
+import javax.inject.Named;
+import javax.inject.Singleton;
+
+import org.jclouds.compute.domain.NodeMetadata;
+import org.jclouds.compute.domain.NodeMetadataBuilder;
+import org.jclouds.compute.functions.GroupNamingConvention;
+import org.jclouds.compute.reference.ComputeServiceConstants;
+import org.jclouds.logging.Logger;
+import org.jclouds.packet.domain.Device;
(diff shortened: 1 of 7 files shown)


packet/src/main/java/org/jclouds/packet/compute/functions/DeviceToNodeMetadata.java
```diff
@@ -0,0 +1,112 @@
...
+    private final OperatingSystemToImage operatingSystemToImage;
+    private final FacilityToLocation facilityToLocation;
+    private final Function<Device.State, NodeMetadata.Status> toPortableStatus;
+    private final GroupNamingConvention groupNamingConvention;
+
+    @Inject
+    DeviceToNodeMetadata(PlanToHardware planToHardware, OperatingSystemToImage operatingSystemToImage, FacilityToLocation facilityToLocation,
+                         Function<Device.State, NodeMetadata.Status> toPortableStatus,
+                         GroupNamingConvention.Factory groupNamingConvention) {
+        this.planToHardware = checkNotNull(planToHardware, "planToHardware cannot be null");
```
nacx: No need for null checks in the injection constructors.

packet/src/main/java/org/jclouds/packet/compute/functions/FacilityToLocation.java
```diff
@@ -0,0 +1,55 @@
...
+        this.justProvider = justProvider;
+    }
+
+    @Override
+    public Location apply(final Facility facility) {
+        final LocationBuilder builder = new LocationBuilder();
+        builder.id(facility.code());
+        builder.description(facility.name());
+        builder.parent(getOnlyElement(justProvider.get()));
+        builder.scope(LocationScope.REGION);
```
nacx: Can the ISO-3166 code be inferred in a consistent way, for example from the `state` field?

packet/src/main/java/org/jclouds/packet/compute/functions/PlanToHardware.java
```diff
@@ -0,0 +1,86 @@
...
+                .name(plan.name())
+                .hypervisor("none")
+                .processors(getProcessors(plan))
+                .ram(getMemory(plan))
+                .volumes(getVolumes(plan));
+        return builder.build();
+    }
+
+    private Integer getMemory(Plan plan) {
+        if (plan.specs() == null || plan.specs().drives() == null) return -1;
```
nacx: Better return 0 ?

packet/src/main/java/org/jclouds/packet/compute/functions/FacilityToLocation.java
andreaturli: it'd be a bit tricky as the API returns https://github.com/jclouds/jclouds-labs/blob/master/packet/src/test/resources/facilities.json and it is not straightforward to get the correct ISO-3166 from the `name` and/or `code` I think

nacx: I was referring to the `state` field (according to their docs). Are there new fields that have been added recently on their side?
https://www.packet.net/developers/api/devices/#devices-locations

andreaturli: I think the docs is not in sync: there is no `state` in the live json object :(

nacx: Let's leave it as-is then. Thanks for clarifying!

andreaturli:
squashing and merging

andreaturli:
merged at [master](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/a91ff3bd)

PR #351, opened 2017-01-27
[azurecompute-arm] performance improvements misc

andreaturli:
- improve cleanup performances: by using resourceGroupApi.resources
- improve list images performance: by avoiding to get image details

/cc @nacx

### DIFF
--- a/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
+++ b/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
@@ -16,20 +16,6 @@
  */
 package org.jclouds.azurecompute.arm.compute;
 
-import static com.google.common.base.Preconditions.checkState;
-import static com.google.common.collect.Iterables.contains;
-import static com.google.common.collect.Iterables.filter;
-import static com.google.common.collect.Iterables.find;
-import static com.google.common.collect.Iterables.getOnlyElement;
-import static org.jclouds.azurecompute.arm.compute.extensions.AzureComputeImageExtension.CONTAINER_NAME;
-import static org.jclouds.azurecompute.arm.compute.extensions.AzureComputeImageExtension.CUSTOM_IMAGE_OFFER;
-import static org.jclouds.azurecompute.arm.compute.functions.VMImageToImage.decodeFieldsFromUniqueId;
-import static org.jclouds.azurecompute.arm.compute.functions.VMImageToImage.encodeFieldsToUniqueIdCustom;
-import static org.jclouds.azurecompute.arm.compute.functions.VMImageToImage.getMarketplacePlanFromImageMetadata;
-import static org.jclouds.azurecompute.arm.config.AzureComputeProperties.IMAGE_PUBLISHERS;
-import static org.jclouds.compute.util.ComputeServiceUtils.metadataAndTagsAsCommaDelimitedValue;
-import static org.jclouds.util.Closeables2.closeQuietly;
-
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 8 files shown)


azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/Identity.java
```diff
@@ -0,0 +1,34 @@
...
+ * limitations under the License.
+ */
+package org.jclouds.azurecompute.arm.domain;
+
+import org.jclouds.json.SerializedNames;
+
+import com.google.auto.value.AutoValue;
+
+@AutoValue
+public abstract class Identity {
```
nacx: Is there any reason why this class is not an inner class of Resource?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/features/ResourceGroupApi.java
```diff
@@ -76,12 +78,20 @@
    @Nullable
    ResourceGroup get(@PathParam("name") String name);
 
+   @Named("resourcegroup:resources")
+   @GET
+   @Path("/{name}/resources")
+   @SelectJson("value")
+   @Fallback(EmptyListOnNotFoundOr404.class)
+   @Nullable
```
nacx: Given the configured fallback the returned link is never going to be null.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/Identity.java
andreaturli: not really, I will make it an inner class

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/features/ResourceGroupApi.java
andreaturli: right!

nacx:
@andreaturli Having a deeper look I've noticed we actually need the version properties if we want to be able to support images from the marketplace. We need to know the marketplace information from the image (done [here]( https://github.com/jclouds/jclouds-labs/blob/master/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/functions/VMImageToImage.java#L137-L145)).

Without that, deploying a node from an image from the marketplace will fail. VMs from the market place need to configure the *plan*, as done [here](https://github.com/jclouds/jclouds-labs/blob/master/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java#L162-L165).

Given that we already have an image cache and impact when listing images should be constrained to the first time they are looked up, I'd say it is more important to keep the ability to deploy marketplace images than improving the image list response time.

andreaturli:
I think it makes sense, I will revert that second commit

andreaturli:
rebuild please

andreaturli:
@nacx is it ok?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
```diff
@@ -200,8 +200,8 @@
             for (Version version : versionList) {
                Version versionDetails = osImageApi.getVersion(publisherName, offer.name(), sku.name(), version.name());
                VMImage vmImage = VMImage.azureImage().publisher(publisherName).offer(offer.name()).sku(sku.name())
-                     .version(versionDetails.name()).location(location).versionProperties(versionDetails.properties())
-                     .build();
+                       .version(versionDetails.name()).location(location).versionProperties(versionDetails.properties())
+                       .build();
                osImagesRef.add(vmImage);
             }
          }
```
nacx: Discard the changes in this class.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/functions/VMImageToImage.java
```diff
@@ -176,10 +176,21 @@ public static Plan getMarketplacePlanFromImageMetadata(Image image) {
...
+               
+//               Version versionDetails = osImageApi.getVersion(publisherName, offer.name(), sku.name(), version.name());
+//                              VMImage vmImage = VMImage.azureImage().publisher(publisherName).offer(offer.name()).sku(sku.name())
+//                                    .version(versionDetails.name()).location(location).versionProperties(versionDetails.properties())
+//                                    .build();
+
+//               family = OsFamily.fromValue(image.versionProperties().osDiskImage().operatingSystem().toUpperCase());
             }
 
             // only 64bit OS images are supported by Azure ARM
```
nacx: Discard the changes in this class.

andreaturli: ops

nacx:
Pushed to [master](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/6b35f6fd) and [2.0.x](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/a5a4f30b). Thanks!

PR #353, opened 2017-01-27
Inspect all known OSFamilies

nacx:
Inspect all known OS families when trying to determine the OSFamily of an image. It also performs a case-insensitive comparison now, so we'll have more chances to get images with the proper OS.

/cc @andreaturli


nacx:
Pushed to [master](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/64bdc801) and [2.0.x](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/84f31cb8).

PR #359, opened 2017-02-08
Improve image lookup when getting node info

nacx:
Listing images in azure is **very** expensive, as it needs to list the publishers, offers, SKUs and versions.

The function that transforms a VM to the jclouds Node, uses a cached image map (populated with the contents of the image cache) to get the details of the image corresponding to the node. The problem is that when listing nodes (or getting a single one), if the image cache has not yet been populated, a call to the `listImages` method is done. This is likely to be a performance killer in many situations where users might only be interested in nodes.

This PR changes this behavior to just call the `getImage` method to get the details of the VM image. This call is not cached, though, and it means that when listing N nodes, there will be N calls to get the image. This call, however, is not expensive, and I believe in most cases this approach will perform better than having to get the entire image list.

Also take into account that the image cache is destroyed when a context is destroyed, so the use cases where users need to create/destroy the contexts more often than desired would perform bad too in the current approach.

I still have to test this and compare several runs of the compute service live tests with and without this change, but @andreaturli, @aledsage I'd love to have your feedback on this.

(This PR also refactors some logic in the function to other classes and removes redundant code).


neykov:
LGTM. Could still use the image cache and fall back to `getImage` only if empty.

nacx:
Thanks @neykov! I've applied the change. The ImageCache already provides [a method](https://github.com/jclouds/jclouds/blob/master/compute/src/main/java/org/jclouds/compute/suppliers/ImageCacheSupplier.java#L152-L165) to do that, so it's been easy. That method directly uses the underlying cache, without initializing the memoized supplier that fetches all images. Calling there should do what we need here.
Feedback welcome!

andreaturli:
+1 looks like a desired improvement.
not yet tested though

andreaturli:
@nacx looks like the failures are valid

https://jclouds.ci.cloudbees.com/job/jclouds-labs-pull-requests/1694/org.apache.jclouds.labs$azurecompute-arm/testReport/junit/org.jclouds.azurecompute.arm.features/SubnetApiMockTest/start/

can you fix them so that we can merge it?

nacx:
Oops, I build locally the "classic" Azure provider! It's fixed now. The ImageCacheSupplier is intentionally not marked as "injectable", to avoid people relying on internal classes. It is safe to inject the image supplier and check its type.

nacx:
Squashed and pushed to [master](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/cd48a2eb) and [2.0.x](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/40050076). Thanks for having a loog @neykov @andreaturli!

andreaturli:
cool! thanks @nacx

PR #366, opened 2017-03-01
Remove all sonatype snapshot repositories

nacx:


b2/pom.xml
```diff
@@ -41,20 +41,6 @@
...
-      <id>apache-snapshots</id>
-      <url>https://repository.apache.org/content/repositories/snapshots</url>
-      <releases>
-        <enabled>false</enabled>
-      </releases>
-      <snapshots>
-        <enabled>true</enabled>
-      </snapshots>
-    </pluginRepository>
-  </pluginRepositories>
```
demobox: @nacx: I'm assuming we're basing this change on the idea that users have configured the Apache snapshots repo in their settings, as described in http://jclouds.apache.org/start/install/?

nacx: This pom inherits the labs root pom, which inherits from the jclouds-project one, so it should already have the plugin repo configured.

nacx:
Pushed to master as [4d80bd7a](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/4d80bd7a).

PR #377, opened 2017-03-24
Adding Dimension Data Infrastructure API

gatj98:
The next pull request from DD, this one adds the infrastructure API


gatj98:
rebuild please

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/filters/DatacenterIdListDatacentersFilter.java
```diff
@@ -0,0 +1,45 @@
...
+
+import javax.inject.Inject;
+import java.util.Set;
+
+/**
+ * Adds set of Datacenter IDs as set in jclouds.zones JVM property.
+ */
+public class DatacenterIdListDatacentersFilter implements HttpRequestFilter {
+   @Inject
+   @Zone
```
nacx: Locations in jclouds are defined in a hierarchy such as: Provider (unique; DimensionData) > Region > Zone.
Are we going to model the locations as zones, and a Datacenter will be one zone? In that case, what is going to be a region? Otherwise change the annotation to `@Region` and adopt that if it is more appropriate (just asking).

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/parsers/ParseDatacenters.java
```diff
@@ -0,0 +1,62 @@
...
+
+import javax.inject.Inject;
+import javax.inject.Singleton;
+import java.util.List;
+
+@Singleton
+public class ParseDatacenters extends ParseJson<Datacenters> {
+
+    @Inject
+    public ParseDatacenters(Json json) {
```
nacx: Remove the `public` modifier from all injection constructors.

Also, move all these parser classes that support the pagination iterables as inner classes of the API interface. It's the preferred approach now, to limit their visibility.

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/parsers/ParseOperatingSystems.java
```diff
@@ -0,0 +1,63 @@
...
+            this.api = api;
+        }
+
+        @Override
+        protected Function<Object, IterableWithMarker<OperatingSystem>> markerToNextForArgs(List<Object> args) {
+            return new Function<Object, IterableWithMarker<OperatingSystem>>() {
+                @Override
+                public IterableWithMarker<OperatingSystem> apply(Object input) {
+                    PaginationOptions paginationOptions = PaginationOptions.class.cast(input);
+                    return api.getInfrastructureApi().listOperatingSystems(getArgs(request).get(0).toString(), paginationOptions);
```
nacx: No need for `getArgs(request)`; the `args` parameter in the method is already that; just use it.

nacx: BTW, since in most cases, pagination methods take one argument (we've found it common in many APIs) jclouds provides the convenience class `Arg0ToPagedIterable`. It is the same but instead of getting the list of args, you just get one optional arg parameter (the first one passed to the original method). Perhaps you could consider extending that class if it is more convenient, but not a big deal :)

dimensiondata/src/test/java/org/jclouds/dimensiondata/cloudcontrol/features/InfrastructureApiLiveTest.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.List;
+
+import static org.testng.Assert.assertNotNull;
+
+@Test(groups = "live", testName = "InfrastructureApiLiveTest", singleThreaded = true)
+public class InfrastructureApiLiveTest extends BaseDimensionDataCloudControlApiLiveTest {
+
+    @Test
+    public void testListDatacenters() {
+        List<Datacenter> datacenters = api().listDatacenters().concat().toList();
```
nacx: Better not call `toList()`. That will copy all elements into a list, losing the benefits of lazy loading each page. Not a big deal in the tests, but let's use pagination properly (this also serves as example for users).

dimensiondata/src/test/java/org/jclouds/dimensiondata/cloudcontrol/features/InfrastructureApiMockTest.java
```diff
@@ -0,0 +1,114 @@
...
+
+      Uris.UriBuilder uriBuilder = Uris
+            .uriBuilder("/caas/2.4/6ac1e746-b1ea-4da5-a24e-caf1a978789d/infrastructure/datacenter");
+      uriBuilder.addQuery(paginationOptions.buildQueryParameters());
+      assertSent(HttpMethod.GET, uriBuilder.toString());
+   }
+
+   public void testListDatacenters404() {
+      server.enqueue(response404());
+      assertTrue(api.getInfrastructureApi().listDatacenters().concat().isEmpty());
```
nacx: Also verify the sent request and that there is only on request generated.

dimensiondata/src/test/java/org/jclouds/dimensiondata/cloudcontrol/features/InfrastructureApiMockTest.java
```diff
@@ -0,0 +1,114 @@
...
+      Uris.UriBuilder uriBuilder = Uris
+            .uriBuilder("/caas/2.4/6ac1e746-b1ea-4da5-a24e-caf1a978789d/infrastructure/operatingSystem");
+      uriBuilder.addQuery("datacenterId", "NA9");
+      uriBuilder.addQuery(paginationOptions.buildQueryParameters());
+      assertSent(HttpMethod.GET, uriBuilder.toString());
+   }
+
+   public void testListOperatingSystems404() {
+      server.enqueue(response404());
+      assertTrue(api.getInfrastructureApi().listOperatingSystems("NA9").concat().isEmpty());
```
nacx: Also verify the generated request.

dimensiondata/src/test/java/org/jclouds/dimensiondata/cloudcontrol/features/InfrastructureApiMockTest.java
```diff
@@ -0,0 +1,114 @@
...
+   public void testListDatacenters() throws Exception {
+      server.enqueue(jsonResponse("/datacenters.json"));
+      Iterable<Datacenter> datacenters = api.getInfrastructureApi().listDatacenters().concat();
+
+      assertEquals(size(datacenters), 2); // Force the PagedIterable to advance
+      assertEquals(server.getRequestCount(), 2);
+
+      Uris.UriBuilder uriBuilder = Uris
+            .uriBuilder("/caas/2.4/6ac1e746-b1ea-4da5-a24e-caf1a978789d/infrastructure/datacenter");
+      Set<String> zones = ctx.utils().injector().getInstance(ZoneIdsSupplier.class).get();
```
nacx: Is there a value configured there for this? Let's make sure (perhaps in the base class) to configure some zones so this test properly verifies the behaviour.

dimensiondata/src/test/java/org/jclouds/dimensiondata/cloudcontrol/options/PaginationOptionsTest.java
```diff
@@ -0,0 +1,84 @@
...
+import com.google.common.collect.Multimap;
+import org.testng.annotations.BeforeMethod;
+import org.testng.annotations.Test;
+
+import java.util.Collection;
+
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertTrue;
+
+@Test(groups = "unit", testName = "PaginationOptionsTest", singleThreaded = true)
```
nacx: No need for this to be single-threaded. Most single-threaded tests are due to the fact that mockwebserver starts the server in a before method. Tests that don't use MWS don't need to be single-threaded in general.

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/filters/DatacenterIdListDatacentersFilter.java
gatj98: Good Q - we were having a discussion about this ourselves the other day. Your input would be appreciated @nacx!

DD has a number of 'Geos', each has it's own cloud control API endpoint, and consists of 1 or more datacenters. At present we see our provider as being for a single Geo (as it's the simplest thing to do to start with), but maybe we should aim to have a single provider to cover them all in time. There is no global API endpoint, so when creating a compute service instance we'd have to call out to multiple Geos to find out where the user is based (their home Geo) and which 'shaddow' Geos they also have access to. We do have an API that can be called on any Geo to find a users home Geo that does this fan-out, but I'm not sure if it's publicly published or not.

Anyway, to answer your Q I think the mapping should be:
Region -> Geo (only 1 for now - we prob need to do something to model this?)
Zone -> Datacenter

Does that sound ok?

dimensiondata/src/test/java/org/jclouds/dimensiondata/cloudcontrol/options/PaginationOptionsTest.java
gatj98: Initially it wasn't single threaded, but push request build failed for one of the tests at line 81 where the size of the value collection is checked. I figured this was because the threads were accessing the paginationOptions field at the same time. 

Given it's such a quick test didn't really see the benefit of multi threading so I made it single threaded rather than putting 'PaginationOptions paginationOptions = new PaginationOptions();' into each test method and ditching the instance fields. Will change it though

dimensiondata/src/test/java/org/jclouds/dimensiondata/cloudcontrol/features/InfrastructureApiMockTest.java
gatj98: :+1:

dimensiondata/src/test/java/org/jclouds/dimensiondata/cloudcontrol/features/InfrastructureApiMockTest.java
gatj98: 👍

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/filters/DatacenterIdListDatacentersFilter.java
nacx: Makes sense. In that case, take into account that the default Region/Zone suppliers (those bound with the `@Region` and `@Zone` qualifiers) take the values from configuration properties (as in [ProfitBricks](https://github.com/jclouds/jclouds/blob/master/providers/profitbricks/src/main/java/org/jclouds/profitbricks/ProfitBricksProviderMetadata.java#L64-L74).
In this case, I think it would make sense to configure all the regions via properties (including their endpoints with the `jclouds.region.<regionname>.endpoint` property) and then bind specific Zone suppliers to get them using the API, as done in [EC2](https://github.com/jclouds/jclouds/blob/master/apis/ec2/src/main/java/org/jclouds/ec2/config/BaseEC2HttpApiModule.java#L53-L57).
WDYT?

dimensiondata/src/test/java/org/jclouds/dimensiondata/cloudcontrol/options/PaginationOptionsTest.java
nacx: Instead of making it single-threaded, I'd rather declare the `PaginationOptions` in each method. It's a very simple object and won't introduce much overhead (and you're already *creating* it before every method, so it should be OK to move the declaration and creation to each test method instead of having a class variable).

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/parsers/ParseDatacenters.java
gatj98: :+1:

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/parsers/ParseOperatingSystems.java
gatj98: :+1:

gatj98:
Thanks for the comments @nacx - new commit with the changes (apart from region/zone stuff) made.
Looking at the region/zone stuff now - might be better if we do a separate pull request for that?

dimensiondata/src/test/java/org/jclouds/dimensiondata/cloudcontrol/features/InfrastructureApiMockTest.java
gatj98: Will look at this in conjunction with the region/zone config discussed above

nacx:
>Looking at the region/zone stuff now - might be better if we do a separate pull request for that?

Sure!

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/features/InfrastructureApi.java
```diff
@@ -76,4 +85,63 @@
...
+         }
+
+         @Override
+         protected Function<Object, IterableWithMarker<OperatingSystem>> markerToNextForArg0(
+               final Optional<Object> arg) {
+            return new Function<Object, IterableWithMarker<OperatingSystem>>() {
+               @Override
+               public IterableWithMarker<OperatingSystem> apply(Object input) {
+                  PaginationOptions paginationOptions = PaginationOptions.class.cast(input);
+                  return api.getInfrastructureApi().listOperatingSystems(arg.toString(), paginationOptions);
```
nacx: Take care! `arg.get().toString()`. Otherwise you're passing the `Optional` object.

nacx: This should have been spotted by the Mock test. Let's modify the mock tests that use pagination and verify *all* the requests being sent. The second one should contain the pagination parameters and we should also verify that request.

gatj98:
Hi @nacx 
Thanks for your comments - have reworked the pagination tests in InfrastructureApiMockTest, and saw it fail because of the bug you spotted with the optional.

Have added an IHookable in BaseDimensionDataCloudControlMockTest that checks after each test runs that we've assertted all of the requests received by the mock web server - hopefully this will help us write better tests going forward.

dimensiondata/src/test/java/org/jclouds/dimensiondata/cloudcontrol/features/InfrastructureApiMockTest.java
```diff
@@ -52,12 +51,12 @@ public void testListDatacentersWithPagination() throws Exception {
       server.enqueue(jsonResponse("/datacenters-page2.json"));
       Iterable<Datacenter> datacenters = api.getInfrastructureApi().listDatacenters().concat();
 
-      assertEquals(server.getRequestCount(), 2);
-      assertEquals(size(datacenters), 2);
+      assertEquals(server.getRequestCount(), 2); // 1 for account, 1 for page 1
+      assertEquals(size(datacenters), 2); // Causes all pages to be requested
       assertEquals(server.getRequestCount(), 3); // 1 for account details, 2 pages
```
gatj98: Am not 100% happy with this as I feel it needs code comments to explain what the test is doing.

Maybe a helper method of the form "consumeIterableAndAssertPagesRequested(Iterable iterable, int expectedExtraPagesRequested)" would be better?

nacx: Sounds good :)

nacx:
>Have added an IHookable in BaseDimensionDataCloudControlMockTest that checks after each test runs that we've assertted all of the requests received by the mock web server - hopefully this will help us write better tests going forward.

And this looks like a good pattern we can propagate to all other providers. Thanks!

dimensiondata/src/test/java/org/jclouds/dimensiondata/cloudcontrol/features/InfrastructureApiMockTest.java
gatj98: Cool - added that in latest commit

dimensiondata/src/test/java/org/jclouds/dimensiondata/cloudcontrol/internal/BaseAccountAwareCloudControlMockTest.java
```diff
@@ -0,0 +1,59 @@
...
+      server.enqueue(jsonResponse("/account.json"));
+   }
+
+   @BeforeMethod
+   public void reset() {
+      accountRetrieved = false;
+   }
+
+   protected Uris.UriBuilder addPageNumberToUriBuilder(Uris.UriBuilder uriBuilder, int pageNumber) {
+      uriBuilder.addQuery(PaginationOptions.Builder.pageNumber(pageNumber).buildQueryParameters());
```
nacx: It's better not to use the pagination class. I'd say it would be better to directly call `uriBuilder.addQuery("param-name", pageNumber)`, to catch any (accidental or not) issue if the pagination options class is changed in the future.

dimensiondata/src/test/java/org/jclouds/dimensiondata/cloudcontrol/internal/BaseDimensionDataCloudControlMockTest.java
```diff
@@ -68,10 +76,42 @@ public void start() throws IOException {
...
+            }
+            unAssertedRequestCount--;
+         }
+         fail(messageBuilder.toString());
+      }
    }
 
    @AfterMethod(alwaysRun = true)
-   public void stop() throws IOException {
+   public void stop(ITestContext testContext) throws Exception {
```
nacx: Is this parameter actually needed?

gatj98: Nuts - no it's not - put it in whilst trying something else and forgot it was there!

dimensiondata/src/test/java/org/jclouds/dimensiondata/cloudcontrol/internal/BaseAccountAwareCloudControlMockTest.java
gatj98: Fair point.
Actually this method should prob be on BaseDimensionDataCloudControlMockTest rather than this class

gatj98:
Thanks @nacx - will tweak those and squash as requested

gatj98:
Ok that's squashed now @nacx - slight false start with a checkstyle error, but should be sorted now :)

nacx:
Pushed to [master](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/84ebe3d3) and [2.0.x](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/42dc110f). Thanks @gatj98!

PR #379, opened 2017-03-30
Removed fixed versioning from the Mock tests. 

trevorflanagan:
We can use the DimensionDataCloudControlProviderMetadata to give us the CloudControl API version.

Removed caas from Api Path. In previous versions of CloudControl API there was alternative base paths, either 'oec' or 'caas', but this is no longer required in the implementation. All paths are caas so this can be configured  in the endpoint metadata.


nacx:
Merged to [master](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/95f274b8) and [2.0.x](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/9ffceea3). Thanks @trevorflanagan!

PR #381, opened 2017-04-13
Live tests are failing due to not including /caas in path. 

trevorflanagan:
This pull request is for the Mock tests and OrganisationIdFilter  changes before changing live tests.

I intend to create a proper test for @Zone in the next pull request.


dimensiondata/pom.xml
```diff
@@ -144,6 +144,7 @@
                                         <test.dimensiondata-cloudcontrol.credential>
                                             ${test.dimensiondata-cloudcontrol.credential}
                                         </test.dimensiondata-cloudcontrol.credential>
+                                        <jclouds.zones>LAB04_N2_VMWARE_1</jclouds.zones>
```
trevorflanagan: we have working mock and live tests against zones now

nacx: We'll probably have to have a closer look at this when implementing the compute service, but it is OK for now.

Just FTR, but not to have it fixed now, the compute service has the `listAssignableLocations` method that returns the locations (zones, as per previous discussions) where users can deploy VMs. jclouds will then cache those locations in a memoized supplier. Having HTTP filters such as the DataCenterId filter that have location suppliers injected, may cause infinite loops if the `listAssignableLocations` method in the compute service calls api methods with those filters.

This is not something we need to fix now, but to keep in mind when implementing the compute service, and for a possible upcoming discussion about how to properly model the locations for this provider.

trevorflanagan:
thanks @nacx for taking look at this PR. I have squashed the commit and noted your comments re `listAssignableLocations`

nacx:
Pushed to [master](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/92b548c2) and [2.0.x](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/de26ec2c)!

PR #383, opened 2017-04-20
Update README.md

andreaturli:


packet/README.md
```diff
@@ -1 +1,6 @@
 # Apache jclouds Packet (packet.net) provider
+
+
+## Pre-requisites
+
+If you are using Oracle JDK you want to install JCE following the [official documentation](http://www.oracle.com/technetwork/java/javase/downloads/jce8-download-2133166.html).
```
nacx: Out of curiosity, why are the strong encryption extensions needed for Packet?

andreaturli: not an expert, but it is useful to solve
```
Caused by: javax.net.ssl.SSLHandshakeException: Received fatal alert: handshake_failure
```
when calling their https endpoints

nacx: Ah, thx. Worth mentioning in the README the reason why this is required?

nacx:
Thx!

andreaturli:
Merged [master](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/8c4d97fb) and back-ported at [2.0.x](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/9a22e808)

nacx:
I've just seen our CI live tests for Packet fail for this reason. We'll have to configure the build to install the JCE policy files. I'm quite out of cycles for this now. @demobox do you know if it is possible to setup that in our CI?

andreaturli:
if our CI uses openjdk that shouldn't be required, I think

nacx:
Same failure with the OpenJDK:

```bash
[jclouds-live-tests-packet] $ /bin/bash /tmp/hudson9028556270177564970.sh
Apache Maven 3.3.3 (7994120775791599e205a5524ec3e0dfe41d4a06; 2015-04-22T07:57:37-04:00)
Maven home: /opt/maven/latest
Java version: 1.7.0_60-ea, vendor: Oracle Corporation
Java home: /opt/java/sdk/OpenJDK/jdk1.7.0_60/jre
Default locale: en_US, platform encoding: UTF-8
OS name: "linux", version: "3.13.0-110-generic", arch: "amd64", family: "unix"
```

nacx:
I've configured the CI with the `workaround #1` mentioned here and looks like it is working:
https://support.cloudbees.com/hc/en-us/articles/204911534-How-to-use-Java-Cryptography-Extension-JCE-Unlimited-Strength-Jurisdiction-Policy-Files

PR #385, opened 2017-04-24
JCLOUDS-1273: Support multiple resource groups in ARM

nacx:
https://issues.apache.org/jira/browse/JCLOUDS-1273

The compute service live tests pass (there is one transient failure unrelated to this change):

```bash
Failed tests: 
  AzureComputeServiceLiveTest>BaseComputeServiceLiveTest.testOptionToNotBlock:870 duration(37) longer than expected(30) seconds! 

Tests run: 27, Failures: 1, Errors: 0, Skipped: 0
Total time: 01:10 h (Wall Clock)
```

I still have some manual tests pending to make sure everything works as expected, but I'm submitting this for early feedback. The changes in this PR are:

* The ID returned for the nodes, security groups, etc, has been changed from `location/name` to `resourceGroup/name` to better reflect the path they have in the ARM API.
* Added helper methods to the `IdReference` class to extract the resource group and name from the object's URI. This helps hen getting resources by id and to build the APIs pointing to the right resource group (not the default one for the location).
* Methods that list resources (listNodes, listSecurityGroups, etc) have been changed to consider all existing resource groups.
* All transformation functions take into account the resource group of the current object instead of the default one (this was the root cause of the linked issue).
* The `resourceGroupMap` has been renamed to `defaultResourceGroup` and its usage has been removed from all classes. The only class using it is the SecurityGroupExtension, since there is no mean to specify the resource group where the security group should be created. They are created in the default one when using the extension.
* A `resourceGroup` method has been added to the `AzureTemplateOptions` to let users configure the resource group where all stuff will be created when creating nodes.

@andreaturli @danielestevez feedback and tests welcome!

### DIFF
--- a/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/strategy/CreateResourcesThenCreateNodes.java
+++ b/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/strategy/CreateResourcesThenCreateNodes.java
@@ -16,6 +16,12 @@
  */
 package org.jclouds.azurecompute.arm.compute.strategy;
 
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.collect.Iterables.getOnlyElement;
+import static org.jclouds.azurecompute.arm.config.AzureComputeProperties.DEFAULT_SUBNET_ADDRESS_PREFIX;
+import static org.jclouds.azurecompute.arm.config.AzureComputeProperties.DEFAULT_VNET_ADDRESS_SPACE_PREFIX;
+
 import java.util.Arrays;
 import java.util.Map;
 import java.util.Set;
@@ -27,12 +33,12 @@ import javax.inject.Singleton;
 
 import org.jclouds.Constants;
 import org.jclouds.azurecompute.arm.AzureComputeApi;
-import org.jclouds.azurecompute.arm.compute.domain.RegionAndIdAndIngressRules;
+import org.jclouds.azurecompute.arm.compute.domain.ResourceGroupAndName;
+import org.jclouds.azurecompute.arm.compute.domain.ResourceGroupAndNameAndIngressRules;
 import org.jclouds.azurecompute.arm.compute.functions.TemplateToAvailabilitySet;
 import org.jclouds.azurecompute.arm.compute.options.AzureTemplateOptions;
 import org.jclouds.azurecompute.arm.domain.AvailabilitySet;
 import org.jclouds.azurecompute.arm.domain.NetworkSecurityGroup;
-import org.jclouds.azurecompute.arm.domain.RegionAndId;
 import org.jclouds.azurecompute.arm.domain.ResourceGroup;
 import org.jclouds.azurecompute.arm.domain.Subnet;
 import org.jclouds.azurecompute.arm.domain.VirtualNetwork;
@@ -42,7 +48,6 @@ import org.jclouds.compute.config.CustomizationResponse;
 import org.jclouds.compute.domain.NodeMetadata;
 import org.jclouds.compute.domain.Template;
 import org.jclouds.compute.functions.GroupNamingConvention;
-import org.jclouds.compute.options.TemplateOptions;
 import org.jclouds.compute.reference.ComputeServiceConstants;
 import org.jclouds.compute.strategy.CreateNodeWithGroupEncodedIntoName;
 import org.jclouds.compute.strategy.CustomizeNodeAndAddToGoodMapOrPutExceptionIntoBadMap;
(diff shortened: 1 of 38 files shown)


azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/strategy/CreateResourcesThenCreateNodes.java
```diff
@@ -117,24 +114,22 @@ protected CreateResourceGroupThenCreateNodes(
 
       // This sill create the resource group if it does not exist
       String location = template.getLocation().getId();
-      ResourceGroup resourceGroup = resourceGroupMap.getUnchecked(location);
-      String azureGroupName = resourceGroup.name();
 
-      getOrCreateVirtualNetworkWithSubnet(location, options, azureGroupName);
-      configureSecurityGroupForOptions(group, azureGroupName, template.getLocation(), options);
+      createResourceGroupIfNeeded(group, location, options);
```
andreaturli: as you are refactoring it, wouldn't be better to avoid function with side-effects (jclouds doesn't have many of them) and set the `ResourceGroup` on `options` ?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/strategy/CreateResourcesThenCreateNodes.java
```diff
@@ -117,24 +114,22 @@ protected CreateResourceGroupThenCreateNodes(
...
+      getOrCreateVirtualNetworkWithSubnet(location, options);
```
andreaturli: same as above

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/strategy/CreateResourcesThenCreateNodes.java
```diff
@@ -117,24 +114,22 @@ protected CreateResourceGroupThenCreateNodes(
...
+      configureSecurityGroupForOptions(group, template.getLocation(), options);
```
andreaturli: same as above

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceLiveTest.java
```diff
@@ -54,19 +52,17 @@
 @Test(groups = "live", singleThreaded = true, testName = "AzureComputeServiceLiveTest")
 public class AzureComputeServiceLiveTest extends BaseComputeServiceLiveTest {
    
-   private LoadingCache<String, ResourceGroup> resourceGroupMap;
    private Predicate<URI> resourceDeleted;
+   private String resourceGroupName;
 
    public AzureComputeServiceLiveTest() {
       provider = "azurecompute-arm";
+      resourceGroupName = getClass().getSimpleName().toLowerCase();
```
andreaturli: +1

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/strategy/CreateResourcesThenCreateNodes.java
nacx: I'm not really sure about this. If we do so, we should expose the method in the options that requires the object, which is not very convenient when it does not exist and you expect jclouds to create it for you. One possible option would be to keep the method to set the name, and add one to provide the ResourceGroup object, but that does not fix the side effect here: even if an object is set in the options, we need to check if that resource group exists and create it if missing, so the logic would be pretty much the same.
Regarding the jclouds code that consumes the resourceGroup option, the only thing that is used is actually the name, to create the different feature APIs, so the object is not really needed internally.
Thoughts?

nacx:
BTW, all tests green! They took almost 4h to complete. I've seen several tests using Windows VMs that will take less for sure if using small Linux boxes. Deleting stuff is also quite expensive and ARM takes its time. We should try to improve the tests to reduce the time it takes to run them all, but this is a different story :)

```bash
Results :

Tests run: 143, Failures: 0, Errors: 0, Skipped: 2

[INFO]
[INFO] --- duplicate-finder-maven-plugin:1.1.1:check (default) @ azurecompute-arm ---
[WARNING] <ignoreResources> has been deprecated and replaced with <ignoreResourcePatterns>. It will go away in version 1.2.0. Please update your POM accordingly!
[INFO] Checking compile classpath
[INFO] Checking runtime classpath
[INFO] Checking test classpath
[INFO]
[INFO] --- maven-checkstyle-plugin:2.15:check (default) @ azurecompute-arm ---
[INFO]
[INFO] --- modernizer-maven-plugin:1.4.0:modernizer (modernizer) @ azurecompute-arm ---
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 03:56 h (Wall Clock)
[INFO] Finished at: 2017-04-24T20:44:28+02:00
[INFO] Final Memory: 51M/633M
[INFO] ------------------------------------------------------------------------
```

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/strategy/CreateResourcesThenCreateNodes.java
andreaturli: I think you are right although I'd probably prefer to read something like
```
...
      if (options.getResourceGroup() != null) {
         group = options.getResourceGroup();
      }


   private ResourceGroup getOrCreateResourceGroup(String group, String location) {
      if (group != null) {
      logger.debug(">> using resource group [%s]", group);
      ResourceGroup rg = api.getResourceGroupApi().get(group);
      if (rg != null)  return rg;

      logger.debug(">> resource group [%s] does not exist. Creating!", group);
      return api.getResourceGroupApi().create(group, location,
            ImmutableMap.of("description", "jclouds default resource group"));
      }
   }
```
and then set the resourceGroup.name on the options object where needed.

but maybe it is not significantly better. Again, just a minor comment :)

andreaturli:
Here's my result, @nacx ([INFO] Total time: 03:35 h !!!)
```
Failed tests:
  AzureComputeServiceLiveTest>BaseComputeServiceLiveTest.testConcurrentUseOfComputeServiceToCreateNodes:499 » Execution
  AzureComputeSecurityGroupExtensionLiveTest.testCreateNodeWithInboundPorts:130 expected [{ipProtocol=tcp, fromPort=22, toPort=24, tenantIdGroupNamePairs={}, groupIds=[], cidrBlocks=[0.0.0.0/0], exclusionCidrBlocks=[]}] but found [{ipProtocol=tcp, fromPort=8000, toPort=8000, tenantIdGroupNamePairs={}, groupIds=[], cidrBlocks=[0.0.0.0/0], exclusionCidrBlocks=[]}]
  ImageApiLiveTest.testCreateImage:116 » RunNodes error running 1 node group(ima...
  LoadBalancerApiLiveTest.testAttachNodesToBackendPool:249->createVirtualMachinesInGroupAndGetNicRefs:350 » RunNodes

Tests run: 143, Failures: 4, Errors: 0, Skipped: 5
```

andreaturli:
as per the irc discussion I agree those failures are mostly due to operation timeout (maybe azure was slower than usual today?) so I think it is a v good improvement overall! 

+1

nacx:
Amended the first commit message to properly reference all fixed issues and merged. Thanks @andreaturli for all testing. It's much appreciated!

PR #386, opened 2017-04-26
JCLOUDS-1278: Allow to configure virtual machine NICs in Azure ARM

nacx:
https://issues.apache.org/jira/browse/JCLOUDS-1278

Allows to configure the NIC layout of a virtual machine using the template options. This PR removes the `virtualnetworkName` and `subnetId` from the `AzureTemplateOptions` class and leverages a more flexible way to configure networking for nodes. Configuration can be done now as follows:

* If no options are set, jclouds will create a default network and a subnet as it did in previous versions, allocate a private and a public ip and attach them to a NIC attached to the node.
* Using the portable options to specify the IDs of the subnets where the nodes will be attached to.
    ```java
    options.networks(subnetId2, subnetId2, ...);
    ```
    For each configured subnet, jclouds will create a NIC with a dynamically allocated private ip address in the subnet.
* Using the `AzureTemplateOptions` to have more control on how the ips are allocated. The following configurations are valid examples:
    ```java
    options.ipOptions(
       IpOptions.builder().subnet(subnet1).address("10.0.0.5").allocatePublicIp(true).build(),
       IpOptions.builder().subnet(subnet1).allocatePublicIp(true).build(),
       IpOptions.builder().subnet(subnet2).build()
    );
    ```
    In this example, the options will create 3 NICs and attach them to the virtual machine. The first NIC will have a statically configured private address and public ip address associated, the second one will have a dynamically allocated private address and a public address, and the third one will just have a dynamically allocated private address.

I'm running the test now, but I've submitted this to have early feedback. as usual, comments welcome!

I'd say this is the only thing that was remaining in order to properly support all options in the portable abstraction. Once this is merged I'd say we are in a very good position to test the provider for a while and consider promoting it to the main repo.

/cc @andreaturli @danielestevez


andreaturli:
thanks @nacx I really like the kind of flexibility jclouds azurecompute-arm can gain with that!

couple of minor inital comments: 
1. don't we need to deprecate `subnetId` and `virtualNetworkName` before removing them, maybe?
2. with the second option `options.networks(subnetId2, subnetId2, ...);` how the nodes will get a public IP?

---

as a broader discussion, I really think we are pushing jclouds compute to the limit: it's original ability to manage nodes nowadays requires a lot of networking resource to be created "around" the node before the provisioning can happen. I wonder if it's time for a `jclouds network` abstraction where jclouds can manage concepts like vpc, subnets, security groups so the `jclouds compute` abstraction can then consume those resources created in a portable way. Of course this is well beyond the scope of this PR, but I'd like to know your thoughts on that.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
```diff
@@ -383,39 +391,94 @@ private OSProfile createOsProfile(String computerName, Template template) {
...
-   private NetworkInterfaceCard createNetworkInterfaceCard(String subnetId, String name, String locationName,
-                                                           String azureGroup, TemplateOptions options) {
-      final PublicIPAddressApi ipApi = api.getPublicIPAddressApi(azureGroup);
+   private List<NetworkInterfaceCard> createNICs(String nodeName, String location, AzureTemplateOptions options) {
+      ImmutableList.Builder<NetworkInterfaceCard> nics = ImmutableList.builder();
+      String securityGroup = getOnlyElement(options.getGroups(), null);
 
-      PublicIPAddressProperties properties = PublicIPAddressProperties.builder().publicIPAllocationMethod("Static")
-              .idleTimeoutInMinutes(4).build();
+      for (IpOptions ipConfig : publicIpsFirst(options.getIpOptions())) {
```
andreaturli: maybe this logic deserves to live in a private method? or a Function name `ipOptionsTo IpConfiguration` maybe?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
```diff
@@ -383,39 +391,94 @@ private OSProfile createOsProfile(String computerName, Template template) {
...
-      String networkInterfaceCardName = "jc-nic-" + name;
-      return api.getNetworkInterfaceCardApi(azureGroup).createOrUpdate(networkInterfaceCardName, locationName,
-              networkInterfaceCardProperties.build(), ImmutableMap.of("jclouds", name));
+      return nics.build();
+   }
+   
+   private NetworkProfile createNetworkProfile(List<NetworkInterfaceCard> nics) {
+      List<NetworkInterface> nicAttachments = new ArrayList<NetworkInterface>(nics.size());
+      for (int i = 0; i < nics.size(); i++) {
+         nicAttachments.add(NetworkInterface.create(nics.get(i).id(), NetworkInterfaceProperties.create(i == 0)));
```
andreaturli: maybe a javadoc to explain the method as it is not obvious, especially the `i == 0` part

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
```diff
@@ -383,39 +391,94 @@ private OSProfile createOsProfile(String computerName, Template template) {
...
+      }
+      return NetworkProfile.create(nicAttachments);
+   }
+   
+   private static List<IpOptions> publicIpsFirst(List<IpOptions> ipOptions) {
```
andreaturli: javadoc maybe to explain the reason why you are sorting this list?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/extensions/AzureComputeSecurityGroupExtension.java
```diff
@@ -135,12 +137,12 @@ public boolean apply(SecurityGroup input) {
...
          throw new IllegalArgumentException("Node " + nodeId + " was not found");
       }
-      List<IdReference> networkInterfacesIdReferences = vm.properties().networkProfile().networkInterfaces();
+      List<NetworkInterface> networkInterfacesIdReferences = vm.properties().networkProfile().networkInterfaces();
       List<NetworkSecurityGroup> networkGroups = new ArrayList<NetworkSecurityGroup>();
 
-      for (IdReference networkInterfaceCardIdReference : networkInterfacesIdReferences) {
-         String nicName = networkInterfaceCardIdReference.name();
-         String nicResourceGroup = networkInterfaceCardIdReference.resourceGroup();
+      for (NetworkInterface networkInterfaceCardIdReference : networkInterfacesIdReferences) {
```
andreaturli: maybe `networkInterfaceCardIdReference` is now `networkInterface` ?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/extensions/AzureComputeSecurityGroupExtension.java
```diff
@@ -135,12 +137,12 @@ public boolean apply(SecurityGroup input) {
       if (vm == null) {
          throw new IllegalArgumentException("Node " + nodeId + " was not found");
       }
-      List<IdReference> networkInterfacesIdReferences = vm.properties().networkProfile().networkInterfaces();
+      List<NetworkInterface> networkInterfacesIdReferences = vm.properties().networkProfile().networkInterfaces();
```
andreaturli: [minor] is maybe `networkInterfaces` rather than `networkInterfacesIdReferences` ?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/functions/VirtualMachineToNodeMetadata.java
```diff
@@ -136,9 +138,9 @@ public NodeMetadata apply(VirtualMachine virtualMachine) {
       return builder.build();
    }
 
-   private Iterable<String> getPrivateIpAddresses(List<IdReference> idReferences) {
+   private Iterable<String> getPrivateIpAddresses(List<NetworkInterface> idReferences) {
```
andreaturli: [minor] `networkInterfaces` instead of `idReferences` ?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/options/AzureTemplateOptions.java
```diff
@@ -32,28 +32,11 @@
...
-   private String subnetId;
    private AvailabilitySet availabilitySet;
    private String availabilitySetName;
    private List<DataDisk> dataDisks = ImmutableList.of();
    private String resourceGroup;
-
-   /**
-    * Sets the virtual network name
-    */
-   public  AzureTemplateOptions virtualNetworkName(String virtualNetworkName) {
```
andreaturli: do we need to deprecate them first or because it is still in labs we can change the public api this way?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/strategy/CleanupResources.java
```diff
@@ -86,10 +89,10 @@ public boolean cleanupNode(final String id) {
    }
 
    public void cleanupVirtualMachineNICs(VirtualMachine virtualMachine) {
-      for (IdReference nicRef : virtualMachine.properties().networkProfile().networkInterfaces()) {
-         String nicResourceGroup = nicRef.resourceGroup();
-         String nicName = nicRef.name();
-         NetworkInterfaceCard nic = api.getNetworkInterfaceCardApi(nicRef.resourceGroup()).get(nicName);
+      for (NetworkInterface nicRef : virtualMachine.properties().networkProfile().networkInterfaces()) {
```
andreaturli: just `nic` instead of `nicRef` ?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/strategy/CreateResourcesThenCreateNodes.java
```diff
@@ -196,4 +197,35 @@ private void createResourceGroupIfNeeded(String group, String location, AzureTem
...
    }
+   
+   private void normalizeNetworkOptions(AzureTemplateOptions options) {
+      if (!options.getNetworks().isEmpty() && !options.getIpOptions().isEmpty()) {
+         throw new IllegalArgumentException(
+               "The options.networks and options.ipOptions are exclusive. Just configure the networking options once.");
+      }
+      
+      if (!options.getNetworks().isEmpty() && options.getIpOptions().isEmpty()) {
+         // The portable interface allows to configure network IDs (subnet IDs),
```
andreaturli: move this comment to javadoc maybe?

andreaturli: we need a good testing coverage for this method :D

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/Subnet.java
```diff
@@ -29,6 +31,8 @@
 @AutoValue
 public abstract class Subnet {
 
+   private static final Pattern NETWORK_PATTERN = Pattern.compile("^.*/virtualNetworks/([^/]+)(/.*)?$");
```
andreaturli: maybe `VIRTUAL_NETWORK_PATTERN` ?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
nacx: Not sure a function would be that appropriate here. This method (as a private method) depends a lot on the current context. It not only depends on the ip configurations, but also on other template options such as the security groups and the name of the node. I think it makes sense here as a private method.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/options/AzureTemplateOptions.java
nacx: We don't keep backwards compatibility in labs.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/strategy/CleanupResources.java
nacx: Well, it is actually a reference to a network interface card, since it just holds its ID and a flag indicating if it is the primary one.

nacx:
@andreaturli I've addressed your comments and added a method to allow binding an existing public ip, which is also convenient. Something like: `IpOptions.builder().subnet(subnet1).publicIpId("pub-ip-id").build()`. I've also fixed a couple race conditions when creating security groups; the jclouds code did not wait until they were in provisioning state "Succeeded".

Regarding your comments:

>don't we need to deprecate subnetId and virtualNetworkName before removing them, maybe?

We don't do that in labs. Labs is a place to experiment with the providers, get familiar with their APIs and best practices. if we kept backwards compatibility there we wouldn't be able to never move forward with the development of new providers.

>with the second option options.networks(subnetId2, subnetId2, ...); how the nodes will get a public IP?

It doesn't. If you are explicit about network configuration you also need to be explicit about public IPs. That method exists to properly implement the portable options, but portability has limitations. If you need more fine grained control you need to use the provider-specific ones. (Also take into account that when building nodes in concrete subnets only the gateway one is likely to have an ip address, so I think it is a good compromise for the portable abstraction).

>as a broader discussion, I really think we are pushing jclouds compute to the limit: it's original ability to manage nodes nowadays requires a lot of networking resource to be created "around" the node before the provisioning can happen. I wonder if it's time for a jclouds network abstraction where jclouds can manage concepts like vpc, subnets, security groups so the jclouds compute abstraction can then consume those resources created in a portable way. Of course this is well beyond the scope of this PR, but I'd like to know your thoughts on that.

Absolutely. We should file a story in JIRA and create a page in the wiki to define how we want it to look like, by looking at how the different major providers implement networking, what they offer and what we want to cover.

geomacy:
This sounds good but I think it may run into the problem we in Apache Brooklyn found when experimenting with arrangements for multiple NICs in Azure - just because you attach the NIC to the machine, that does not mean that the image will actually recognise and enable it when the operating system boots.     E.g. we attached one extra NIC to a node but on boot when you used `ifconfig` you could only see the one, default, interface.  You would have to do additional machine configurations to actually attach the second NIC.

geomacy:
@neykov this will interest you I'm sure

nacx:
There is the option to create secondary ip addresses in the NICs, but that does not look very convenient. A common use case for several NICs is to connect the VM to different subnets, and that does not seem right with secondary ip addresses. That's why I implemented the multiple IP configurations as new NICs. There is also a limitation on the number of NICs you can attach to a VM depending on the size you select, but that validation is also out of the scope of this PR.

I agree that some OSs need to be prepared to enable the network interfaces on boot, but that can easily be achieved with bootstrap scripts and/or cloud init?

geomacy:
> I agree that some OSs need to be prepared to enable the network interfaces on boot, but that can easily be achieved with bootstrap scripts and/or cloud init?

I guess so long as it is understood that this may be necessary, depending on the image, then that's ok.   It may be worth adding this note in the javadoc. Otherwise users may think that just configuring the NICs on the node may mean that they automatically get interfaces connected to the various subnets, and may then wonder what they are doing wrong if in some cases they don't.

nacx:
> It may be worth adding this note in the javadoc.

I'll do. I've also created [JCLOUDS-1282](https://issues.apache.org/jira/browse/JCLOUDS-1282) to add support for the `customData` in the template options (as we do in several other providers that support cloud-init or bootstrap scripts natively in their APIs) so we can inject it when [creating the OS profile](https://github.com/jclouds/jclouds-labs/blob/master/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java#L383). This would make it easier for people to inject cloud-init scripts.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/options/AzureTemplateOptions.java
```diff
@@ -75,13 +75,26 @@ public AzureTemplateOptions dataDisks(DataDisk... dataDisks) {
       return dataDisks(ImmutableList.copyOf(checkNotNull(dataDisks, "dataDisks")));
    }
    
+   /**
+    * Configure the NICs that will be attached to the created nodes.
+    * <p>
+    * Note that the number of NICs that can be attached depends on the size of
+    * the virtual machine, and that the guest operating system needs to be
+    * prepared to setup ont only the first network interface.
```
geomacy: typo `ont`

nacx: Ouch! Fixed :)

nacx:
Compute service live tests are green. @andreaturli @geomacy Good to merge?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/options/AzureTemplateOptions.java
```diff
@@ -75,13 +75,26 @@ public AzureTemplateOptions dataDisks(DataDisk... dataDisks) {
...
+    * <p>
+    * Depending on the image being used, a cloud-init or bootstrap script might
+    * be needed to make the interface setup.
```
geomacy: is it worth adding a distinction between the main NIC which will normally(?) be attached, and additional ones that won't?  Or do you mean above 
"and that the guest operating system will typically set up only the first network interface."

nacx: Guests typically set up the first interface that is exposed to them (eth0, ens0, whatever). Which NIC is exposed to the guest as the first one is up to the hypervisor. In ARM, when creating the network profile you mark the primary NIC with a `primary` flag, and I'd like to think that is the one that is exposed as the first NIC, although I don't know the details of the internals here.
For this reason, the code [sorts](https://github.com/nacx/jclouds-labs/blob/c1c15e9465912244fc0d0ac704f51ace14bafd37/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java#L400) the IpOptions objects in the template options and puts the ones with public addresses first, as they are the publicly reachable ones and likely to be the primary NICs.

What about something like "the guest operating system should be prepared to set up all the configured interfaces"?

nacx:
@andreaturli @geomacy I've added a last change to cleanup only the public ip addresses that have been auto-generated by jclouds. If users are using the template options to attach nodes to existing public ips, we don't want those ips to be removed when removing the node.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/options/AzureTemplateOptions.java
geomacy: I see; that's a good explanation. Why not copy-and-paste it into the javadocs?!

geomacy:
I can't really say if it's good to merge as such, as I haven't reviewed the code in detail, but the updated java docs per our conversations do address the point I made, so find by me.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/compute/strategy/CreateResourcesThenCreateNodesTest.java
```diff
@@ -0,0 +1,114 @@
...
+import org.jclouds.azurecompute.arm.domain.PublicIPAddressProperties;
+import org.jclouds.azurecompute.arm.domain.Subnet;
+import org.jclouds.azurecompute.arm.features.PublicIPAddressApi;
+import org.jclouds.azurecompute.arm.features.SubnetApi;
+import org.testng.annotations.Test;
+
+import com.google.common.collect.ImmutableList;
+
+@Test(groups = "unit", testName = "CreateResourcesThenCreateNodesTest")
+public class CreateResourcesThenCreateNodesTest {
```
andreaturli: interesting test!

andreaturli:
thanks for taking time to address my comments and give your opinion on jclouds network @nacx 

thanks @geomacy for the great comment on the multi nics scenario. 

I'm still slightly concerned about giving multiple options using the template options (portable and custom) to do something similar. I think javadoc is good, but maybe this advanced support for multi subnets should documented on the readme and potentially jclouds.org website, not in this pr anyway. wdyt?

---

As a side note, the more I think about things like 

> If no options are set, jclouds will create a default network and a subnet as it did in previous versions, allocate a private and a public ip and attach them to a NIC attached to the node.

the more I think we need a `jclouds network` abstraction as those assumptions are not immediate and not consistent across jclouds

nacx:
> maybe this advanced support for multi subnets should documented on the readme and potentially jclouds.org website, not in this pr anyway. wdyt?

+1

> the more I think we need a jclouds network abstraction

+1000 It is very much needed for any provider supporting networking (AWS, GCE, Rackspace, OpenStack, Azure, and others).

> those assumptions are not immediate and not consistent across jclouds

That assumption in particular is what most providers do, including those that don't support networking: they just create a node and give it a public ip. You tell jclouds "create a server in the cloud", and jclouds replies "here it is. you can access it at this IP".

andreaturli:
think you need to rebase and then it should be good to merge

nacx:
Squashed, rebased and pushed to [maste](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/dbadb279) and [2.0.x](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/eeccc63d). Thanks for the feedback @andreaturli and @geomacy!

PR #393, opened 2017-05-23
Adding support for provisioning of Windows VMs with enabled WinRM

ygy:
- Provisioning using pre-existing KeyVault
- Enabling WinRM with pre-existing certificate

### DIFF
--- a/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/OSProfile.java
+++ b/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/OSProfile.java
@@ -16,15 +16,14 @@
  */
 package org.jclouds.azurecompute.arm.domain;
 
-import com.google.auto.value.AutoValue;
-import com.google.common.collect.ImmutableList;
+import java.util.List;
 
+import org.jclouds.azurecompute.arm.util.GetEnumValue;
 import org.jclouds.javax.annotation.Nullable;
 import org.jclouds.json.SerializedNames;
 
-import java.util.List;
-import java.util.Map;
-import com.google.common.collect.ImmutableMap;
+import com.google.auto.value.AutoValue;
+import com.google.common.collect.ImmutableList;
 
 @AutoValue
 public abstract class OSProfile {
@@ -91,15 +90,52 @@ public abstract class OSProfile {
 
       @AutoValue
       public abstract static class WinRM {
+          public enum Protocol {
+
+              HTTP("http"),
+              HTTPS("https"),
+              UNRECOGNIZED("Unrecognized");
+
+              private String value;
+
+              Protocol(String value) {
+                 this.value = value;
+              }
+
+              public static Protocol fromValue(String value) {
+                  return (Protocol) GetEnumValue.fromValueOrDefault(value, Protocol.UNRECOGNIZED);
+              }
+
+              @Override
+              public String toString() {
+                 return this.value;
+              }
+           }
+
+          @AutoValue
+          public abstract static class ProtocolListener {
+
+             public abstract Protocol protocol();
+
+             @Nullable
+             public abstract String certificateUrl();
+
+             @SerializedNames({"protocol", "certificateUrl"})
+             public static ProtocolListener create(final Protocol protocol, final String certificateUrl) {
+
+                return new AutoValue_OSProfile_WindowsConfiguration_WinRM_ProtocolListener(
+                        protocol, certificateUrl);
+             }
+          }
 
          /**
           * Map of different settings
           */
-         public abstract Map<String, String> listeners();
+         public abstract List<ProtocolListener> listeners();
 
          @SerializedNames({"listeners"})
-         public static WinRM create(final Map<String, String> listeners) {
-            return new AutoValue_OSProfile_WindowsConfiguration_WinRM(listeners == null ? ImmutableMap.<String, String>of() : ImmutableMap.copyOf(listeners));
+         public static WinRM create(final List<ProtocolListener> listeners) {
+            return new AutoValue_OSProfile_WindowsConfiguration_WinRM(listeners == null ? ImmutableList.<ProtocolListener>of() : ImmutableList.copyOf(listeners));
          }
       }
 
@@ -139,27 +175,20 @@ public abstract class OSProfile {
        * unattend content
        */
       @Nullable
-      public abstract AdditionalUnattendContent additionalUnattendContent();
+      public abstract List<AdditionalUnattendContent> additionalUnattendContent();
 
       /**
        * is automatic updates enabled
        */
       public abstract boolean enableAutomaticUpdates();
 
-      /**
-       * list of certificates
-       */
-      @Nullable
-      public abstract List<String> secrets();
-
-      @SerializedNames({"provisionVMAgent", "winRM", "additionalUnattendContent", "enableAutomaticUpdates",
-              "secrets"})
+      @SerializedNames({"provisionVMAgent", "winRM", "additionalUnattendContent", "enableAutomaticUpdates"})
       public static WindowsConfiguration create(final boolean provisionVMAgent, final WinRM winRM,
-                                                final AdditionalUnattendContent additionalUnattendContent,
-                                                final boolean enableAutomaticUpdates, final List<String> secrets) {
+                                                final List<AdditionalUnattendContent> additionalUnattendContent,
+                                                final boolean enableAutomaticUpdates) {
 
          return new AutoValue_OSProfile_WindowsConfiguration(provisionVMAgent, winRM,
-                 additionalUnattendContent, enableAutomaticUpdates, secrets == null ? null : ImmutableList.copyOf(secrets));
+                 additionalUnattendContent, enableAutomaticUpdates);
       }
    }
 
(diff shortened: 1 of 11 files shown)


azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/OSProfile.java
```diff
@@ -91,15 +90,58 @@ public static LinuxConfiguration create(final String disablePasswordAuthenticati
 
       @AutoValue
       public abstract static class WinRM {
+          public enum ListenerProtocol {
```
andreaturli: is `Protocol` a better name?

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/VirtualMachineApiLiveTest.java
```diff
@@ -96,6 +107,11 @@ public void setup() {
       nicName = nic.name();
 
       vmName = String.format("%3.24s", System.getProperty("user.name") + RAND + this.getClass().getSimpleName()).toLowerCase().substring(0, 15);
+
+      vaultResourceGroup = System.getProperty("test.vault.resource.group");
```
andreaturli: I think this makes the test harder to execute: either you add it to `BaseAzureComputeApiLiveTest` or the test needs to create those prelimiary resources during the setup and destroy during tearDown. wdyt?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/OSProfile.java
```diff
@@ -91,15 +90,58 @@ public static LinuxConfiguration create(final String disablePasswordAuthenticati
...
+              UNRECOGNIZED("Unrecognized");
+
+              private String value;
+
+              ListenerProtocol(String value) {
+                 this.value = value;
+              }
+
+              public static ListenerProtocol fromValue(String value) {
+                 ListenerProtocol[] items = ListenerProtocol.values();
```
aledsage: Could use the same pattern as is in https://github.com/jclouds/jclouds-labs/blob/master/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/ComputeNode.java#L28

ygy:
@andreaturli thanks for the review. I've addressed the comments. The live test receives the parameters via system properties, because they have to exist before the test execution. Azurecompute-arm still can not provision keyvaults and secrets.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/OSProfile.java
```diff
@@ -91,15 +90,58 @@ public static LinuxConfiguration create(final String disablePasswordAuthenticati
...
+              UNRECOGNIZED("Unrecognized");
+
+              private String value;
+
+              Protocol(String value) {
+                 this.value = value;
+              }
+
+              public static Protocol fromValue(String value) {
+                 Protocol[] items = Protocol.values();
```
aledsage: Could use the same pattern as https://github.com/jclouds/jclouds-labs/blob/master/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/ComputeNode.java#L28, i.e.

    return (Protocol) GetEnumValue.fromValueOrDefault(value, Protocol.UNRECOGNIZED);

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VaultCertificate.java
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * The URL of the certificate
+     */
+    public abstract String certificateUrl();
+
+    /**
+     * Certificate's store name
+     */
+    public abstract String certificateStore();
```
aledsage: Needs `@Nullable` - both here, and on the arg to `create` - see https://docs.microsoft.com/en-us/rest/api/compute/virtualmachines/virtualmachines-create-or-update#bk_vaultcert, which says that it is not required.

I believe that if you don't include `@Nullable` then `AutoValue`'s auto-generated code will throw a `NullPointerException` if the value is null.

(Actually, probably don't need it on the arg to `create()` - we don't do that elsewhere).

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/VirtualMachineApiMockTest.java
```diff
@@ -122,18 +130,28 @@ public void testCreateWithPlan() throws Exception {
...
+                  + "\"enableAutomaticUpdates\":true},"
+                  + "\"secrets\":[{\"sourceVault\":{\"id\":\"/subscriptions/SUBSCRIPTIONID/resourceGroups/myresourcegroup1/providers/Microsoft.KeyVault/vaults/myvault1\"},\"vaultCertificates\":[{\"certificateUrl\":\"https://myvault1.vault.azure.net/secrets/SECRETNAME/SECRETVERSION\",\"certificateStore\":\"CERTIFICATESTORENAME\"}]}]},"
                   + "\"networkProfile\":{\"networkInterfaces\":[{\"id\":\"/subscriptions/SUBSCRIPTIONID/resourceGroups/groupname/providers/Microsoft.Network/networkInterfaces/windowsmachine167\"}]},"
                   + "\"diagnosticsProfile\":{\"bootDiagnostics\":{\"enabled\":true,\"storageUri\":\"https://groupname2760.blob.core.windows.net/\"}},\"provisioningState\":\"CREATING\"},"
+                  + "\"tags\":{\"foo\":\"bar\"},"
                   + "\"plan\":{\"name\":\"deadline-slave-7-2\",\"publisher\":\"thinkboxsoftware\",\"product\":\"deadline7-2\"}}");
    }
    
+   // See https://docs.microsoft.com/en-us/rest/api/compute/virtualmachines/virtualmachines-create-or-update
```
aledsage: A slightly more accurate comment would be:

    // See https://docs.microsoft.com/en-us/rest/api/compute/virtualmachines/virtualmachines-create-or-update
    // for where part of the example json response comes from. Unfortunately examples in the microsoft docs
    // are not valid json (e.g. missing commas, illegal quotes). Therefore this example merges the original 
    // real-world example (presumably taken from the jclouds wire log), and snippets from the microsoft docs.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/OSProfile.java
```diff
@@ -91,15 +90,52 @@ public static LinuxConfiguration create(final String disablePasswordAuthenticati
 
       @AutoValue
       public abstract static class WinRM {
+          public enum Protocol {
+
+              HTTP("http"),
+              HTTPS("https"),
+              UNRECOGNIZED("Unrecognized");
```
bostko: I cannot find any "Unrecognized" protocol in Azure documentation.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/internal/BaseAzureComputeApiLiveTest.java
```diff
@@ -82,6 +87,20 @@ protected void tearDown() {
...
+   @BeforeClass
+   @Override
+   public void setup() {
+      super.setup();
+
+      // Providing system properties for specifying the required Azure KeyVault configurations for Live tests
+      // They have to be externally provided, because azurecompute-arm doesn't support creating KeyVaults yet
+      //
+      // TODO Replace the used configurations once full KeyVault implementation is added to azurecompute-arm
+      vaultResourceGroup = System.getProperty("test.vault.resource.group");
```
bostko: Rename to `test.azurecompute-arm.vault.resource.group`.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/OSProfile.java
aledsage: @bostko I think this is just a jclouds convention, of including "UNRECOGNIZED" for if we come across an unknown value (rather than it setting the field to null or throwing an exception).

bostko: I do not think it is appropriate for this case since there are no other options.
Even if there are it will be invalid for any jclouds user.

aledsage: @bostko I think it's better to produce an UNRECOGNIZED value rather than an exception or null, if we get back some different value. For example, if microsoft add a third possible value in their API, which jclouds doesn't understand, then we'll say that is `UNRECOGNIZED`.

neykov:
Thanks @ygy and all reviewers. Changes look good. Merging.

neykov:
Merged to [master](https://git1-us-west.apache.org/repos/asf?p=jclouds-labs.git;a=commit;h=6e0cbc915cfb1e3f0cb9cdafd3b56b948e0e80a8).

aledsage:
Thanks @ygy (and @neykov @andreaturli @bostko)!

Could one of you please add this to the 2.0.x branch as well?

neykov:
Added to [2.0.x](https://git1-us-west.apache.org/repos/asf?p=jclouds-labs.git;a=commit;h=c4a91d2d909a4edf52adf1367b5353d81cde681b).

PR #395, opened 2017-05-29
Add extra check on SingleServerApplianceToImage function in case the osFamily is null

alibazlamit:
1&1 allows uploading custom images therefor some of the images might have a null value for the osFamily. The necessary check was added, please merge.

Thanks!


andreaturli:
lgtm

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/compute/function/SingleServerApplianceToImage.java
```diff
@@ -74,7 +74,11 @@ static boolean is64Bit(int architecture) {
       return tryFind(asList(OsFamily.values()), new Predicate<OsFamily>() {
          @Override
          public boolean apply(OsFamily input) {
-            return osFamily.toLowerCase().contains(input.value().toLowerCase());
+            if (osFamily != null) {
```
nacx: `osFamily` is the input parameter of the function. There's no need to do the same redundant check in each iteration. Remove this and directly return absent at the beginning if the input parameter is null. Same for the other function.

alibazlamit:
Changes applied.
Thanks!

nacx:
Can you please squash the commits?

alibazlamit:
Squashed .

nacx:
Pushed to [master](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/3b8f5c60) and [2.0.x](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/f294ab10). Thanks @alibazlamit!

PR #396, opened 2017-06-13
Implements metrics and metricdefinitions API

danielestevez:
- https://docs.microsoft.com/en-us/rest/api/monitor/metrics
- https://docs.microsoft.com/en-us/rest/api/monitor/metricdefinitions

### DIFF
--- a/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/AzureComputeApi.java
+++ b/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/AzureComputeApi.java
@@ -27,6 +27,8 @@ import org.jclouds.azurecompute.arm.features.ImageApi;
 import org.jclouds.azurecompute.arm.features.JobApi;
 import org.jclouds.azurecompute.arm.features.LoadBalancerApi;
 import org.jclouds.azurecompute.arm.features.LocationApi;
+import org.jclouds.azurecompute.arm.features.MetricDefinitionsApi;
+import org.jclouds.azurecompute.arm.features.MetricsApi;
 import org.jclouds.azurecompute.arm.features.NetworkInterfaceCardApi;
 import org.jclouds.azurecompute.arm.features.NetworkSecurityGroupApi;
 import org.jclouds.azurecompute.arm.features.NetworkSecurityRuleApi;
@@ -208,4 +210,22 @@ public interface AzureComputeApi extends Closeable {
     */
    @Delegate
    ImageApi getVirtualMachineImageApi(@PathParam("resourcegroup") String resourcegroup);
+
+   /**
+    * The metrics API includes operations to get insights into entities within your
+    * subscription.
+    *
+    * @see <a href="https://docs.microsoft.com/en-us/rest/api/monitor/metrics">docs</a>
+    */
+   @Delegate
+   MetricsApi getMetricsApi(@PathParam("resourceid") String resourceid);
+
+   /**
+    * The metric definitions API includes operations to get insights available for entities within your
+    * subscription.
+    *
+    * @see <a href="https://docs.microsoft.com/en-us/rest/api/monitor/metricdefinitions">docs</a>
+    */
+   @Delegate
+   MetricDefinitionsApi getMetricsDefinitionsApi(@PathParam("resourceid") String resourceid);
 }
--- a/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/Metric.java
+++ b/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/Metric.java
@@ -1,0 +1,64 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jclouds.azurecompute.arm.domain;
+
+import java.util.List;
+
+import org.jclouds.javax.annotation.Nullable;
+import org.jclouds.json.SerializedNames;
+
+import com.google.auto.value.AutoValue;
+
+/**
+ * A Metric with its values for a resource
+ */
+@AutoValue
+public abstract class Metric {
+
+   public abstract List<MetricData> data();
+
+   public abstract String id();
+
+   @Nullable
+   public abstract Metric.MetricName name();
+
+   public abstract String type();
+
+   public abstract String unit();
+
+   @SerializedNames({ "data", "id", "name", "type", "unit" })
+   public static Metric create(final List<MetricData> data, final String id, final MetricName name, final String type,
+         final String unit) {
+      return new AutoValue_Metric(data, id, name, type, unit);
+   }
+
+   @AutoValue
+   public abstract static class MetricName {
+      MetricName() {
+
+      }
+
+      @SerializedNames({ "value", "localizedValue" })
+      public static Metric.MetricName create(String value, String localizedValue) {
+         return new AutoValue_Metric_MetricName(value, localizedValue);
+      }
+
+      public abstract String value();
+
+      public abstract String localizedValue();
+   }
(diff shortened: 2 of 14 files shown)


azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/Metric.java
```diff
@@ -0,0 +1,64 @@
...
+   public abstract Metric.MetricName name();
+
+   public abstract String type();
+
+   public abstract String unit();
+
+   @SerializedNames({ "data", "id", "name", "type", "unit" })
+   public static Metric create(final List<MetricData> data, final String id, final MetricName name, final String type,
+         final String unit) {
+      return new AutoValue_Metric(data, id, name, type, unit);
```
nacx: Enforce immutable and non-nullable lists. If this is a read-only object, not something users will build and send to ARM in a request, then avoid having nullable collections (the typical `tags` field of an object is a counter-example as we need to send it as `null`, so we enforce immutability but not its presence):
```java
new AutoValue_Metric(data == null ? ImmutableList.of() : ImmutableList.copyOf(data), id, name, type, unit);
```
Apply this pattern to all lists in the new model classes.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/MetricData.java
```diff
@@ -0,0 +1,72 @@
...
+   /**
+    * The timestamp for the metric value in ISO 8601 format.
+    */
+   public abstract Date timeStamp();
+
+   /**
+    * The average value in the time range
+    */
+   @Nullable
+   public abstract String total();
```
nacx: Why is this a String? Can we better make this a Double/Long? The same applies to all other numeric value fields.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/MetricDefinition.java
```diff
@@ -0,0 +1,90 @@
...
+   public abstract String resourceid();
+
+   public abstract MetricDefinition.MetricName name();
+
+   @Nullable
+   public abstract Boolean isDimensionRequired();
+
+   public abstract String unit();
+
+   public abstract String primaryAggregationType();
```
nacx: Is this a set of fixed values? Can we codify them in an enum?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/MetricDefinition.java
```diff
@@ -0,0 +1,90 @@
...
+import com.google.auto.value.AutoValue;
+
+/**
+ * A Metric definition for a resource
+ */
+@AutoValue
+public abstract class MetricDefinition {
+
+   @Nullable
+   public abstract String resourceid();
```
nacx: Better use camel case.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/MetricDefinition.java
```diff
@@ -0,0 +1,90 @@
...
+
+   public abstract String id();
+
+   @SerializedNames({ "resourceid", "name", "isDimensionRequired", "unit", "primaryAggregationType",
+         "metricAvailabilities", "id" })
+   public static MetricDefinition create(final String resourceid, final MetricName name,
+         final Boolean isDimensionRequired, final String unit, final String primaryAggregationType,
+         List<MetricAvailability> metricAvailabilities, final String id) {
+      return new AutoValue_MetricDefinition(resourceid, name, isDimensionRequired, unit, primaryAggregationType,
+            metricAvailabilities, id);
```
nacx: Same comment about collection immutability.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/MetricDefinition.java
```diff
@@ -0,0 +1,90 @@
...
+      }
+
+      @SerializedNames({ "timeGrain", "retention" })
+      public static MetricDefinition.MetricAvailability create(String timeGrain, String retention) {
+         return new AutoValue_MetricDefinition_MetricAvailability(timeGrain, retention);
+      }
+   }
+
+   @AutoValue
+   public abstract static class MetricName {
```
nacx: This is already defined in the `Metric` class. Better extract it to a top level type.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/MetricDefinitionsApiLiveTest.java
```diff
@@ -0,0 +1,130 @@
...
+
+   @Override
+   @BeforeClass
+   public void setupContext() {
+      super.setupContext();
+      NodeMetadata node = null;
+      try {
+         node = getOnlyElement(view.getComputeService().createNodesInGroup(group, 1, resourceGroup(group)));
+      } catch (RunNodesException e) {
+         e.printStackTrace();
```
nacx: Fail the test instead of silently printing the error to the stdout?

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/MetricsApiLiveTest.java
```diff
@@ -0,0 +1,142 @@
...
+      super.setupContext();
+
+      dateFormat = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss'Z'");
+      startTime = dateFormat.format(new Date());
+
+      NodeMetadata node = null;
+      try {
+         node = getOnlyElement(view.getComputeService().createNodesInGroup(group, 1, resourceGroup(group)));
+      } catch (RunNodesException e) {
+         e.printStackTrace();
```
nacx: Same comment

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/AzureComputeApi.java
```diff
@@ -208,4 +210,22 @@ NetworkSecurityRuleApi getNetworkSecurityRuleApi(@PathParam("resourcegroup") Str
...
+   MetricsApi getMetricsApi(@PathParam("resourceid") String resourceid);
+
+   /**
+    * The metric definitions API includes operations to get insights available for entities within your
+    * subscription.
+    *
+    * @see <a href="https://docs.microsoft.com/en-us/rest/api/monitor/metricdefinitions">docs</a>
+    */
+   @Delegate
+   MetricDefinitionsApi getMetricsDefinitionsApi(@PathParam("resourceid") String resourceid);
```
danielestevez: That was my original idea but they're two different APIs in Azure and therefore deal with different entities and api-versions so i thought it'd be more maintainable if we'd keep it that way

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/MetricData.java
danielestevez: It's a String according to Azure documentation but yes, i think we could get them as Double/Long. I'll test this

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/MetricDefinition.java
danielestevez: Correct. Will change to enum: None, Average, Count, Total, Minimum, Maximum seem to be the only possible values

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/MetricData.java
danielestevez: Ok, we could get Double for all fields but i still think we should return a String as Azure API does. If we force Double for let's say "total" this could be ok for metrics in bytes (as Network In) but would make no sense for metrics with no decimals (like Disk Read Operations/Sec that is a smaller count that could be even Int)

Since jclouds will not do any calculations with these values i think we should return them as they are and let the user decide how to parse them.

Thoughts?

nacx: I still think a `Number` makes more sense. Having a number with `.0` decimal is not a big deal, as you can still do arithmetic operations, etc, normally. Letting the users how to parse them propagates unnecessary complexity to them.

nacx:
Squashed and pushed to [master](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/47c4e72a) and [2.0.x](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/4183e2f9).
I also changed the metrics mock test so it uses the date service to parse dates (see https://github.com/jclouds/jclouds-labs/commit/dffc2bba7153a8222b18139298d76dcb2b4bbb6d). Otherwise, there are locale-dependent issues when parsing dates and the results were non-deterministic.

PR #397, opened 2017-06-14
Improve duplicate ssh key check in Packet

nacx:
@andreaturli please review.

Apologies for the noise in the PR; I've formatted the class to 3-space indent. There is a small set of changes:

* Generate a unique name for the generated key pair if the user does not provide a key. We were generating them with repeated names.
* Improve the check for the key existence (keys can't be duplicated in Packet). previously we just checked the names. Now the code checks the fingerprints of existing keys to determine if it already exists or not.
* When cleaning up the ssh keys, just delete the ones generated by jclouds, but do not delete any pre-existing key that might be used. if it was there, we leave it there.
* Synchronized the key check and creation to avoid race conditions in concurrent environments.


packet/src/main/java/org/jclouds/packet/compute/strategy/CreateSshKeysThenCreateNodes.java
```diff
@@ -65,152 +66,170 @@
...
-                cleanupAutoGeneratedKeyPairs(generatedSshKeyIds);
-            }
+   private void createKeyPairForPublicKeyInOptionsAndAddToSet(PacketTemplateOptions options,
+         Set<String> generatedSshKeyIds) {
+      logger.debug(">> checking if the key pair already exists...");
 
+      PublicKey userKey = readPublicKey(options.getPublicKey());
+      final String fingerprint = computeFingerprint(userKey);
+      
+      synchronized (CreateSshKeysThenCreateNodes.class) {
```
andreaturli: why is it needed?

packet/src/main/java/org/jclouds/packet/compute/strategy/CreateSshKeysThenCreateNodes.java
```diff
@@ -65,152 +66,170 @@
...
+         boolean keyExists = api.sshKeyApi().list().concat().anyMatch(new Predicate<SshKey>() {
```
andreaturli: isn't `SshKey key = api.sshKeyApi().get(fingerprint);` more appropriate here?

packet/src/main/java/org/jclouds/packet/compute/strategy/CreateSshKeysThenCreateNodes.java
nacx: To avoid race conditions on this "getOrCreate" method. You could have apps/clients accessing the same project with different API keys (that is, two different jclouds contexts), and race conditions could appear here.

packet/src/main/java/org/jclouds/packet/compute/strategy/CreateSshKeysThenCreateNodes.java
nacx: No. The parameter to the `get` method is the "name" of the key. It happens that we were setting the fingerprint as the name when creating the keys (for convenience), but the old implementation, that already did this, was not able to find keys with normal names. We need to inspect the fingerprint field of all keys.

nacx:
>why only against 2.0.x?

I developed it in 2.0.x, but will be merged to master too.

packet/src/main/java/org/jclouds/packet/compute/strategy/CreateSshKeysThenCreateNodes.java
andreaturli: ok makes sense

andreaturli:
lgtm, thanks @nacx

nacx:
Pushed to [master](http://git-wip-us.apache.org/repos/asf/jclouds/commit/b77ea069) and [2.0,.x](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/c1fb7620). Thanks @andreaturli!

PR #398, opened 2017-06-16
JCLOUDS-1304: add azure customdata, JCLOUDS-1305: add support for ssh keys in keyvault

VRanga000:


azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/options/AzureTemplateOptions.java
```diff
@@ -41,7 +41,17 @@
    private List<IpOptions> ipOptions = ImmutableList.of();
    private WindowsConfiguration windowsConfiguration;
    private List<Secrets> secrets = ImmutableList.of();
-   
+   private String customData;
```
nacx: In mots providers, we declare this as a `byte[]`, so we don't have to deal with charset issues, etc. We just get the bytes to be base64-encoded. Mind changing the type and method signatures to accept a byte array?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/options/AzureTemplateOptions.java
```diff
@@ -114,6 +124,14 @@ public AzureTemplateOptions windowsConfiguration(WindowsConfiguration windowsCon
     }
 
    /**
+    * Sets the KeyVault id and secret separated with ":"
+    */
+   public  AzureTemplateOptions keyVaultIdAndSecret(String keyVaultIdAndSecret) {
+      this.keyVaultIdAndSecret = keyVaultIdAndSecret;
```
nacx: Validate that the parameter has the right format.

VRanga000:
Thanks for the review comments @nacx ! I do need to add (back) the code to actually use the keyvaultidAndSecret, will do that. Will address the other comments as well.

VRanga000:
Hello @nacx - need your input on implementation for keyvault support. The azure compute adapter code was refactored to use the VirtualMachineAPI instead of the deployment API. As  a result, keyvaults cannot be accessed merely by reference (as we could in the deployment template). So - I think we need to implement a new "feature" for azurecompute-arm - namely the keyvault API (https://docs.microsoft.com/en-us/rest/api/keyvault/). So I think we should restrict this PR to just "custom data" support" and remove the keyvault refs. Then create a new issue for keyvault support. Thoughts? Thanks much!

nacx:
>Thoughts? Thanks much!

Looks good to me!

neykov:
@VRanga000 what do you think about splitting the PR into `customData` support and another one for keyvault support? The two don't seem to depend on each other, and we can merge `customData` without waiting for the rest.

jmspring:
@VRanga000 Is it possible to split off the custom data support sooner than later and issue the PR with that?  That way we can merge the PR and get azure-armcompute promoted out of labs.  The Keyvault work still relies on the things we talked about last week.  Let me know if I can help at all.

nacx:
I've extracted the `customData` stuff and pushed to [master](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/c9f272b9) and [2.0.x](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/eec594dd). Should we better close this pull request and work on keyvault in a new one?

andreaturli:
thanks @nacx fyi I'm looking at `KeyVaultApi`, focusing on the feature first and then we can together add that at the level of the ComputeAdapter?

jmspring:
Andrea, Ignasi -

I guess Vikas never came back on the one PR?

I'm actually working on the KeyVault stuff myself at the moment.

-j

andreaturli:
thanks @jmspring!

Do you think it would make sense to work on the same branch? my wip is available at https://github.com/andreaturli/jclouds-labs/tree/feature/vault-api in case you want to have a look.

I can make `*keys` operations work for some authorization issues I'm still trying to figure out

jmspring:
Andrea -

I'll be getting back to this in depth tomorrow.  I looked at your work and
some that was done internally.  I'm happy to look at your branch and do PRs
to it.

For the next couple of days it's my focus.  What auth issues are you having?

-j

andreaturli:
Excellent @jmspring 

I can't list/create keys or certificates into a newly created key vault because of an Authorization issue

jmspring:
Ok, I will start with looking into that.


On October 3, 2017 at 3:04:57 AM, Andrea Turli (notifications@github.com)
wrote:

Excellent @jmspring <https://github.com/jmspring>

I can't list/create keys or certificates into a newly created key vault
because of an Authorization issue

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/jclouds/jclouds-labs/pull/398#issuecomment-333797199>,
or mute the thread

jmspring:
@andreaturli - I hope to have some progress on this this week.  I'm going through the live tests now for Vault API.  

Do you have thoughts on externalizing the tenantId which is hard coded in the live test?

nacx:
>Do you have thoughts on externalizing the tenantId which is hard coded in the live test?

I don't have the context of the discussion, but isn't it part of the `oauth.endpoint` property? It could be parsed and exposed, for example in a provider method in the AzureHttpApiModule so it can be injected as a qualified String where needed?

nacx:
Closing this. With the recent addition of the keyvault API and the configuration of the custom data in a different PR this should be better done in a new one.

PR #406, opened 2017-07-27
Subscription ID wasn't being substituted

jmspring:
At some point the substitution of the "SUBSCRIPTION_ID" for the AzureComputeProviderMetadata endpoint stopped working.  This PR fixes that substitution issue to work properly.


jmspring:
@nacx -- at some point the endpoint URL for Azure ARM calls broke.  This PR fixes that.  I'm not sure when/how the exact breakage (change to JClouds maybe?) occurred.  Tested using jclouds-examples/compute-basics.

nacx:
I think [this](https://github.com/jclouds/jclouds-labs/commit/17f7c4528472c87799df28512b7a97571f11f625) is the offending commit.

Having the property makes it easier to create the context. Following the same pattern, it would be good to also preset the `oauth.endpoint` property by default in the provider metadata default properties with a placeholder for the `tenantId` to be substituted the same way. WDYT? This way users will only have to set the two properties and the credentials (and they'll still be able to configure the entire endpoints if they want to).

jmspring:
It makes sense.

I'll plumb it into the pr today.

-j

jmspring:
@nacx Default oauth.endpoint added.  Tested with compute-basics.

nacx:
Pushed to master as [caf91f88](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/caf91f88). Thanks, @jmspring!

PR #407, opened 2017-08-01
Parenthesize confusing conditional

andrewgaul:
Found via error-prone.

### DIFF
--- a/dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/handlers/DimensionDataCloudControlErrorHandler.java
+++ b/dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/handlers/DimensionDataCloudControlErrorHandler.java
@@ -50,7 +50,7 @@ public class DimensionDataCloudControlErrorHandler implements HttpErrorHandler {
             if (message.contains("RESOURCE_NOT_FOUND") || message.contains("OPERATION_NOT_SUPPORTED")) {
                exception = new ResourceNotFoundException(message, exception);
             } else if (message.contains("INVALID_INPUT_DATA") || message.contains("ORGANIZATION_NOT_VERIFIED")
-                  || message.contains("SYSTEM_ERROR") && !message.contains("RETRYABLE_SYSTEM_ERROR") || message
+                  || (message.contains("SYSTEM_ERROR") && !message.contains("RETRYABLE_SYSTEM_ERROR")) || message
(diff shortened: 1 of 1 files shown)


andrewgaul:
@jawnclarke Please review.

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/handlers/DimensionDataCloudControlErrorHandler.java
```diff
@@ -49,9 +49,12 @@ public void handleError(HttpCommand command, HttpResponse response) {
...
                exception = new ResourceNotFoundException(message, exception);
-            } else if (message.contains("INVALID_INPUT_DATA") || message.contains("ORGANIZATION_NOT_VERIFIED")
-                  || message.contains("SYSTEM_ERROR") && !message.contains("RETRYABLE_SYSTEM_ERROR") || message
-                  .contains("CPU_SPEED_NOT_AVAILABLE") || message.contains("CONFIGURATION_NOT_SUPPORTED")) {
+            } else if ((message.contains("CONFIGURATION_NOT_SUPPORTED") ||
+                  message.contains("CPU_SPEED_NOT_AVAILABLE") ||
+                  message.contains("INVALID_INPUT_DATA") ||
+                  message.contains("ORGANIZATION_NOT_VERIFIED") ||
+                  message.contains("SYSTEM_ERROR")) &&
+                  !message.contains("RETRYABLE_SYSTEM_ERROR")) {
```
demobox: @andrewgaul I'm not sure this is the same? As I read the original, it's

`... || ... || (message.contains("SYSTEM_ERROR") && !message.contains("RETRYABLE_SYSTEM_ERROR")) || ... || ...`

which seems to be different from the new version?

E.g. if `message.contains("INVALID_INPUT_DATA") == true` and `message.contains("RETRYABLE_SYSTEM_ERROR") == true`, the old version would evaluate to true but the new version would evaluate to false?

andrewgaul: The intent is not clear to me.  @jawnclarke?

andrewgaul: Reading this more closely, I agree with your interpretation.

PR #262, opened 2016-04-21
JCLOUDS-664 Azurecompute-arm VNetApi and SubnetApi

ritazh:
@nacx @andreaturli 
This PR is for VirtualNetworkApi and SubnetApi, part of the foundations needed for interacting with Azure Resource Manager (ARM) Rest APIs.

FYI
@jmspring @jtjk @isalento

PR #412, opened 2017-10-03
JCLOUDS-1342 : Azure-ARM Virtual Machine Scale Set Support

code4clouds:
This features provide Azure ARM support of Virtual Machine Scale Sets.

### DIFF
--- a/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/Extension.java
+++ b/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/Extension.java
@@ -1,0 +1,40 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jclouds.azurecompute.arm.domain;
+
+import com.google.auto.value.AutoValue;
+import org.jclouds.json.SerializedNames;
+
+@AutoValue
+public abstract class Extension {
+
+   /**
+    * The name reference of the extension profile
+    */
+   public abstract String name();
+
+   /**
+    * The properties reference of the extension profile
+    */
+   public abstract ExtensionProperties properties();
+
+
+   @SerializedNames({"name", "properties"})
+   public static Extension create(final String name, final ExtensionProperties properties) {
+      return new AutoValue_Extension(name, properties);
+   }
+}
--- a/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/ExtensionProfile.java
+++ b/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/ExtensionProfile.java
@@ -1,0 +1,39 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jclouds.azurecompute.arm.domain;
+
+import com.google.auto.value.AutoValue;
+import com.google.common.collect.ImmutableList;
+import org.jclouds.json.SerializedNames;
+
+import java.util.List;
+
+@AutoValue
+public abstract class ExtensionProfile {
+
+   /**
+    * The list of extensions of the extension profile
+    */
+   public abstract List<Extension> extensions();
+
+
+   @SerializedNames({"extensions"})
+   public static ExtensionProfile create(final List<Extension> extensions) {
+      return new AutoValue_ExtensionProfile(extensions == null ?
+         ImmutableList.<Extension>of() : ImmutableList.copyOf(extensions));
+   }
+}
--- a/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/ExtensionProfileSettings.java
+++ b/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/ExtensionProfileSettings.java
@@ -1,0 +1,42 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jclouds.azurecompute.arm.domain;
+
+import com.google.auto.value.AutoValue;
+import com.google.common.collect.ImmutableList;
+import org.jclouds.json.SerializedNames;
+
+import java.util.List;
+
+@AutoValue
+public abstract class ExtensionProfileSettings {
+
+   /**
+    * The fileUris reference of the extension profile settings
+    */
+   public abstract List<String> fileUris();
+
+   /**
+    * The commandToExecute of the extension profile settings
+    */
+   public abstract String commandToExecute();
+
+   @SerializedNames({"fileUris", "commandToExecute"})
+   public static ExtensionProfileSettings create(final List<String> fileUris, final String commandToExecute) {
+      return new AutoValue_ExtensionProfileSettings(fileUris == null ? ImmutableList.<String>of() : ImmutableList.copyOf(fileUris), commandToExecute);
+   }
+}
--- a/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/ExtensionProperties.java
+++ b/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/ExtensionProperties.java
@@ -1,0 +1,70 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jclouds.azurecompute.arm.domain;
+
+import com.google.auto.value.AutoValue;
+import com.google.common.collect.ImmutableMap;
+import org.jclouds.json.SerializedNames;
+
+import java.util.Map;
+
+@AutoValue
+public abstract class ExtensionProperties {
+
+   /**
+    * The publisher reference of the extension properties
+    */
+   public abstract String publisher();
+
+   /**
+    * The type reference of the extension properties
+    */
+   public abstract String type();
+
+   /**
+    * The typeHandlerVersion reference of the extension properties
+    */
+   public abstract String typeHandlerVersion();
+
+   /**
+    * The autoUpgradeMinorVersion reference of the extension properties
+    */
+   public abstract Boolean autoUpgradeMinorVersion();
+
+   /**
+    * The ExtensionProfileSettings of the extension properties
+    */
+   public abstract ExtensionProfileSettings settings();
+
+   /**
+    * The list of the protectedSettings of the extension properties
+    */
+   public abstract Map<String, String> protectedSettings();
+
+   @SerializedNames({ "publisher", "type", "typeHandlerVersion",
+      "autoUpgradeMinorVersion", "settings", "protectedSettings"})
+   public static ExtensionProperties create(final String publisher, String type,
+                                            final String typeHandlerVersion,
+                                            final Boolean autoUpgradeMinorVersion,
+                                            final ExtensionProfileSettings settings,
+                                            final Map<String, String> protectedSettings) {
+      return new AutoValue_ExtensionProperties(publisher, type, typeHandlerVersion, autoUpgradeMinorVersion,
+         settings, protectedSettings == null ?
+         ImmutableMap.<String, String>of() : ImmutableMap.copyOf(protectedSettings));
+   }
+}
+
--- a/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/NetworkInterfaceCard.java
+++ b/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/NetworkInterfaceCard.java
@@ -51,6 +51,7 @@ public abstract class NetworkInterfaceCard {
                                              final String location,
                                              final NetworkInterfaceCardProperties properties,
                                              final Map<String, String> tags) {
-      return new AutoValue_NetworkInterfaceCard(name, id, etag, location, properties, tags == null ? null : ImmutableMap.copyOf(tags));
+      return new AutoValue_NetworkInterfaceCard(name, id, etag, location, properties,
+         tags != null ? ImmutableMap.copyOf(tags) : null);
    }
 }
--- a/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/NetworkInterfaceCardProperties.java
+++ b/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/NetworkInterfaceCardProperties.java
@@ -32,7 +32,8 @@ public abstract class NetworkInterfaceCardProperties implements Provisionable {
    @Nullable public abstract List<IpConfiguration> ipConfigurations();
    @Nullable public abstract IdReference networkSecurityGroup();
 
-   @SerializedNames({"provisioningState", "resourceGuid", "enableIPForwarding", "ipConfigurations", "networkSecurityGroup"})
+   @SerializedNames({"provisioningState", "resourceGuid", "enableIPForwarding", "ipConfigurations",
+      "networkSecurityGroup"})
    public static NetworkInterfaceCardProperties create(final String provisioningState, final String resourceGuid,
          final Boolean enableIPForwarding, final List<IpConfiguration> ipConfigurations,
          final IdReference networkSecurityGroup) {
--- a/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/NetworkInterfaceConfiguration.java
+++ b/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/NetworkInterfaceConfiguration.java
@@ -1,0 +1,44 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jclouds.azurecompute.arm.domain;
+
+import com.google.auto.value.AutoValue;
+import org.jclouds.javax.annotation.Nullable;
+import org.jclouds.json.SerializedNames;
+
+@AutoValue
+public abstract class NetworkInterfaceConfiguration {
+   /**
+    * The name of the NetworkInterfaceConfiguration
+    */
+   @Nullable
+   public abstract String name();
+
+   /**
+    * The networkConfigurationProperties of the NetworkInterfaceConfiguration
+    */
+   @Nullable
+   public abstract NetworkInterfaceConfigurationProperties networkInterfaceConfigurationProperties();
+
+   @SerializedNames({"name", "properties"})
+   public static NetworkInterfaceConfiguration create(
+      final String name, NetworkInterfaceConfigurationProperties networkInterfaceConfigurationProperties) {
+
+      return new AutoValue_NetworkInterfaceConfiguration(name, networkInterfaceConfigurationProperties);
+   }
+
+}
--- a/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/NetworkInterfaceConfigurationProperties.java
+++ b/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/NetworkInterfaceConfigurationProperties.java
@@ -1,0 +1,68 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jclouds.azurecompute.arm.domain;
+
+import com.google.auto.value.AutoValue;
+import org.jclouds.javax.annotation.Nullable;
+import org.jclouds.json.SerializedNames;
+
+import java.util.List;
+
+
+@AutoValue
+public abstract class NetworkInterfaceConfigurationProperties {
+    /**
+     * The primary of the NetworkInterfaceConfigurationProperties
+     */
+    @Nullable
+    public abstract Boolean primary();
+
+    /**
+     * The enableAcceleratedNetworking of the NetworkInterfaceConfigurationProperties
+     */
+    @Nullable
+    public abstract Boolean enableAcceleratedNetworking();
+
+    /**
+     * The networkSecurityGroup of the NetworkInterfaceConfigurationProperties
+     */
+    @Nullable
+    public abstract VirtualMachineScaleSetNetworkSecurityGroup networkSecurityGroup();
+
+   /**
+    * The dnsSettings of the NetworkInterfaceConfigurationProperties
+    */
+   @Nullable
+   public abstract VirtualMachineScaleSetDNSSettings dnsSettings();
+
+   /**
+    * The ipConfigurations of the NetworkInterfaceConfigurationProperties
+    */
+   public abstract List<VirtualMachineScaleSetIpConfiguration> ipConfigurations();
+
+
+    @SerializedNames({"primary", "enableAcceleratedNetworking", "networkSecurityGroup", "dnsSettings", "ipConfigurations"})
+    public static NetworkInterfaceConfigurationProperties create(final Boolean primary,
+                                                               final Boolean enableAcceleratedNetworking,
+                                                               final VirtualMachineScaleSetNetworkSecurityGroup networkSecurityGroup,
+                                                               final VirtualMachineScaleSetDNSSettings dnsSettings,
+                                                               final List<VirtualMachineScaleSetIpConfiguration> ipConfigurations) {
+
+        return new AutoValue_NetworkInterfaceConfigurationProperties(primary, enableAcceleratedNetworking,
+                networkSecurityGroup, dnsSettings, ipConfigurations);
+    }
+}
--- a/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSet.java
+++ b/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSet.java
@@ -1,0 +1,110 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jclouds.azurecompute.arm.domain;
+
+import com.google.auto.value.AutoValue;
+import com.google.common.collect.ImmutableMap;
+import org.jclouds.javax.annotation.Nullable;
+import org.jclouds.json.SerializedNames;
+
+import java.util.Map;
+
+/**
+ * VirtualMachineScaleSet for subscription
+ */
+@AutoValue
+public abstract class VirtualMachineScaleSet {
+
+   /**
+    * The id of the virtual machine scale set
+    */
+   @Nullable
+   public abstract String id();
+
+   /**
+    * The name of the virtual machine scale set
+    */
+   @Nullable
+   public abstract String name();
+
+   /**
+    * The location of the virtual machine scale set
+    */
+   @Nullable
+   public abstract String location();
+
+   /**
+    * Specifies the sku of the virtual machine scale set
+    */
+   public abstract VirtualMachineScaleSetSKU sku();
+
+   /**
+    * Specifies the tags of the virtual machine scale set
+    */
+   @Nullable
+   public abstract Map<String, String> tags();
+
+
+   /**
+    * Specifies the optional plan of the virtual machine scale set (only for market image)
+    */
+   @Nullable
+   public abstract VirtualMachineScaleSetPlan plan();
+
+   /**
+    * Specifies the properties of the availability set
+    */
+   @Nullable
+   public abstract VirtualMachineScaleSetProperties properties();
+
+   @SerializedNames({  "id", "name", "location", "sku", "tags", "plan", "properties"})
+   public static VirtualMachineScaleSet create(final String id, final String name, final String location,
+                                               VirtualMachineScaleSetSKU sku, final Map<String, String> tags,
+                                               VirtualMachineScaleSetPlan plan,
+                                               VirtualMachineScaleSetProperties properties) {
+      return builder().id(id).name(name).location(location).sku(sku).tags(tags)
+         .plan(plan).properties(properties)
+         .build();
+   }
+
+   public abstract Builder toBuilder();
+
+   private static Builder builder() {
+      return new AutoValue_VirtualMachineScaleSet.Builder();
+   }
+
+
+   @AutoValue.Builder
+   public abstract static class Builder {
+      public abstract Builder id(String id);
+      public abstract Builder name(String name);
+      public abstract Builder location(String location);
+      public abstract Builder sku(VirtualMachineScaleSetSKU sku);
+      public abstract Builder tags(Map<String, String> tags);
+      public abstract Builder plan(VirtualMachineScaleSetPlan plan);
+      public abstract Builder properties(VirtualMachineScaleSetProperties properties);
+
+      abstract Map<String, String> tags();
+      abstract VirtualMachineScaleSet autoBuild();
+
+      public VirtualMachineScaleSet build() {
+         tags(tags() != null ? ImmutableMap.copyOf(tags()) : null);
+         return autoBuild();
+      }
+   }
+}
--- a/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetDNSSettings.java
+++ b/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetDNSSettings.java
@@ -1,0 +1,41 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jclouds.azurecompute.arm.domain;
+
+import com.google.auto.value.AutoValue;
+import com.google.common.collect.ImmutableList;
+import org.jclouds.json.SerializedNames;
+
+import java.util.List;
+
+/**
+ * VirtualMachineScaleSetDNSSettings
+ */
+@AutoValue
+public abstract class VirtualMachineScaleSetDNSSettings {
+   /**
+    * The list of DNS servers of the Virtual Machine Scale Set DNS Settings
+    */
+   public abstract List<String> dnsServers();
+
+   @SerializedNames({"dnsServers"})
+   public static VirtualMachineScaleSetDNSSettings create(final List<String> dnsServers) {
+
+      return new AutoValue_VirtualMachineScaleSetDNSSettings(
+         dnsServers == null ? ImmutableList.<String>of() : ImmutableList.copyOf(dnsServers));
+   }
+}
--- a/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetIpConfiguration.java
+++ b/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetIpConfiguration.java
@@ -1,0 +1,61 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jclouds.azurecompute.arm.domain;
+
+import com.google.auto.value.AutoValue;
+import org.jclouds.javax.annotation.Nullable;
+import org.jclouds.json.SerializedNames;
+
+@AutoValue
+public abstract class VirtualMachineScaleSetIpConfiguration {
+
+   /**
+    * The name of the Virtual Machine Scale Set Ip Configuration
+    */
+   @Nullable
+   public abstract String name();
+
+   /**
+    * The properties of the Virtual Machine Scale Set Ip Configuration
+    */
+   @Nullable
+   public abstract VirtualMachineScaleSetIpConfigurationProperties properties();
+
+   @SerializedNames({"name", "properties"})
+   public static VirtualMachineScaleSetIpConfiguration create(
+      final String name,
+      final VirtualMachineScaleSetIpConfigurationProperties properties) {
+      return builder()
+         .name(name)
+         .properties(properties)
+         .build();
+   }
+
+   public abstract Builder toBuilder();
+
+   public static Builder builder() {
+      return new AutoValue_VirtualMachineScaleSetIpConfiguration.Builder();
+   }
+
+   @AutoValue.Builder
+   public abstract static class Builder {
+      public abstract Builder name(String name);
+      public abstract Builder properties(VirtualMachineScaleSetIpConfigurationProperties properties);
+      public abstract VirtualMachineScaleSetIpConfiguration build();
+   }
+}
--- a/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetIpConfigurationProperties.java
+++ b/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetIpConfigurationProperties.java
@@ -1,0 +1,118 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jclouds.azurecompute.arm.domain;
+
+import com.google.auto.value.AutoValue;
+import com.google.common.collect.ImmutableList;
+import org.jclouds.javax.annotation.Nullable;
+import org.jclouds.json.SerializedNames;
+
+import java.util.List;
+
+@AutoValue
+public abstract class VirtualMachineScaleSetIpConfigurationProperties {
+
+   /**
+    * The public IP address configuration of the Virtual Machine Scale Set Ip Configuration Properties
+    */
+   @Nullable
+   public abstract VirtualMachineScaleSetPublicIPAddressConfiguration publicIPAddressConfiguration();
+
+   /**
+    * The subnet of the Virtual Machine Scale Set Ip Configuration Properties
+    */
+   public abstract Subnet subnet();
+
+   /**
+    * The private IP address version of the Virtual Machine Scale Set Ip Configuration Properties
+    */
+   @Nullable
+   public abstract String privateIPAddressVersion();
+
+   /**
+    * The load balancer backend address pools of the Virtual Machine Scale Set Ip Configuration Properties
+    */
+   public abstract List<IdReference> loadBalancerBackendAddressPools();
+
+   /**
+    * The load balancer inbound nat pools of the Virtual Machine Scale Set Ip Configuration Properties
+    */
+   public abstract List<IdReference> loadBalancerInboundNatPools();
+
+   /**
+    * The application gateway backend address Pools of the Virtual Machine Scale Set Ip Configuration Properties
+    */
+   @Nullable
+   public abstract String applicationGatewayBackendAddressPools();
+
+   @SerializedNames({"publicIPAddressConfiguration", "subnet", "privateIPAddressVersion",
+      "loadBalancerBackendAddressPools", "loadBalancerInboundNatPools", "applicationGatewayBackendAddressPools"})
+   public static VirtualMachineScaleSetIpConfigurationProperties create(
+      final VirtualMachineScaleSetPublicIPAddressConfiguration publicIPAddressConfiguration,
+      final Subnet subnet,
+      final String privateIPAddressVersion,
+      final List<IdReference> loadBalancerBackendAddressPools,
+      final List<IdReference> loadBalancerInboundNatPools,
+      final String applicationGatewayBackendAddressPools)
+
+   {
+
+      return builder()
+         .publicIPAddressConfiguration(publicIPAddressConfiguration)
+         .subnet(subnet)
+         .privateIPAddressVersion(privateIPAddressVersion)
+         .lbBackendAddressPools(loadBalancerBackendAddressPools)
+         .lbInboundNatPools(loadBalancerInboundNatPools)
+         .applicationGatewayBackendAddressPools(applicationGatewayBackendAddressPools)
+         .build();
+   }
+
+   public abstract Builder toBuilder();
+
+   public static Builder builder() {
+      return new AutoValue_VirtualMachineScaleSetIpConfigurationProperties.Builder()
+              .lbBackendAddressPools(null)
+              .lbInboundNatPools(null);
+   }
+
+   @AutoValue.Builder
+   public abstract static class Builder {
+      public abstract Builder publicIPAddressConfiguration(VirtualMachineScaleSetPublicIPAddressConfiguration publicIPAddressConfiguration);
+
+      public abstract Builder subnet(Subnet subnet);
+
+      public abstract Builder loadBalancerBackendAddressPools(List<IdReference> loadBalancerBackendAddressPools);
+
+      public abstract Builder loadBalancerInboundNatPools(List<IdReference> loadBalancerInboundNatPools);
+
+      public abstract Builder privateIPAddressVersion(String privateIPAddressVersion);
+
+      public Builder lbBackendAddressPools(List<IdReference> loadBalancerBackendAddressPools) {
+         return loadBalancerBackendAddressPools(loadBalancerBackendAddressPools != null ? ImmutableList
+                 .copyOf(loadBalancerBackendAddressPools) : ImmutableList.<IdReference>of());
+      }
+      public Builder lbInboundNatPools(List<IdReference> loadBalancerInboundNatPools) {
+         return loadBalancerInboundNatPools(loadBalancerInboundNatPools != null ? ImmutableList
+         .copyOf(loadBalancerInboundNatPools) : ImmutableList.<IdReference>of());
+      }
+
+      public abstract Builder applicationGatewayBackendAddressPools(String applicationGatewayBackendAddressPools);
+      public abstract VirtualMachineScaleSetIpConfigurationProperties build();
+
+   }
+}
+
--- a/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetNetworkProfile.java
+++ b/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetNetworkProfile.java
@@ -1,0 +1,58 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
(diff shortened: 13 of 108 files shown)


azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/ExtensionProfile.java
```diff
@@ -0,0 +1,37 @@
...
+
+    /**
+     * The Extensions of the extension profile
+     */
+    public abstract List<Extension> extensions();
+
+
+    @SerializedNames({"extensions"})
+    public static ExtensionProfile create(final List<Extension> extensions) {
+        return new AutoValue_ExtensionProfile(extensions);
```
nacx: Enforce immutability. Also, as the list not declared nullable change to something like:
```java
return new AutoValue_ExtensionProfile(extensions == null ? ImmutableList.of() : ImmutableList.copyOf(extensions));
```

Also review the entire change set to enforce immutability in all collection properties when present.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/Extension.java
```diff
@@ -0,0 +1,40 @@
...
+package org.jclouds.azurecompute.arm.domain;
+
+import com.google.auto.value.AutoValue;
+import org.jclouds.json.SerializedNames;
+
+@AutoValue
+public abstract class Extension {
+
+    /**
+     * The autoUpgradeMinorVersion reference of the extension profile
```
nacx: Is this javadoc accurate?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/ExtensionProfileSettings.java
```diff
@@ -0,0 +1,43 @@
...
+public abstract class ExtensionProfileSettings {
+
+   /**
+    * The fileUris reference of the storage profile
+    */
+   @Nullable
+   public abstract List<String> fileUris();
+
+   /**
+    * The image reference of the storage profile
```
nacx: Wrong javadoc?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/ExtensionProperties.java
```diff
@@ -0,0 +1,66 @@
...
+import org.jclouds.javax.annotation.Nullable;
+import org.jclouds.json.SerializedNames;
+
+import java.util.Map;
+
+@AutoValue
+public abstract class ExtensionProperties {
+
+    /**
+     * The autoUpgradeMinorVersion reference of the extension profile
```
nacx: Review all javadocs in this class.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetIpConfiguration.java
```diff
@@ -0,0 +1,54 @@
...
+ */
+
+package org.jclouds.azurecompute.arm.domain;
+
+import com.google.auto.value.AutoValue;
+import org.jclouds.javax.annotation.Nullable;
+import org.jclouds.json.SerializedNames;
+
+@AutoValue
+public abstract class VirtualMachineScaleSetIpConfiguration {
```
nacx: Are this class and the `VirtualMachineScaleSetIpConfigurationProperties` one different from the existing `IpConfiguration` and `IpConfigurationProperties`?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetNetworkProfile.java
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jclouds.azurecompute.arm.domain;
+
+        import com.google.auto.value.AutoValue;
```
nacx: [minor] Indent.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetNetworkSecurityGroup.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+@AutoValue
+public abstract class VirtualMachineScaleSetNetworkSecurityGroup {
+    /**
+     * The name of the NetworkInterfaceConfiguration
+     */
+    @Nullable
+    public abstract String id();
+
+    @SerializedNames({"dnsSettings"})
```
nacx: Wrong json mapping?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetNetworkSecurityGroup.java
```diff
@@ -0,0 +1,39 @@
...
+
+import com.google.auto.value.AutoValue;
+import org.jclouds.javax.annotation.Nullable;
+import org.jclouds.json.SerializedNames;
+
+/**
+ * VirtualMachineScaleSetNetworkSecurityGroup
+ */
+@AutoValue
+public abstract class VirtualMachineScaleSetNetworkSecurityGroup {
```
nacx: Can't we reuse the existing `NetworkSecurityGroup` class?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetOSProfile.java
```diff
@@ -0,0 +1,267 @@
...
+
+      public abstract Builder linuxConfiguration(LinuxConfiguration linuxConfiguration);
+
+      public abstract Builder windowsConfiguration(WindowsConfiguration windowsConfiguration);
+
+      public abstract Builder secrets(List<Secrets> secrets);
+
+      public abstract VirtualMachineScaleSetOSProfile build();
+   }
+}
```
nacx: is this class really different from the existing `OSProfile`? If it really is, extract the common classes from the existing one so they can be reused.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetPlan.java
```diff
@@ -0,0 +1,51 @@
...
+    @Nullable
+    public abstract String publisher();
+
+    /**
+     * The product of the VirtualMachineScaleSetPlan
+     */
+    @Nullable
+    public abstract String product();
+
+    @SerializedNames({"name", "tier", "capacity"})
```
nacx: Wrong json mapping?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetPublicIPAddressConfiguration.java
```diff
@@ -0,0 +1,56 @@
...
+ * limitations under the License.
+ */
+
+package org.jclouds.azurecompute.arm.domain;
+
+import com.google.auto.value.AutoValue;
+import org.jclouds.json.SerializedNames;
+
+@AutoValue
+public abstract class VirtualMachineScaleSetPublicIPAddressConfiguration {
```
nacx: Same here. Can we use the existing `PublicIpaddress` and its properties class?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetUpgradeMode.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+package org.jclouds.azurecompute.arm.domain;
+
+import org.jclouds.azurecompute.arm.util.GetEnumValue;
+
+public class VirtualMachineScaleSetUpgradeMode {
+    public enum Status {
+        Manual,
+        Automatic,
+        UNRECOGNIZED;
```
nacx: Be consistent with style and use `Unrecognized`;

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/features/VirtualMachineScaleSetApi.java
```diff
@@ -0,0 +1,88 @@
...
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.core.MediaType;
+import java.io.Closeable;
+import java.net.URI;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * The Virtual Machine API includes operations for managing the virtual machines in your subscription.
```
nacx: The "Virtual Machine Scale Sets API"

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/VirtualMachineScaleSetApiLiveTest.java
```diff
@@ -0,0 +1,267 @@
...
+
+import static org.testng.Assert.assertNotNull;
+import static org.testng.Assert.assertTrue;
+
+@Test(groups = "live", testName = "VirtualMachineScaleSetApiLiveTest")
+public class VirtualMachineScaleSetApiLiveTest extends BaseAzureComputeApiLiveTest {
+
+    private String subscriptionid;
+    private String vmssName;
+    private VirtualMachineScaleSetSKU SKU;
```
nacx: Use lowercase for the variable name.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/VirtualMachineScaleSetApiLiveTest.java
```diff
@@ -0,0 +1,267 @@
...
+
+        return VirtualMachineScaleSetNetworkProfile.create(networkInterfaceConfigurations);
+    }
+
+
+    private ExtensionProfile getExtensionProfile() {
+        List<Extension> extensions = new ArrayList<Extension>();
+
+        List<String> uris = new ArrayList<String>();
+        uris.add("https://mystorage1.blob.core.windows.net/winvmextekfacnt/SampleCmd_1.cmd");
```
nacx: Is this something tied to a particular account? Will this prevent the tests from succeeding for any user?

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/VirtualMachineScaleSetApiMockTest.java
```diff
@@ -0,0 +1,271 @@
...
+import org.jclouds.azurecompute.arm.domain.NetworkInterfaceCardProperties;
+import org.jclouds.azurecompute.arm.domain.NetworkInterfaceConfiguration;
+import org.jclouds.azurecompute.arm.domain.NetworkInterfaceConfigurationProperties;
+import org.jclouds.azurecompute.arm.domain.NetworkProfile;
+//import org.jclouds.azurecompute.arm.domain.NetworkProfile.NetworkInterface;
+import org.jclouds.azurecompute.arm.domain.OSDisk;
+//import org.jclouds.azurecompute.arm.domain.OSProfile.LinuxConfiguration;
+//import org.jclouds.azurecompute.arm.domain.OSProfile.WindowsConfiguration.AdditionalUnattendContent;
+//import org.jclouds.azurecompute.arm.domain.OSProfile.WindowsConfiguration.WinRM.Protocol;
+//import org.jclouds.azurecompute.arm.domain.Secrets.SourceVault;
```
nacx: Remove commented code.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/VirtualMachineScaleSetApiMockTest.java
```diff
@@ -0,0 +1,271 @@
...
+import java.util.List;
+import java.util.Map;
+
+//import static com.google.common.collect.Iterables.isEmpty;
+//import static org.jclouds.util.Predicates2.retry;
+//import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertNotNull;
+
+
+@Test(groups = "unit", testName = "VirtualMachineApiMockTest", singleThreaded = true)
```
nacx: Fix test name.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/VirtualMachineScaleSetApiMockTest.java
```diff
@@ -0,0 +1,271 @@
...
+    private VirtualMachineScaleSet getVMSS(VirtualMachineScaleSetPlan plan) {
+        VirtualMachineScaleSet vmss = VirtualMachineScaleSet.create("/subscriptions/SUBSCRIPTIONID/" + ""
+                        + "resourceGroups/groupname/providers/Microsoft.Compute/VirtualMachineScaleSets/virtualmachinescaleset", "windowsmachine",
+                "eastus", getSKU(), null, plan, getVirtualMachineScaleSetProperties());
+        return vmss;
+    }
+
+
+
+    public VirtualMachineScaleSetSKU getSKU() {
```
nacx: All these methods are duplicated. Make them visible in the other class and just reuse them.

Also, this class has no tests. Add the corresponding mock tests.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetIpConfiguration.java
code4clouds: Sadly they are different... I tried it when first wrote the code.  The serialization causes issues when creating the payload.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetNetworkSecurityGroup.java
code4clouds: Sadly they are different... I tried it when first wrote the code. The serialization causes issues when creating the payload.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetPublicIPAddressConfiguration.java
code4clouds: Sadly they are different... I tried it when first wrote the code. The serialization causes issues when creating the payload.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/VirtualMachineScaleSetApiLiveTest.java
code4clouds: I made it upper case to match the payload the server is expecting...

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/VirtualMachineScaleSetApiLiveTest.java
code4clouds: Good point... Do you have location to store a fixed file?  I can check this file first and call it using a RAW GitHub url inside the jclouds repo.

code4clouds:
@nacx Do you take the changes from the amends added or should I create a new PR?

nacx:
>Do you take the changes from the amends added or should I create a new PR?

No need to open a new PR. Thanks for the fixes! I'll review it now.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/Extension.java
```diff
@@ -0,0 +1,40 @@
...
+     * The properties reference of the extension profile
+     */
+    public abstract ExtensionProperties properties();
+
+
+    @SerializedNames({"name", "properties"})
+    public static Extension create(final String name, final ExtensionProperties properties) {
+        return new AutoValue_Extension(name, properties);
+    }
+}
```
nacx: Use 3 space indent.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/ExtensionProfileSettings.java
```diff
@@ -0,0 +1,44 @@
...
+public abstract class ExtensionProfileSettings {
+
+   /**
+    * The fileUris reference of the extension profile settings
+    */
+   @Nullable
+   public abstract List<String> fileUris();
+
+   /**
+    * The commandToExecute of the storage profile settings
```
nacx: "extension profile"

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/NetworkInterfaceConfiguration.java
```diff
@@ -0,0 +1,47 @@
...
+ * limitations under the License.
+ */
+package org.jclouds.azurecompute.arm.domain;
+
+import com.google.auto.value.AutoValue;
+import org.jclouds.javax.annotation.Nullable;
+import org.jclouds.json.SerializedNames;
+
+/**
+ * SKU
```
nacx: Fix or remove this javadoc.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/NetworkInterfaceConfigurationProperties.java
```diff
@@ -0,0 +1,76 @@
...
+
+import com.google.auto.value.AutoValue;
+import com.google.common.collect.ImmutableList;
+import org.jclouds.javax.annotation.Nullable;
+import org.jclouds.json.SerializedNames;
+
+import java.util.List;
+
+/**
+ * NetworkInterfaceConfigurationProperty
```
nacx: Remove this javadoc.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/NetworkInterfaceConfigurationProperties.java
```diff
@@ -0,0 +1,76 @@
...
+   public abstract Boolean enableAcceleratedNetworking();
+
+   /**
+    * The networkSecurityGroup of the NetworkInterfaceConfigurationProperties
+    */
+   @Nullable
+   public abstract VirtualMachineScaleSetNetworkSecurityGroup networkSecurityGroup();
+
+   /**
+    * The dnsSettings of the VirtualMachineScaleSetDNSSettings
```
nacx: Of the "NetworkInterfaceConfigurationProperties"?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/NetworkInterfaceCard.java
```diff
@@ -51,6 +51,7 @@ public static NetworkInterfaceCard create(final String name,
                                              final String location,
                                              final NetworkInterfaceCardProperties properties,
                                              final Map<String, String> tags) {
-      return new AutoValue_NetworkInterfaceCard(name, id, etag, location, properties, tags == null ? null : ImmutableMap.copyOf(tags));
+      return new AutoValue_NetworkInterfaceCard(name, id, etag, location, properties,
+         tags == null ? ImmutableMap.<String, String>of() : ImmutableMap.copyOf(tags));
```
nacx: I think we shouldn't change how the `tags` fields are configured. It is OK to have them `null`(although it is not the default practice). See: https://github.com/jclouds/jclouds-labs/pull/262#r60811082
Could you please revert the changes made to the already existing `tags`fields?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetNetworkSecurityGroup.java
```diff
@@ -0,0 +1,39 @@
...
+   /**
+    * The id of the Virtual Machine Scale Set Network Security Group
+    */
+   @Nullable
+   public abstract String id();
+
+   @SerializedNames({"dnsSettings"})
```
nacx: Is this correct? Is `dnsSettings` the name of the field the war json that maps to the `id` property?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetOSProfile.java
```diff
@@ -0,0 +1,280 @@
...
+
+      @SerializedNames({"provisionVMAgent", "winRM", "additionalUnattendContent", "enableAutomaticUpdates"})
+      public static WindowsConfiguration create(final boolean provisionVMAgent, final WinRM winRM,
+                                                final List<AdditionalUnattendContent> additionalUnattendContent,
+                                                final boolean enableAutomaticUpdates) {
+
+         return new AutoValue_VirtualMachineScaleSetOSProfile_WindowsConfiguration(provisionVMAgent, winRM,
+            additionalUnattendContent, enableAutomaticUpdates);
+      }
+   }
```
nacx: Can't these configuration classes be extracted and reused? They look quite similar to the existing ones.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetOSProfile.java
```diff
@@ -0,0 +1,280 @@
...
+         }
+      }
+
+      /**
+       * The provision VM Agent true of false.
+       */
+      public abstract boolean provisionVMAgent();
+
+      /**
+       * winR
```
nacx: [typo] winRM

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetOSProfile.java
```diff
@@ -0,0 +1,280 @@
...
+
+      public abstract Builder adminUsername(String adminUsername);
+
+      public abstract Builder adminPassword(String adminPassword);
+
+      public abstract Builder linuxConfiguration(LinuxConfiguration linuxConfiguration);
+
+      public abstract Builder windowsConfiguration(WindowsConfiguration windowsConfiguration);
+
+      public abstract Builder secrets(List<Secrets> secrets);
```
nacx: Enforce an immutable collection

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetPlan.java
```diff
@@ -0,0 +1,51 @@
...
+   @Nullable
+   public abstract String publisher();
+
+   /**
+    * The product of the Virtual Machine Scale Set Plan
+    */
+   @Nullable
+   public abstract String product();
+
+   @SerializedNames({"name", "tier", "capacity"})
```
nacx: Wrong json mapping?

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSet.java
```diff
@@ -0,0 +1,110 @@
...
+      public abstract Builder sku(VirtualMachineScaleSetSKU sku);
+      public abstract Builder tags(Map<String, String> tags);
+      public abstract Builder plan(VirtualMachineScaleSetPlan plan);
+      public abstract Builder properties(VirtualMachineScaleSetProperties properties);
+
+      abstract Map<String, String> tags();
+      abstract VirtualMachineScaleSet autoBuild();
+
+      public VirtualMachineScaleSet build() {
+         tags(tags() == null ? ImmutableMap.<String, String>of() : ImmutableMap.copyOf(tags()));
```
nacx: According to my previous comment, keep tags null if null, otherwise, make them immutable.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetProperties.java
```diff
@@ -0,0 +1,60 @@
...
+ * limitations under the License.
+ */
+package org.jclouds.azurecompute.arm.domain;
+
+import com.google.auto.value.AutoValue;
+import org.jclouds.javax.annotation.Nullable;
+import org.jclouds.json.SerializedNames;
+
+/**
+ * SKU
```
nacx: Remove

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetSKU.java
```diff
@@ -0,0 +1,49 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jclouds.azurecompute.arm.domain;
+
+import com.google.auto.value.AutoValue;
+import org.jclouds.json.SerializedNames;
+
+/**
+ * SKU
```
nacx: Remove

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetUpgradePolicy.java
```diff
@@ -0,0 +1,39 @@
...
+ * limitations under the License.
+ */
+package org.jclouds.azurecompute.arm.domain;
+
+import com.google.auto.value.AutoValue;
+import org.jclouds.javax.annotation.Nullable;
+import org.jclouds.json.SerializedNames;
+
+/**
+ * SKU
```
nacx: Remove

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetUpgradeMode.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+package org.jclouds.azurecompute.arm.domain;
+
+import org.jclouds.azurecompute.arm.util.GetEnumValue;
+
+public class VirtualMachineScaleSetUpgradeMode {
+   public enum Status {
+      Manual,
+      Automatic,
+      UNRECOGNIZED;
```
nacx: Be consistent with style and use `Unrecognized`;

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetUpgradePolicy.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+@AutoValue
+public abstract class VirtualMachineScaleSetUpgradePolicy {
+   /**
+    * The name of the Virtual Machine Scale Set Upgrade Policy
```
nacx: "mode"

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/VirtualMachineScaleSetApiMockTest.java
```diff
@@ -0,0 +1,257 @@
...
+      try {
+         assertSent(server, "GET", path);
+      } catch (InterruptedException e) {
+         e.printStackTrace();
+      }
+      return subnet;
+   }
+
+
+}
```
nacx: Tests still missing here.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/VirtualMachineScaleSetApiLiveTest.java
nacx: It could be a file in the `src/test/resources` folder of this provider, if needed.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/NetworkInterfaceConfiguration.java
code4clouds: Missed this one it was wrapped by the IDE.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/Extension.java
code4clouds: fixed...

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/ExtensionProfileSettings.java
code4clouds: fixed...

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetNetworkSecurityGroup.java
code4clouds: fixed...

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/VirtualMachineScaleSetApiMockTest.java
code4clouds: Tests added...

code4clouds:
@nacx All the changes have been done...

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetProperties.java
```diff
@@ -53,8 +51,30 @@ public static VirtualMachineScaleSetProperties create(
...
    }
+
+   @AutoValue.Builder
+   public abstract static class Builder {
+      public abstract Builder singlePlacementGroup(Boolean singlePlacementGroup);
+      public abstract Builder overProvision(Boolean overProvision);
+      public abstract Builder upgradePolicy(VirtualMachineScaleSetUpgradePolicy upgradePolicy);
+      public abstract Builder virtualMachineProfile(VirtualMachineScaleSetVirtualMachineProfile virtualMachineProfile);
+
+      abstract VirtualMachineScaleSetProperties autoBuild();
```
nacx: This is only needed if you are going to change how the "build" method works. You can remove it and leave just the `public abstract VirtualMachineScaleSetProperties build();`.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/VirtualMachineScaleSetApiMockTest.java
nacx: Mock tests are intended to verify that the API methods produce the expected requests. We test the API against a mock webserver that records the requests it receives, and then we assert that the expected ones were generated.
The tests in this class must exercise all the methods in the API (list, createOrUpdate, get, delete), and then validate that all produced requests are the expected ones.
You can use the [DiskApiMockTest](https://github.com/jclouds/jclouds-labs/blob/master/azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/DiskApiMockTest.java) as a model.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/NetworkSecurityGroup.java
```diff
@@ -43,7 +43,7 @@
    @SerializedNames({ "id", "name", "location", "tags", "properties", "etag" })
    public static NetworkSecurityGroup create(final String id, final String name, final String location,
          final Map<String, String> tags, final NetworkSecurityGroupProperties properties, final String etag) {
-      return new AutoValue_NetworkSecurityGroup(id, name, location, (tags == null) ? null : ImmutableMap.copyOf(tags),
+      return new AutoValue_NetworkSecurityGroup(id, name, location, (tags == null) ? ImmutableMap.<String, String>of() : ImmutableMap.copyOf(tags),
```
nacx: Don't initialize to empty when null.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/ExtensionProfileSettings.java
```diff
@@ -0,0 +1,44 @@
...
+
+import java.util.List;
+
+@AutoValue
+public abstract class ExtensionProfileSettings {
+
+   /**
+    * The fileUris reference of the extension profile settings
+    */
+   @Nullable
```
nacx: You are enforcing presence and immutability. Remove this annotation.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/ExtensionProperties.java
```diff
@@ -0,0 +1,72 @@
...
+
+   /**
+    * The ExtensionProfileSettings of the extension properties
+    */
+   public abstract ExtensionProfileSettings settings();
+
+   /**
+    * The list of the protectedSettings of the extension properties
+    */
+   @Nullable
```
nacx: You are enforcing presence and immutability. Remove this annotation.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/NetworkInterfaceCardProperties.java
```diff
@@ -32,15 +32,17 @@
...
    public static NetworkInterfaceCardProperties create(final String provisioningState, final String resourceGuid,
          final Boolean enableIPForwarding, final List<IpConfiguration> ipConfigurations,
          final IdReference networkSecurityGroup) {
       NetworkInterfaceCardProperties.Builder builder = NetworkInterfaceCardProperties.builder()
               .provisioningState(provisioningState)
               .resourceGuid(resourceGuid)
               .enableIPForwarding(enableIPForwarding)
-              .ipConfigurations(ipConfigurations == null ? null : ImmutableList.copyOf(ipConfigurations))
+              .ipConfigurations(ipConfigurations == null ?
+                 ImmutableList.<IpConfiguration>of() : ImmutableList.copyOf(ipConfigurations))
```
nacx: Don't change the behavior from null to an empty list. Leave it null, as it is not related to the changes in this PR.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/NetworkInterfaceConfigurationProperties.java
```diff
@@ -0,0 +1,74 @@
...
+   /**
+    * The dnsSettings of the NetworkInterfaceConfigurationProperties
+    */
+   @Nullable
+   public abstract VirtualMachineScaleSetDNSSettings dnsSettings();
+
+   /**
+    * The ipConfigurations of the NetworkInterfaceConfigurationProperties
+    */
+   @Nullable
```
nacx: You are enforcing presence and immutability. Remove this annotation.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetDNSSettings.java
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * VirtualMachineScaleSetDNSSettings
+ */
+@AutoValue
+public abstract class VirtualMachineScaleSetDNSSettings {
+   /**
+    * The list of DNS servers of the Virtual Machine Scale Set DNS Settings
+    */
+   @Nullable
```
nacx: You are enforcing presence and immutability. Remove this annotation.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetIpConfigurationProperties.java
```diff
@@ -0,0 +1,121 @@
...
+   /**
+    * The private IP address version of the Virtual Machine Scale Set Ip Configuration Properties
+    */
+   @Nullable
+   public abstract String privateIPAddressVersion();
+
+   /**
+    * The load balancer backend address pools of the Virtual Machine Scale Set Ip Configuration Properties
+    */
+   @Nullable
```
nacx: You are enforcing presence and immutability. Remove this annotation.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetIpConfigurationProperties.java
```diff
@@ -0,0 +1,121 @@
...
+   public abstract List<IdReference> loadBalancerBackendAddressPools();
+
+   /**
+    * The load balancer inbound nat pools of the Virtual Machine Scale Set Ip Configuration Properties
+    */
+   @Nullable
```
nacx: You are enforcing presence and immutability. Remove this annotation.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetOSProfile.java
```diff
@@ -0,0 +1,280 @@
...
+      /**
+       * winRM
+       */
+      @Nullable
+      public abstract WinRM winRM();
+
+      /**
+       * unattend content
+       */
+      @Nullable
```
nacx: You are enforcing presence and immutability. Remove this annotation.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetOSProfile.java
```diff
@@ -0,0 +1,280 @@
...
+   /**
+    * The windows configuration of the VM
+    */
+   @Nullable
+   public abstract WindowsConfiguration windowsConfiguration();
+
+   /**
+    * The Secrets configuration of the VM
+    */
+   @Nullable
```
nacx: You are enforcing presence and immutability. Remove this annotation.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetOSProfile.java
```diff
@@ -0,0 +1,280 @@
...
+   public static VirtualMachineScaleSetOSProfile create(final String computerNamePrefix, final String adminUsername,
+                                                        final String adminPassword, final LinuxConfiguration linuxConfiguration,
+                                                        final WindowsConfiguration windowsConfiguration, final List<Secrets> secrets) {
+      return builder()
+         .computerNamePrefix(computerNamePrefix)
+         .adminUsername(adminUsername)
+         .adminPassword(adminPassword)
+         .linuxConfiguration(linuxConfiguration)
+         .windowsConfiguration(windowsConfiguration)
+         .secrets(secrets == null ? ImmutableList.<Secrets>of() : ImmutableList.copyOf(secrets) )
```
nacx: Move this logic to the builder itself.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetPublicIPAddressConfiguration.java
```diff
@@ -0,0 +1,66 @@
...
+   }
+
+   @AutoValue.Builder
+   public abstract static class Builder {
+
+      public abstract Builder name(String name);
+
+      public abstract Builder properties(VirtualMachineScaleSetPublicIPAddressProperties properties);
+
+      abstract VirtualMachineScaleSetPublicIPAddressConfiguration autoBuild();
```
nacx: Same here; this method is not needed.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/ExtensionProfileSettings.java
code4clouds: These @Nullable annotations are required to serialized the json response from the service (especially the when the value is not returned as sometimes there are not setup by the users)

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/ExtensionProperties.java
code4clouds: These @Nullable annotations are required to serialized the json response from the service (especially the when the value is not returned as sometimes there are not setup by the users)

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/NetworkInterfaceConfigurationProperties.java
code4clouds: These @Nullable annotations are required to serialized the json response from the service (especially the when the value is not returned as sometimes there are not setup by the users)

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetDNSSettings.java
code4clouds: These @Nullable annotations are required to serialized the json response from the service (especially the when the value is not returned as sometimes there are not setup by the users)

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetIpConfigurationProperties.java
code4clouds: These @Nullable annotations are required to serialized the json response from the service (especially the when the value is not returned as sometimes there are not setup by the users)

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetOSProfile.java
code4clouds: These @Nullable annotations are required to serialized the json response from the service (especially the when the value is not returned as sometimes there are not setup by the users)

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetOSProfile.java
code4clouds: These @Nullable annotations are required to serialized the json response from the service (especially the when the value is not returned as sometimes there are not setup by the users)

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetPublicIPAddressConfiguration.java
code4clouds: fixed in last ammend

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetProperties.java
code4clouds: fixed in last ammend

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetOSProfile.java
code4clouds: moved to the builders.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/NetworkSecurityGroup.java
code4clouds: fixed...

code4clouds:
@nacx  Added your changes... I have to leave the @nullable these are per spec and needed to properly serialize the response from the service.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/ExtensionProfileSettings.java
nacx: In our domain objects, implemented with Google AutoValue, the `@Nullable` annotation is taken into account inside the `AutoValue_` constructor. In these cases, a `null` value will never get there, since we're always calling it with a non-null value. The questions then re:

* If you need to be able to serialize this object and send it to the Azure API **without** this field, then we need to follow the same approach we use with the `tags`, and remove the annotation and initialize to "null or immutable", as we want to explicitly support null.
* If there is no such a need of **sending** a null value, and it's only us reading it, then we should think about how we want a Java API to look like for users:
  * Null collections is an anti-pattern, so we'd better enforce presence and immutability. This helps to build a cleaner and easier to use API.
  * In this case, as we're enforcing presence in the constructor, the annotation is not required (if I'm not wrong), as it is only used by Google AutoValue (and not for serialization/deserialization) and we're already taking care of always passing a value.

code4clouds: @nacx There are two reasons why these @nulleable are valid and should stay:

1-If the nullable is removed the tests fail as there aren't any value in this case (as real response) null is appropriate.  It is not really the anti-pattern case for this, because null is not used to handle exceptions.
2- The @nulleable came from previous developed/approved code already used by our jclouds customers.  We will be having two different implementations under the same package.

nacx: >If the nullable is removed the tests fail as there aren't any value in this case (as real response) null is appropriate. It is not really the anti-pattern case for this because null is not used to handle exceptions.

I've built your branch and removed the annotation and tests don't fail.
A jclouds API is **not** an exact mapping to existing REST APIs. We go one step further to provide a meaningful experience when consuming an API from Java. We take care of properly mapping response error codes to appropriate exceptions, building good transfer objects, etc. Part of this work is helping users use clean APIs (from a Java perspective), and that means reducing the changes where users can make mistakes. Having `nullable` lists is an anti-pattern that should be avoided, as empty lists are a perfect substitute and make much more sense.

My point is that **your code is already enforcing that**. The code in the `create` method already enforces an empty list when a `null` value (or absent value) cames in the response. With the code in this factory method, **there is no way** an instance of an `ExtensionProfileSettings` object can have a null value in this field, thus, the `@Nullable` annotation is simply not true, as users will never get a `null` value in these objects. Please, remove it and apply this pattern in all the mentioned domain objects.

As I said, the rule of thumb for a consistent implementation and declaration is:

* If your constructor/factory methods enforce presence (as we do in most cases), the `@Nullable` annotation does not apply and should not be there.
* If there is a good reason to not enforce presence in the constructor/factory method (like in tags), then let's not enforce presence and annotate the field accordingly.

Here is a diff you can apply to verify that there are no serialization/deserialization or other issues if you remove the annotation, even if the response does not contain a value for that field. Tests keep passing.

```diff
diff --git a/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/ExtensionProfileSettings.java b/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/ExtensionProfileSettings.java
index 0d89fda3e..61d50d910 100644
--- a/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/ExtensionProfileSettings.java
+++ b/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/ExtensionProfileSettings.java
@@ -16,12 +16,12 @@
  */
 package org.jclouds.azurecompute.arm.domain;
 
-import com.google.auto.value.AutoValue;
-import com.google.common.collect.ImmutableList;
-import org.jclouds.javax.annotation.Nullable;
+import java.util.List;
+
 import org.jclouds.json.SerializedNames;
 
-import java.util.List;
+import com.google.auto.value.AutoValue;
+import com.google.common.collect.ImmutableList;
 
 @AutoValue
 public abstract class ExtensionProfileSettings {
@@ -29,7 +29,6 @@ public abstract class ExtensionProfileSettings {
    /**
     * The fileUris reference of the extension profile settings
     */
-   @Nullable
    public abstract List<String> fileUris();
 
    /**
diff --git a/azurecompute-arm/src/test/resources/createvirtualmachinescalesetresponse.json b/azurecompute-arm/src/test/resources/createvirtualmachinescalesetresponse.json
index ff2971d40..4b118c50a 100644
--- a/azurecompute-arm/src/test/resources/createvirtualmachinescalesetresponse.json
+++ b/azurecompute-arm/src/test/resources/createvirtualmachinescalesetresponse.json
@@ -42,7 +42,7 @@
               "type": "CustomScriptExtension",
               "typeHandlerVersion": "1.1",
               "autoUpgradeMinorVersion": false,
-              "settings": {"fileUris":["https://mystorage1.blob.core.windows.net/winvmextekfacnt/SampleCmd_1.cmd"],"commandToExecute":"SampleCmd_1.cmd"}
+              "settings": {"commandToExecute":"SampleCmd_1.cmd"}
             },
             "name": "extensionName"
           }
diff --git a/azurecompute-arm/src/test/resources/virtualmachinescaleset.json b/azurecompute-arm/src/test/resources/virtualmachinescaleset.json
index c99eb39a5..720f71ff8 100644
--- a/azurecompute-arm/src/test/resources/virtualmachinescaleset.json
+++ b/azurecompute-arm/src/test/resources/virtualmachinescaleset.json
@@ -42,7 +42,7 @@
               "type": "CustomScriptExtension",
               "typeHandlerVersion": "1.1",
               "autoUpgradeMinorVersion": false,
-              "settings": {"fileUris":["https://mystorage1.blob.core.windows.net/winvmextekfacnt/SampleCmd_1.cmd"],"commandToExecute":"SampleCmd_1.cmd"}
+              "settings": {"commandToExecute":"SampleCmd_1.cmd"}
             },
             "name": "extensionName"
           }
```

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetIpConfigurationProperties.java
nacx: Still needs to be removed

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetIpConfigurationProperties.java
nacx: Still needs to be removed

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/VirtualMachineScaleSetApiLiveTest.java
```diff
@@ -0,0 +1,267 @@
...
+
+      vmssName = String.format("%3.24s", System.getProperty("user.name") + RAND + this.getClass().getSimpleName()).toLowerCase().substring(0, 15);
+   }
+
+   @Test
+   public void testCreate() {
+      VirtualMachineScaleSet vmss = api().createOrUpdate(vmssName, LOCATIONDESCRIPTION, getSKU(),
+         Collections.<String, String>emptyMap(), getProperties());
+      assertTrue(!vmss.name().isEmpty());
+   }
```
nacx: Add tests for the list, get and delete methods too, and provide an `@AfterClass(alwaysRun = true)` method that deletes all the stuff created by this test.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/VirtualMachineScaleSetApiMockTest.java
```diff
@@ -0,0 +1,116 @@
...
+   }
+
+   public void testVMSSCreate_NetworkProfile() throws InterruptedException {
+      server.enqueue(jsonResponse("/createvirtualmachinescalesetresponse.json").setResponseCode(200));
+      final VirtualMachineScaleSetApi vmssAPI = api.getVirtualMachineScaleSetApi(resourcegroup);
+      assertEquals(vmssAPI.get("jclouds-vmssname").properties().virtualMachineProfile()
+         .networkProfile()
+         .networkInterfaceConfigurations()
+         .size(), 1);
+   }
```
nacx: These mock tests still don't meet their purpose:

* There should be only **one** test for each method in the API to verify that the method generates the expected HTTP request. Verifying that means the test method should:
  - [x] Validate the returned object (this verifies serialization)
  - [ ] Validate the generated requests using the `assertSent` method. This validates that method invocation generated the expected HTTP request (expected URI, headers, and body).
* If the API method defines a `@Fallback`, add another test for that method that exercises it.
  - [ ] For all methods with fallbacks, add a test that returns the response code that triggers the fallback, and verify that the method returns the expected result (usually `null`, empty lists, etc) instead of throwing an exception.

In summary, this test class should just have the following test methods: `testList`, `testListWhen404`, `testGet`, `testGetWhen404`, `testCreateOrUpdate`, `testDelete` and `testDeleteWhen404`.

The [AvailabilitySetApiMockTest](https://github.com/jclouds/jclouds-labs/blob/master/azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/AvailabilitySetApiMockTest.java) is a good example to follow.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetIpConfigurationProperties.java
code4clouds: Hi @nacx  I should have been more clear and place the mock error in the right comment section.  This is the file with the @nullable that I was referring when talking about the mocks failing.  I removed the others as you suggested, but the ones in this file seems they need to stay.

This is the error:
```testVMSSGet_StorageProfile(org.jclouds.azurecompute.arm.features.VirtualMachineScaleSetApiMockTest)  Time elapsed: 0.035 sec  <<< FAILURE!
org.jclouds.http.HttpResponseException: Error parsing input: Null loadBalancerInboundNatPools```

Comment (code4clouds) [reply azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/VirtualMachineScaleSetIpConfigurationProperties.java]: Hi @nacx  I should have been more clear and place the mock error in the right comment section.  This is the file with the @nullable that I was referring when talking about the mocks failing.  I removed the others as you suggested, but the ones in this file seems they need to stay.

This is the error:
```testVMSSGet_StorageProfile(org.jclouds.azurecompute.arm.features.VirtualMachineScaleSetApiMockTest)  Time elapsed: 0.035 sec  <<< FAILURE!
org.jclouds.http.HttpResponseException: Error parsing input: Null loadBalancerInboundNatPools```

File: azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/VirtualMachineScaleSetApiLiveTest.java (hunk 50 above)
Comment (code4clouds) [reply azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/VirtualMachineScaleSetApiLiveTest.java]: That's been done on the BaseAzureComputeApiLiveTest.java using the tearDown().  The system cleans after the live test is complete...

Comment (nacx) [reply azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/VirtualMachineScaleSetApiLiveTest.java]: Yes, but relying on that you're not testing the list, get, and delete methods. It is OK to rely on that teardown, but the methods in the API must have explicit tests.

File: azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/VirtualMachineScaleSetApiMockTest.java [hunk 52]
```diff
@@ -0,0 +1,488 @@
...
+         .networkProfile()
+         .networkInterfaceConfigurations()
+         .size(), 1);
+
+      assertSent(server,
+              "GET",
+              "/subscriptions/SUBSCRIPTIONID/resourceGroups/myresourcegroup/providers/" +
+                      "Microsoft.Compute"
+              + "/VirtualMachineScaleSets/" + vmssname + "?api-version=2017-03-30");
+   }
```
nacx: All these additional tests are redundant since they call the same "get" API method than the one tested in the "testGet" test. If you want to validate additional fields from the returned object, are more assertions to the "testGet" instead of adding one test for each property you want to validate.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/VirtualMachineScaleSetApiMockTest.java
```diff
@@ -0,0 +1,488 @@
...
+      final VirtualMachineScaleSetApi vmssAPI = api.getVirtualMachineScaleSetApi(resourcegroup);
+      VirtualMachineScaleSet vmss = CreateOrUpdateVMSS(vmssAPI);
+
+      assertNotNull(vmss);
+      assertSent(server,
+              "PUT",
+              "/subscriptions/SUBSCRIPTIONID/resourceGroups/myresourcegroup/providers/" +
+                      "Microsoft.Compute"
+                      + "/VirtualMachineScaleSets/" + vmssname + "?api-version=2017-03-30");
+   }
```
nacx: For requests that generate a body you also need to verify it. Check how it is done in the example test class I provided.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/VirtualMachineScaleSetApiMockTest.java
```diff
@@ -0,0 +1,488 @@
...
+                                              "False",
+                                              null),
+                                      null,
+                                      new ArrayList<Secrets>()),
+                              getNetworkProfile(),
+                              getExtensionProfile()
+
+                      )
+              ));
+   }
```
nacx: Same here. Use just one "testCreate" complete enough to validate all properties you want to validate and remove all the redundant ones.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/VirtualMachineScaleSetApiMockTest.java
```diff
@@ -0,0 +1,488 @@
...
+              "DELETE",
+              "/subscriptions/SUBSCRIPTIONID/resourceGroups/myresourcegroup/providers/" +
+                      "Microsoft.Compute"
+              + "/VirtualMachineScaleSets/" + vmssname + "?api-version=2017-03-30");
+   }
+
+   public void testDelete() throws InterruptedException {
+      server.enqueue(response202WithHeader());
+      final VirtualMachineScaleSetApi vmssAPI = api.getVirtualMachineScaleSetApi(resourcegroup);
+      vmssAPI.delete(vmssname);
```
nacx: Verify that returns a non-null URI.

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/VirtualMachineScaleSetApiMockTest.java
```diff
@@ -0,0 +1,488 @@
...
+                      "Microsoft.Compute"
+              + "/VirtualMachineScaleSets?api-version=2017-03-30");
+      assertTrue(isEmpty(vmssList));
+   }
+
+
+   public void testDeleteWhen404() throws InterruptedException {
+      server.enqueue(jsonResponse("/virtualmachinescalesetlist.json").setResponseCode(404));
+      final VirtualMachineScaleSetApi vmssAPI = api.getVirtualMachineScaleSetApi(resourcegroup);
+      vmssAPI.delete(vmssname);
```
nacx: Verify that returns null

azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/VirtualMachineScaleSetApiLiveTest.java
nacx: Tests for the list, get and delete operations are still missing.

nacx:
Many thanks @code4clouds for all the fixes and the continued effort. Ready to merge!
Mind squashing all the commits into a single one and get it rebased to the latest version of master so I can cleanly merge the PR? Thanks!

code4clouds:
@nacx I merged and rebased the code... 

You should be able to do the squash and merge from within GitHub.  It now does it automatically with a single button (https://github.com/blog/2141-squash-your-commits).

nacx:
I know @code4clouds, but I always prefer the contributors to do it, so they can write their commit message and properly explain the contents of the contribution in a single commit.

code4clouds:
@nacx I squashed and re-based the code.  Good to merge?

nacx:
@code4clouds the PR still has 89 commits, whilst after squashing and rebasing it should have just one.

code4clouds:
@nacx If you have a procedure to squash for this project, it would be nice if you can share it.  I spent all night merging, squashing and picking before submitting.  I noticed previous PRs had their history, so its hard to know what/why are you seen them.

code4clouds:
@nacx Has Jenkin changed... RAT is complaining on the POM and the file has not changed.

```*****************************************************
Summary
-------
Generated at: 2017-11-13T20:47:34-05:00

Notes: 0
Binaries: 0
Archives: 0
Standards: 1

Apache Licensed: 1
Generated Documents: 0

JavaDocs are generated, thus a license header is optional.
Generated files do not require license headers.

0 Unknown Licenses

*****************************************************
  Files with Apache License headers will be marked AL
  Binary files (which do not require any license headers) will be marked B
  Compressed archives will be marked A
  Notices, licenses etc. will be marked N
  AL    pom.xml
 
*****************************************************
```

code4clouds:
@nacx How many commits do you see on the last three I submitted?

code4clouds:
Would it be better to close this PR, create a new one and patch it?

nacx:
This is a good and small tutorial on rebasing and squashing commits: https://help.github.com/articles/about-git-rebase/

I'm not confident on rebasing the PR myself, as it now lists 103 commits and I don't know exactly which ones should I cherry-pick to properly merge the feature.

I'd suggest you proceed however works best for you: if you want to try rebasing the PR and reduce it to a single clean commit, it is OK. If you find it easier to close this PR and open a new one starting from master, it is OK too. In the end, we just want to have this merged in a single commit, so the commit history cleanly reflects the commit where the feature was introduced.

code4clouds:
@nacx I think this is it... Finally!  :-)

nacx:
Merged to [master](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/30a500f8) and [2.0.x](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/96fca7eb).
The commit in this PR contained changes not related to this PR (it has 108 files) because the branch was not properly rebased. I've rebased it before merging and the commit includes only the relevant changes.
Thanks @code4clouds!

andreaturli:
Thanks @code4clouds for the hard work. And thanks @nacx for the great help!

PR #415, opened 2017-10-26
Azure ARM fix OSProfile domain object

andreaturli:


andreaturli:
merged at [master](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/5fdfd58f)

PR #417, opened 2017-11-06
Add tags to virtual network creation

nacx:


azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/strategy/CreateResourcesThenCreateNodes.java
```diff
@@ -159,7 +159,7 @@ protected synchronized void createDefaultNetworkIfNeeded(String group, String lo
          logger.debug(">> network options have not been configured. Creating network %s(%s) and subnet %s(%s)", name,
                defaultVnetAddressPrefix, name, defaultSubnetAddressPrefix);
          
-         api.getVirtualNetworkApi(options.getResourceGroup()).createOrUpdate(name, location, properties);
+         api.getVirtualNetworkApi(options.getResourceGroup()).createOrUpdate(name, location, null, properties);
```
andreaturli: out of curiosity, how do you plan to use the `tags` is the `CreateResourceThenCreateNodes` has them set to null?

nacx: Using the provider-specific API. Without this change, there is no way to configure the tags for a network.

andreaturli: sure, that makes sense!
would it make sense to think about a default strategy to tag the `virtualNetwork` so the clean up resources step can take advantage of it?

nacx: Yes, We could do that and follow the same approach than we use for the availability sets (although leaving a default network that will be reused per resource group is not a big deal). Anyway, stuff for a different PR, right?

andreaturli: yes, for a subsequent PR!

nacx:
Pushed to [master](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/c9cd6eaf) and [2.0.x](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/fd22d3ad).

PR #421, opened 2017-11-27
Implement OsImageToImage Function

trevorflanagan:
This PR shows the approach we will use for the next steps involved in creating the JCLouds abstraction now that the feature APIs are complete.


dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/compute/functions/ImageDescriptionToOsFamily.java
```diff
@@ -0,0 +1,49 @@
...
+   public OsFamily apply(final String description) {
+      if (description != null) {
+         if (description.contains(CENTOS))
+            return OsFamily.CENTOS;
+         else if (description.contains(UBUNTU))
+            return OsFamily.UBUNTU;
+         else if (description.contains(REDHAT))
+            return OsFamily.RHEL;
+         else if (description.contains(SUSE))
+            return OsFamily.RHEL;
```
nacx: Shouldn't it be `OsFamily.SUSE`?

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/compute/functions/OsImageToImage.java
```diff
@@ -0,0 +1,87 @@
...
+   @Override
+   public Image apply(OsImage input) {
+
+      OsFamily osFamily = imageDescriptionToOsFamily.apply(input.description());
+      String osVersion = parseVersion(input.description());
+
+      OperatingSystem os = OperatingSystem.builder().name(input.name()).description(input.description())
+            .family(osFamily).version(osVersion).is64Bit(is64bit(input)).build();
+
+      return new ImageBuilder().id(input.id()).name(input.name()).description(input.description())
```
nacx: When configuring the image IDs, think about the "id" and the "providerId" field this way:

* **providerId** is the *real id* of the image in the provider (I guess `input.id()`), so users can retrieve the image using the provider-specific APIs.
* **id** is the *portable id jclouds will use to uniquelly identify the image*. This is slightly different from the previous one. For example, in some providers, images in different regions may have the same id (centos-6, ami-0001, etc), but jcouds needs to provide unique IDs for each, so we encode the region in the id (something like: "region/id").

I don't know if it is the case in DimensionData, but please consider this when configuring the id and providerId fields. If the ID is unique, then just assign the same value to both fields :)

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/compute/functions/OsImageToImage.java
```diff
@@ -0,0 +1,87 @@
...
+            .status(Image.Status.AVAILABLE).operatingSystem(os).location(
+                  FluentIterable.from(locations.get()).firstMatch(LocationPredicates.idEquals(input.datacenterId()))
```
nacx: Just one consideration about the meaning of the `location` field in an image:

* If the image is globally available in all locations (the same image, same id, etc), then the location field should be `null`.
* if the image is only available in certain locations, then the location field must be set accordingly.

dimensiondata/src/test/java/org/jclouds/dimensiondata/cloudcontrol/compute/functions/ImageDescriptionToOsFamilyTest.java
```diff
@@ -0,0 +1,70 @@
...
+   @Test
+   public void apply_Centos() {
+      String description = "CentOS 6.0 (Final)";
+      assertEquals(OsFamily.CENTOS, imageDescriptionToOsFamily.apply(description));
+   }
+
+   @Test
+   public void apply_Suse() {
+      String description = "SuSE Linux Enterprise Server 11 SP4";
+      assertEquals(OsFamily.RHEL, imageDescriptionToOsFamily.apply(description));
```
nacx: `OsFamily.SUSE`?

dimensiondata/src/test/java/org/jclouds/dimensiondata/cloudcontrol/compute/functions/ImageDescriptionToOsFamilyTest.java
```diff
@@ -0,0 +1,70 @@
...
+ */
+package org.jclouds.dimensiondata.cloudcontrol.compute.functions;
+
+import org.jclouds.compute.domain.OsFamily;
+import org.testng.annotations.BeforeMethod;
+import org.testng.annotations.Test;
+
+import static org.testng.AssertJUnit.assertEquals;
+
+public class ImageDescriptionToOsFamilyTest {
```
nacx: Configure the test in the "unit" group.

trevorflanagan:
@nacx thanks for taking a look. I will make the required changes shortly.

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/compute/functions/OsImageToImage.java
trevorflanagan: @nacx thanks for the background information. In DimensionData each image is associated with a single datacenter (zone) and will the ID will be unique within a Region. Therefore I will keep the `input.id()` as is.

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/compute/functions/OsImageToImage.java
trevorflanagan: Image is only available in certain locations, so what I have here is valid.

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/compute/functions/OsImageToImage.java
nacx: Is it possible that that unique ID within that region collisions with a unique id within another region?

nacx: I mean, you should think about it this way: calling `ComputeService.getImage(id)` with this ID will succeed? Or should the ID carry/encode additional info?

nacx: Perfect! Thanks for clarifying. Then, mind just changing the code to `ImageBuilder().ids(input.id())` (note the plural there)? This will set both the id and providerId fields.

trevorflanagan:
Hi @nacx I would like to make some further changes for the PR. Since you are busy with the release I think it's safe to do so.

nacx:
Sure! :smile:

trevorflanagan:
@nacx I have refactored OsImageToImage in order to handle OsImage and CustomerImage. This means that JClouds can expose all of our image types. I have also improved the version and os family handling to be more robust.

trevorflanagan:
rebuild please

nacx:
Merged to [master](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/2b8bfcb8) and [2.0.x](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/8c1449bd). Thanks @trevorflanagan!

PR #422, opened 2017-11-29
Implementing BaseImageToHardware and CleanupServer functions.

trevorflanagan:


dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/compute/functions/CleanupServer.java
```diff
@@ -0,0 +1,153 @@
...
+
+   @Resource
+   @Named(ComputeServiceConstants.COMPUTE_LOGGER)
+   protected Logger logger = Logger.NULL;
+
+   private final DimensionDataCloudControlApi api;
+   private final Timeouts timeouts;
+
+   @Inject
+   public CleanupServer(final DimensionDataCloudControlApi api, final Timeouts timeouts) {
```
nacx: Remove the public modifier

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/compute/functions/CleanupServer.java
```diff
@@ -0,0 +1,153 @@
...
+               .filter(new Predicate<NatRule>() {
+                  @Override
+                  public boolean apply(NatRule natRule) {
+                     return natRule.internalIp().equals(internalIp);
+                  }
+               }).toList();
+
+         for (final NatRule natRule : natRulesToBeDeleted) {
+            if (natRule.state().isNormal()) {
+               networkApi.deleteNatRule(natRule.id());
```
nacx: Surround with a try/catch and log a warning, for a best-effort cleanup? Otherwise, it will stop deleting stuff here if this call fails. (Apply the same criteria to the other cleanup calls).

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/compute/functions/CleanupServer.java
```diff
@@ -0,0 +1,153 @@
...
+
+      if (server == null) {
+         return true;
+      }
+
+      if (server.state().isFailed()) {
+         rollbackOperation(format("Server(%s) not deleted as it is in state(%s).", serverId, server.state()));
+      }
+
+      if (server.state().isNormal()) {
```
nacx: Just return false if it is not normal and continue the rest of the method "outside" the conditional. It is easier to read.

trevorflanagan: Much tidier, I will update.

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/compute/functions/CleanupServer.java
trevorflanagan: Makes sense to cleanup as much as possible without failing the full operation, I will update to have this, but I am not sure about the need for a try/catch.

nacx: Calls to the API can fail for a variety of reasons, say there is a conflict, a transient, error, whatever unexpected error that makes this raise an `HttpResponseException`. The `CleanupServer` is likely to be used when users want to destroy nodes and associated resources to "stop paying for them". We should be as defensive as possible and try to continue deleting stuff even if an unexpected exception occurred when trying to delete intermediate resources.

trevorflanagan: Understood, thanks for clarifying. I will make that change.

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/compute/functions/CleanupServer.java
```diff
@@ -80,70 +80,104 @@ public Boolean apply(final String serverId) {
...
+            networkApi.deleteNatRule(natRule.id());
+         } else {
+            logger.warn(format("Server(%s) has an associated NatRule(%s) that was not deleted as it was in state(%s).",
+                  serverId, natRule.id(), natRule.state()));
+         }
+      } catch (Throwable t) {
+         logger.warn(
+               format("Failed to delete NatRule(%s) associated with Server(%s). Due to - (%s)", natRule.id(), serverId,
+                     t.getMessage()));
+      }
```
nacx: Better pass the `Throwable` as the first parameter to the logger to have the complete stacktrace logged. Often just the exception message does not give enough contex.

nacx:
Merged to [master](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/172d6f34) and [2.0.x](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/caf4c19a). Thanks @trevorflanagan!
I had to remove [an unnecessary call to `EasyMockSupport.injectMocks`](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/b4998732) to make it compile in 2.0.x (probably due to the easyMock version), and I've also applied it to master.

PR #424, opened 2017-12-01
Injectable current service principal

nacx:
This PR exposes the current service principal in the Azure API and provides several improvements to allow APIs interact with the current user information and the different OAuth resources.

* It uses the *Graph RBAC* API to retrieve the current Service Principal.
* If binds it to the Guice context to make it injectable as a `Supplier<ServicePrincipal>`. This way it is available where needed (for example in the `KeyVault` classes, to get its `obejctId`).
* It also makes the `tenantId` injectable as: `@Tenant String tenantId`.
* Provides a generic way to specify custom OAuth resources for APIs classes and concrete API methods. To use a custom OAuth resource in an API call just annotate the API or method with `@OAuthResource("custom resource")` and authentication will automatically use the custom one.

@jmspring I think this will be very useful for the KeyVault pull request. Could you give this approach a try? In concrete:

* You'll be able to remove the need to ask for the `objectId` in the properties, as now it is automatically resolved and you can inject the service principal supplier wherever you need the `objectId`.
* It also allows you to configure the entire KeyVault API with the custom OAuth endpoint annotation for a clean OAuth configuration.

### DIFF
--- a/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/config/AzureComputeHttpApiModule.java
+++ b/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/config/AzureComputeHttpApiModule.java
@@ -17,7 +17,6 @@
 package org.jclouds.azurecompute.arm.config;
 
 import static org.jclouds.Constants.PROPERTY_SESSION_INTERVAL;
-import static org.jclouds.rest.config.BinderUtils.bindHttpApi;
 
 import java.net.URI;
 import java.util.concurrent.TimeUnit;
@@ -26,14 +25,10 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 import javax.inject.Singleton;
-import javax.ws.rs.Consumes;
-import javax.ws.rs.GET;
-import javax.ws.rs.Path;
-import javax.ws.rs.core.MediaType;
 
 import org.jclouds.azurecompute.arm.AzureComputeApi;
+import org.jclouds.azurecompute.arm.config.GraphRBAC.GraphRBACForTenant;
 import org.jclouds.azurecompute.arm.domain.ServicePrincipal;
-import org.jclouds.azurecompute.arm.filters.ApiVersionFilter;
 import org.jclouds.azurecompute.arm.handlers.AzureComputeErrorHandler;
 import org.jclouds.http.HttpErrorHandler;
 import org.jclouds.http.annotation.ClientError;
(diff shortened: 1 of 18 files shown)


jmspring:
@nacx - this worked well.  

Some thoughts re: ObjectID and KeyVault -- the current tests use the configured service principal and give access to that SP for the generated vault for purposes of the tests.  It is possible to grant access to other service principals as well.  So access to the GraphRBAC API to resolve the ObjectId for a general SP might be useful.

Also, I've added a patch to extracting out the TenantID since the URL can vary.  For instance `https://login.microsoft.com` is another valid OAuth endpoint.

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/config/AzureComputeHttpApiModule.java
```diff
@@ -16,23 +16,56 @@
...
+import com.google.inject.Provides;
 import com.google.inject.Scopes;
+import com.google.inject.name.Named;
 
 @ConfiguresHttpApi
 public class AzureComputeHttpApiModule extends HttpApiModule<AzureComputeApi> {
 
+   private static final Pattern OAUTH_TENANT_PATTERN = Pattern
+         .compile("https://login.microsoftonline.com/([^/]+)/oauth2/token");
+
```
jmspring:
```
   private static final Pattern OAUTH_TENANT_PATTERN_MICROSOFT_ONLINE = Pattern
           .compile("https://login.microsoftonline.com/([^/]+)/oauth2/token");
   private static final Pattern OAUTH_TENANT_PATTERN_MICROSOFT = Pattern
           .compile("https://login.microsoft.com/([^/]+)/oauth2/token");
```

azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/config/AzureComputeHttpApiModule.java
```diff
@@ -46,10 +79,58 @@ protected void installLocations() {
...
+   @Singleton
+   @Tenant
+   protected String provideTenant(@Named("oauth.endpoint") final String oauthEndpoint) {
+      Matcher m = OAUTH_TENANT_PATTERN.matcher(oauthEndpoint);
+      if (!m.matches()) {
+         throw new IllegalArgumentException("Could not parse tenantId from: " + oauthEndpoint);
+      }
+      return m.group(1);
+   }
+
```
jmspring:
```
   protected String provideTenant(@Named("oauth.endpoint") final String oauthEndpoint) {
      Matcher m = OAUTH_TENANT_PATTERN_MICROSOFT_ONLINE.matcher(oauthEndpoint);
      if (!m.matches()) {
         m = OAUTH_TENANT_PATTERN_MICROSOFT.matcher(oauthEndpoint);
         if(!m.matches()) {
            throw new IllegalArgumentException("Could not parse tenantId from: " + oauthEndpoint);
         }
      }
      return m.group(1);
   }
```

nacx:
Thanks for the review @jmspring! According to your comments I've slightly changed the PR to:

* Support both formats of the OAuth endpoint (tests added).
* Extracted the service principal call into a GraphRBACApi class, so we can add additional methods there as needed.
* Modified the structure of the HTTPApiModule and the base mock test class to allow overriding the custom endpoints used in the graph API, and possibly others (as they differ from the endpoint configured when creating the context, and overriding is not trivial).

PR #426, opened 2017-12-08
Implement GetOrCreateNetworkDomainThenCreateNodes Strategy

trevorflanagan:
Extending `org.jclouds.compute.strategy.impl.CreateNodesWithGroupEncodedIntoNameThenAddToSet` for jclouds abstraction strategy.

### DIFF
--- a/dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/compute/options/DimensionDataCloudControlTemplateOptions.java
+++ b/dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/compute/options/DimensionDataCloudControlTemplateOptions.java
@@ -1,0 +1,153 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jclouds.dimensiondata.cloudcontrol.compute.options;
+
+import org.jclouds.compute.options.TemplateOptions;
+import org.jclouds.javax.annotation.Nullable;
+
+public class DimensionDataCloudControlTemplateOptions extends TemplateOptions implements Cloneable {
+
+   public static final String DEFAULT_NETWORK_DOMAIN_NAME = "JCLOUDS_NETWORK_DOMAIN";
+   public static final String DEFAULT_VLAN_NAME = "JCLOUDS_VLAN";
+   public static final String DEFAULT_PRIVATE_IPV4_BASE_ADDRESS = "10.0.0.0";
+   public static final Integer DEFAULT_PRIVATE_IPV4_PREFIX_SIZE = 24;
+
+   private String networkDomainName;
+   private String defaultPrivateIPv4BaseAddress;
+   private Integer defaultPrivateIPv4PrefixSize;
+
+   public String getNetworkDomainName() {
+      return networkDomainName;
+   }
+
+   public String getDefaultPrivateIPv4BaseAddress() {
+      return defaultPrivateIPv4BaseAddress;
+   }
+
+   public Integer getDefaultPrivateIPv4PrefixSize() {
+      return defaultPrivateIPv4PrefixSize;
+   }
+
+   public DimensionDataCloudControlTemplateOptions networkDomainName(@Nullable String networkDomainName) {
+      this.networkDomainName = networkDomainName;
+      return this;
+   }
+
+   public DimensionDataCloudControlTemplateOptions defaultPrivateIPv4BaseAddress(
+         @Nullable String defaultPrivateIPv4BaseAddress) {
+      this.defaultPrivateIPv4BaseAddress = defaultPrivateIPv4BaseAddress;
+      return this;
+   }
+
+   public DimensionDataCloudControlTemplateOptions defaultPrivateIPv4PrefixSize(
+         @Nullable Integer defaultPrivateIPv4PrefixSize) {
+      this.defaultPrivateIPv4PrefixSize = defaultPrivateIPv4PrefixSize;
+      return this;
+   }
+
+   @Override
+   public DimensionDataCloudControlTemplateOptions clone() {
+      final DimensionDataCloudControlTemplateOptions options = new DimensionDataCloudControlTemplateOptions();
+      copyTo(options);
+      return options;
+   }
+
+   @Override
+   public boolean equals(Object o) {
+      if (this == o) {
+         return true;
+      }
+      if (!(o instanceof DimensionDataCloudControlTemplateOptions)) {
+         return false;
+      }
+      if (!super.equals(o)) {
+         return false;
+      }
+
+      DimensionDataCloudControlTemplateOptions that = (DimensionDataCloudControlTemplateOptions) o;
+
+      if (networkDomainName != null ?
+            !networkDomainName.equals(that.networkDomainName) :
+            that.networkDomainName != null) {
+         return false;
+      }
+
+      if (defaultPrivateIPv4BaseAddress != null ?
+            !defaultPrivateIPv4BaseAddress.equals(that.defaultPrivateIPv4BaseAddress) :
+            that.defaultPrivateIPv4BaseAddress != null) {
+         return false;
+      }
+      return defaultPrivateIPv4PrefixSize != null ?
+            defaultPrivateIPv4PrefixSize.equals(that.defaultPrivateIPv4PrefixSize) :
+            that.defaultPrivateIPv4PrefixSize == null;
+
+   }
+
+   @Override
+   public int hashCode() {
+      int result = super.hashCode();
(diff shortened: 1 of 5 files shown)


dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/compute/strategy/GetOrCreateNetworkDomainThenCreateNodes.java
```diff
@@ -0,0 +1,175 @@
...
+@Singleton
+public class GetOrCreateNetworkDomainThenCreateNodes extends CreateNodesWithGroupEncodedIntoNameThenAddToSet {
+
+   private final DimensionDataCloudControlApi api;
+   private final ComputeServiceConstants.Timeouts timeouts;
+
+   public static final String DEFAULT_NETWORK_DOMAIN_NAME = "JCLOUDS_NETWORK_DOMAIN";
+   public static final String DEFAULT_VLAN_NAME = "JCLOUDS_VLAN";
+   public static final String DEFAULT_PRIVATE_IPV4_BASE_ADDRESS = "10.0.0.0";
+   public static final Integer DEFAULT_PRIVATE_IPV4_PREFIX_SIZE = 24;
```
nacx: Better define these constants in the options class.

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/compute/options/DimensionDataCloudControlTemplateOptions.java
```diff
@@ -0,0 +1,217 @@
...
+
+import org.jclouds.compute.options.TemplateOptions;
+import org.jclouds.javax.annotation.Nullable;
+
+public class DimensionDataCloudControlTemplateOptions extends TemplateOptions implements Cloneable {
+
+   private String networkDomainName;
+   private String vlanName;
+   private String networkDomainId;
+   private String vlanId;
```
nacx: It looks like these IDs are not meant to be configured by users and only used to carry information to the adapter implementation, so they should not be exposed in the public API.
Instead of doing this, you could use an intermediate (and internal, not exposed to users) options class to just carry those computed values. We do something like this [in the ProfitBricks](https://github.com/jclouds/jclouds/blob/master/providers/profitbricks/src/main/java/org/jclouds/profitbricks/compute/strategy/AssignDataCenterToTemplate.java#L106) provider.

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/compute/strategy/GetOrCreateNetworkDomainThenCreateNodes.java
```diff
@@ -0,0 +1,175 @@
...
+   @Override
+   public Map<?, ListenableFuture<Void>> execute(final String group, final int count, final Template template,
+         final Set<NodeMetadata> goodNodes, final Map<NodeMetadata, Exception> badNodes,
+         final Multimap<NodeMetadata, CustomizationResponse> customizationResponses) {
+
+      final DimensionDataCloudControlTemplateOptions templateOptions = template.getOptions()
+            .as(DimensionDataCloudControlTemplateOptions.class);
+
+      String networkDomainName = firstNonNull(templateOptions.getNetworkDomainName(), DEFAULT_NETWORK_DOMAIN_NAME);
+      String vlanName = firstNonNull(templateOptions.getVlanName(), DEFAULT_VLAN_NAME);
```
nacx: The jclouds portable template options have the `networks` attribute (names, ids, or whatever makes sense for the provider) so users can configure to which networks the nodes will be connected to. Instead of providing two different options to configure that, we should try to accommodate the current behavior to using that `networks` option in a way that is understandable and makes sense to users.

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/compute/strategy/GetOrCreateNetworkDomainThenCreateNodes.java
```diff
@@ -0,0 +1,175 @@
...
+      logger.debug("Creating a vlan %s in network domain '%s' ...", vlanName, networkDomainId);
+      String defaultPrivateIPv4BaseAddress = firstNonNull(templateOptions.getDefaultPrivateIPv4BaseAddress(),
+            DEFAULT_PRIVATE_IPV4_BASE_ADDRESS);
+      Integer defaultPrivateIPv4PrefixSize = firstNonNull(templateOptions.getDefaultPrivateIPv4PrefixSize(),
+            DEFAULT_PRIVATE_IPV4_PREFIX_SIZE);
+
+      String vlanId = api.getNetworkApi()
+            .deployVlan(networkDomainId, vlanName, "vlan created by jclouds", defaultPrivateIPv4BaseAddress,
+                  defaultPrivateIPv4PrefixSize);
+      boolean isVlanReady = retry(new VlanState(api.getNetworkApi(), State.NORMAL), timeouts.nodeRunning).apply(vlanId);
```
nacx: This is a small PR and I think we should introduce the Guice-configured predicates here and stop initializing them manually. Let's better properly configure now the predicates now that we have the compute service module and get them injected where needed. This can be done in a different PR that unifies them and removes the static utils class.

trevorflanagan:
@nacx thanks for the feedback. I will make the change for the State and Status classes in another pull request before making the requested changes you have suggested.

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/compute/strategy/GetOrCreateNetworkDomainThenCreateNodes.java
```diff
@@ -79,17 +74,21 @@ protected GetOrCreateNetworkDomainThenCreateNodes(final CreateNodeWithGroupEncod
       final DimensionDataCloudControlTemplateOptions templateOptions = template.getOptions()
             .as(DimensionDataCloudControlTemplateOptions.class);
 
-      String networkDomainName = firstNonNull(templateOptions.getNetworkDomainName(), DEFAULT_NETWORK_DOMAIN_NAME);
-      String vlanName = firstNonNull(templateOptions.getVlanName(), DEFAULT_VLAN_NAME);
+      final Set<String> networks = templateOptions.getNetworks();
+
+      String networkDomainName = firstNonNull(!networks.isEmpty() ? networks.iterator().next() : null,
+            DimensionDataCloudControlTemplateOptions.DEFAULT_NETWORK_DOMAIN_NAME);
```
nacx: Is it possible in DD to attach a server to more than one network? (That's why the `networks` property is a list). And in that case What considerations/preconditions/limitations should be taken into account?
If not, instead of silently ignoring all but the first network, it would be better to validate that there is only one and fail otherwise.

trevorflanagan: @nacx the change I have made appears to be incorrect. In DD model a Server can be attached to many Vlan. A Vlan is associated with a single Network Domain only. A Server can only be associated with Vlans on the **same** Network Domain.
I think that recording the lists of VLANs in  `templateOptions.network` might be a better approach. My thoughts are that we revert the change to revert the `networkDomainName` removal from `DimensionDataCloudControlTemplateOptions`
Replace `DimensionDataCloudControlTemplateOptions vlanName` with `templateOptions.network`

I do have a question though about what exactly we store in `templateOptions.network`. We could store the Vlans for a given Network Domain name `org.jclouds.dimensiondata.cloudcontrol.features.NetworkApi#listVlans(java.lang.String)` **OR** we could use a Server to retrieve this information and therefore reduce the set of Vlans we store in `templateOptions.network`. Just not sure what you think is the best approach?

nacx: >I think that recording the lists of VLANs in templateOptions.network might be a better approach. My thoughts are that we revert the change to revert the networkDomainName removal from DimensionDataCloudControlTemplateOptions
Replace DimensionDataCloudControlTemplateOptions vlanName with templateOptions.network

Sounds good. Use the DD options to configure the domain name, and read the VLAN names from the `templateOptions.networks` option.

>OR we could use a Server to retrieve this information and therefore reduce the set of Vlans we store in templateOptions.network. Just not sure what you think is the best approach?

I'm not sure I get what you mean with this. The `templateOptions.networks` options are used to tell jclouds where to attach the servers it is about to create... Could you please clarify what do you mean?

trevorflanagan: >Sounds good. Use the DD options to configure the domain name, and read the VLAN names from the templateOptions.networks option.

Cool @nacx this is the approach I am going to take. 

>> OR we could use a Server to retrieve this information and therefore reduce the set of Vlans we store in templateOptions.network. Just not sure what you think is the best approach?

> I'm not sure I get what you mean with this. The templateOptions.networks options are used to tell jclouds where to attach the servers it is about to create... Could you please clarify what do you mean?

I took a look a bit further at the core jclouds code and that option does not make any sense, since we are actually creating nodes here.

trevorflanagan:
@nacx I have squashed my commits. As agreed earlier on slack I have done with the approach of defaulting to a single vlan for the nodes created. I am also using the approach of storing `vlanName` in `templateOptions.networks(vlanName);`

nacx:
Pushed to [master](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/5a3b5991) and [2.0.x](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/0351481d). I've added [a commit](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/199e7789) in 2.0.x to fix the Guava imports and remove unnecessary calls to `injectMocks`.
Thanks @trevorflanagan!

PR #400, opened 2017-07-07
JCLOUDS-1255 Server API migration.

trevorflanagan:
Adding the Server API - deploy, clone, list, start, stop server.

Live Tests

Mock Tests

PR #427, opened 2017-12-14
Refactor DimensionDataCloudControlResponseUtils waitFor* methods so they are injectable

trevorflanagan:
@nacx as requested I have done this piece of refactoring that was requested in https://github.com/jclouds/jclouds-labs/pull/426 and https://github.com/jclouds/jclouds-labs/pull/400#discussion_r127640050. All of our live tests continue to work with this change.


dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/config/DimensionDataCloudControlComputeServiceContextModule.java
```diff
@@ -0,0 +1,254 @@
...
+import org.jclouds.dimensiondata.cloudcontrol.features.ServerApi;
+import org.jclouds.logging.Logger;
+
+import javax.annotation.Resource;
+import java.text.MessageFormat;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static org.jclouds.util.Predicates2.retry;
+
+public class DimensionDataCloudControlComputeServiceContextModule extends AbstractModule {
```
trevorflanagan: This was in the test package for some reason. I have moved it across to production code. I have also simplified the Module so that it extends `com.google.inject.AbstractModule`. Originally it extended `ComputeServiceAdapterContextModule<ServerWithExternalIp, BaseImage, BaseImage, Datacenter>` but this causes problems in our live tests since not all of our functions are complete yet.

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/config/DimensionDataCloudControlComputeServiceContextModule.java
```diff
@@ -0,0 +1,254 @@
...
+   @Override
+   protected void configure() {
+
+   }
+
+   @Provides
+   @Named(VLAN_DELETED_PREDICATE)
+   protected Predicate<String> provideVlanDeletedPredicate(final DimensionDataCloudControlApi api,
+         final ComputeServiceConstants.Timeouts timeouts, final ComputeServiceConstants.PollPeriod pollPeriod) {
+      return retry(new VlanState(api.getNetworkApi(), State.DELETED), timeouts.imageAvailable,
```
nacx: The `imageAvailable` timeout is meant to be used by the `ImageExtension` when waiting for a new image to be created. Instead of using it, just create a property such as `operation-timeout` that can be configured as a generic timeout for DimensionDataOperations, and get it injected here and in other methods that don't have a matching timeout property in the `ComputeServiceConstants.Timeouts` class.

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/config/DimensionDataCloudControlComputeServiceContextModule.java
```diff
@@ -0,0 +1,254 @@
...
+         final ComputeServiceConstants.Timeouts timeouts, final ComputeServiceConstants.PollPeriod pollPeriod) {
+      return retry(new NetworkDomainState(api.getNetworkApi(), State.NORMAL), timeouts.imageAvailable,
+            pollPeriod.pollInitialPeriod, pollPeriod.pollMaxPeriod);
+   }
+
+   @Provides
+   @Named(SERVER_STARTED_PREDICATE)
+   protected Predicate<String> provideServerStartedPredicate(final DimensionDataCloudControlApi api,
+         final ComputeServiceConstants.Timeouts timeouts, final ComputeServiceConstants.PollPeriod pollPeriod) {
+      return retry(new ServerStatus(api.getServerApi(), true, true), timeouts.imageAvailable,
```
nacx: Use `timeouts.nodeRunning`.

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/config/DimensionDataCloudControlComputeServiceContextModule.java
```diff
@@ -0,0 +1,254 @@
...
+            pollPeriod.pollInitialPeriod, pollPeriod.pollMaxPeriod);
+   }
+
+   @Provides
+   @Named(SERVER_STOPPED_PREDICATE)
+   @VisibleForTesting
+   public Predicate<String> provideServerStoppedPredicate(final DimensionDataCloudControlApi api,
+         final ComputeServiceConstants.Timeouts timeouts, final ComputeServiceConstants.PollPeriod pollPeriod) {
+      return retry(new ServerStatus(api.getServerApi(), false, true), timeouts.imageAvailable,
```
nacx: `timeouts.nodeSuspended`

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/config/DimensionDataCloudControlComputeServiceContextModule.java
```diff
@@ -0,0 +1,254 @@
...
+            pollPeriod.pollInitialPeriod, pollPeriod.pollMaxPeriod);
+   }
+
+   @Provides
+   @Named(SERVER_DELETED_PREDICATE)
+   @VisibleForTesting
+   public Predicate<String> provideServerDeletedPredicate(final DimensionDataCloudControlApi api,
+         final ComputeServiceConstants.Timeouts timeouts, final ComputeServiceConstants.PollPeriod pollPeriod) {
+      return retry(new ServerState(api.getServerApi(), State.DELETED), timeouts.imageAvailable,
```
nacx: `timeouts.nodeTerminated`

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/config/DimensionDataCloudControlComputeServiceContextModule.java
```diff
@@ -0,0 +1,254 @@
...
+            pollPeriod.pollInitialPeriod, pollPeriod.pollMaxPeriod);
+   }
+
+   @Provides
+   @Named(VM_TOOLS_RUNNING_PREDICATE)
+   protected Predicate<String> provideVMToolsRunningPredicate(final DimensionDataCloudControlApi api,
+         final ComputeServiceConstants.Timeouts timeouts, final ComputeServiceConstants.PollPeriod pollPeriod) {
+      return retry(new VMToolsRunningStatus(api.getServerApi()), timeouts.imageAvailable, pollPeriod.pollInitialPeriod,
+            pollPeriod.pollMaxPeriod);
+   }
```
nacx: In general, use the existing timeouts if they make sense, or use a Dimensiondata specific timeout. The timeouts configured in the `ComputeServiceConstants.Timeouts` class are automatically used by jclouds when waiting for a node being created. For example, perhaps there is no need to use the "nodeRunning" timeout for a server creation, as that is the timeout jclouds uses to wait for the entire process since the user calls "createNode" until it finishes creating it, which involves several additional steps.
Configure timeout values at your own criteria and provide the corresponding properties, with reasonable defaults, to let users configure them as needed.

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/config/DimensionDataCloudControlComputeServiceContextModule.java
```diff
@@ -0,0 +1,254 @@
...
+
+   private class VlanState implements Predicate<String>
+
```
nacx: [style minor] Remove this line.

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/config/DimensionDataCloudControlComputeServiceContextModule.java
```diff
@@ -0,0 +1,254 @@
...
+      @Override
+      public boolean apply(final String serverId) {
+         checkNotNull(serverId, "serverId");
+         logger.trace("looking for state on Server %s", serverId);
+         final Server server = api.getServer(serverId);
+
+         // perhaps request isn't available, yet
+         if (server == null)
+            return false;
+         logger.trace("%s: looking for server %s deployed: currently: %s", server, server.state());
```
nacx: Missing one parameter for the log message.

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/config/DimensionDataCloudControlComputeServiceContextModule.java
```diff
@@ -0,0 +1,254 @@
...
+
+      @Override
+      public boolean apply(final String networkDomainId) {
+         checkNotNull(networkDomainId, "networkDomainId");
+         logger.trace("looking for state on network domain %s", networkDomainId);
+         final NetworkDomain networkDomain = networkApi.getNetworkDomain(networkDomainId);
+         final boolean isDeleted = networkDomain == null && state == State.DELETED;
+         return isDeleted || (networkDomain != null && networkDomain.state() == state);
+      }
+   }
```
nacx: If this predicate pattern is meant to be used in more classes, consider extracting it to a base class.You could create an interface `WithStatus` (or whatever) that provides the `state()` method, and that would allow you to write a generic predicate that checks the status with just an abstract method to get the `WithStatus` object to be checked. Just an idea to avoid duplicating code if the pattern is going to be the same for all objects with state.

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/config/DimensionDataCloudControlComputeServiceContextModule.java
```diff
@@ -0,0 +1,254 @@
...
+         checkNotNull(serverId, "serverId");
+         logger.trace("looking for state on Server %s", serverId);
+         final Server server = api.getServer(serverId);
+
+         // perhaps request isn't available, yet
+         if (server == null)
+            return false;
+         logger.trace("%s: looking for server %s deployed: currently: %s", server, server.state());
+         if (server.state().isFailed()) {
+            throw new IllegalStateException(MessageFormat.format("Server {0} is in FAILED state", server.id()));
```
nacx: Better `String.format`?

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/config/DimensionDataCloudControlComputeServiceContextModule.java
```diff
@@ -0,0 +1,254 @@
...
+      private final ServerApi api;
+
+      private VMToolsRunningStatus(final ServerApi api) {
+         this.api = api;
+      }
+
+      @Override
+      public boolean apply(final String serverId) {
+         checkNotNull(serverId, "serverId");
+         logger.trace("looking for state on Server %s", serverId);
```
nacx: "Looking for guest tools state"?

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/config/DimensionDataCloudControlComputeServiceContextModule.java
trevorflanagan: Makes sense, for now we should not require more of these Predicates. But I might take this on during my next set of refactoring.

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/config/DimensionDataCloudControlComputeServiceContextModule.java
```diff
@@ -61,49 +61,49 @@ protected void configure() {
...
+            pollPeriod.pollMaxPeriod);
    }
 
    @Provides
    @Named(SERVER_STOPPED_PREDICATE)
    @VisibleForTesting
    public Predicate<String> provideServerStoppedPredicate(final DimensionDataCloudControlApi api,
          final ComputeServiceConstants.Timeouts timeouts, final ComputeServiceConstants.PollPeriod pollPeriod) {
-      return retry(new ServerStatus(api.getServerApi(), false, true), timeouts.imageAvailable,
+      return retry(new ServerStatus(api.getServerApi(), false, true), timeouts.nodeTerminated,
```
nacx: nodeSuspended?

trevorflanagan: @nacx good spot, thanks.

nacx:
Squashed and pushed tp [master](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/889b34b9) and [2.0.x](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/bf34887e). Thanks @trevorflanagan!

PR #428, opened 2017-12-21
Implement ServerWithNatRuleToNodeMetadata Function

btrishkin:


dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/compute/functions/ServerWithNatRuleToNodeMetadata.java
```diff
@@ -0,0 +1,123 @@
...
+      this.api = checkNotNull(api, "api cannot be null");
+   }
+
+   @Override
+   public NodeMetadata apply(final ServerWithExternalIp serverWithExternalIp) {
+      NodeMetadataBuilder builder = new NodeMetadataBuilder();
+      Server server = serverWithExternalIp.server();
+      builder.ids(server.id());
+      builder.name(server.name());
+      builder.hostname(serverWithExternalIp.server().description());
```
nacx: Is this really the hostname? Don't put that here if not.

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/compute/functions/ServerWithNatRuleToNodeMetadata.java
```diff
@@ -0,0 +1,123 @@
...
+      if (server.datacenterId() != null) {
```
nacx: Under which circumstances can this be null?

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/compute/functions/ServerWithNatRuleToNodeMetadata.java
```diff
@@ -0,0 +1,123 @@
...
+
+      String privateAddress = null;
+      if (server.networkInfo() != null && server.networkInfo().primaryNic() != null
+            && server.networkInfo().primaryNic().privateIpv4() != null) {
+         privateAddress = server.networkInfo().primaryNic().privateIpv4();
+         builder.privateAddresses(ImmutableSet.of(privateAddress));
+      }
+      if (privateAddress != null && serverWithExternalIp.externalIp() != null) {
+         builder.publicAddresses(ImmutableSet.of(serverWithExternalIp.externalIp()));
+      }
```
nacx: We should consider not only the primary NIC here. Let's return all IP addresses of the server.

nacx: BTW, why read the external IP only if there is a private IP?

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/compute/functions/ServerWithNatRuleToNodeMetadata.java
```diff
@@ -0,0 +1,123 @@
...
+
+      // DimensionData does not provide a way to get the credentials.
+      // Try to return them from the credential store
+      Credentials credentials = credentialStore.get("node#" + server.id());
+      if (credentials instanceof LoginCredentials) {
+         builder.credentials(LoginCredentials.class.cast(credentials));
+      }
```
nacx: Remove this. The credentials are [already set by jclouds](https://github.com/jclouds/jclouds/blob/master/compute/src/main/java/org/jclouds/compute/strategy/impl/AdaptingComputeServiceStrategies.java#L89-L90).

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/compute/functions/ServerWithNatRuleToNodeMetadata.java
```diff
@@ -0,0 +1,123 @@
...
+
+      return builder.build();
+   }
+
+   BaseImage getBaseImage(final String sourceImageId) {
+      BaseImage baseImage = api.getServerImageApi().getOsImage(sourceImageId);
+      return baseImage == null ? api.getServerImageApi().getCustomerImage(sourceImageId) : baseImage;
+   }
```
nacx: This means two additional calls for each existing node, to get the image info. Instead of doing this, could you inject the image supplier to this class?
jclouds caches the results of the `ComptueService.listImages` call (as it is usually an expensive one). You could look for the image in the list of cached images and save these two calls for each node (that would be a significant amount of extra calls when calling `listNodes` in large environments).

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/compute/functions/ServerWithNatRuleToNodeMetadata.java
trevorflanagan: Should not be null. Our xsd has it as a required attribute:
`<xs:attribute name="datacenterId" type="xs:string" use="required"/>`

btrishkin:
@nacx small issue fixed and commits squashed.

nacx:
Pushed to [master](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/efb2a3de) and [2.1.x](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/87ff3afe). Thanks @btrishkin!

PR #423, opened 2017-11-30
DatacenterToLocation / ServerToServerWithExternalIp Functions implementation for dimensiondata

btrishkin:

PR #429, opened 2017-12-22
Configure DimensionData locations

nacx:
This configures the locations for DimensionData. Users won't have to configure the `endpoint` when creating the context, as jclouds will already know all region endpoints.

The configured suppliers preconfigure all the region endpoints, and for each one, it retrieves the list of available zones. jclouds will preconfigure all regions based on that information.

If users only want to use a subset of the existing regions or just have access to several regions, they will just to override the `jclouds.regions` property to the comma-separated list of region names they want to use.

With this change, the `DatacenterToLocation` function from https://github.com/jclouds/jclouds-labs/pull/423 is not needed, as the compute service won't be used to retrieve the locations.

/cc @trevorflanagan


nacx:
I'm still working on this @trevorflanagan. I'll ping you when it is ready to test.

PR #431, opened 2018-02-14
 JCLOUDS-1386  1&1 Baremetal servers

alibazlamit:
This PR holds the latest updates in the 1&1 API


alibazlamit:
Compute test results :
https://gist.github.com/alibazlamit/fab05a5e3235be8202f34c627d8bdc65

andreaturli:
thanks @alibazlamit could you please open a jira issue that describes the rationale for this PR?
Thanks

alibazlamit:
@andreaturli one main feature that this PR is introducing is the Bare-metal Servers, now i have implemented that as a feature but it is not integrated with the compute API. Basically there is a new property introduced `server_type` which defaults to `cloud` or could be `baremetal`, do you have any idea/examples where you can choose the server type when building a template, i know that with templates you can choose which images to use and what hardware goes, but how can we introduce choosing a `baremetal` or `cloud` server type?

Thanks

andreaturli:
Thanks @alibazlamit for the summary.

Would you mind to copy that into a jira issue as described at point 2 at https://cwiki.apache.org/confluence/display/JCLOUDS/How+to+Contribute
Ideally you can rename the PR using the jira issue I'd.

That would hugely help us to manage the project.Thanks

alibazlamit:
I am sorry for the confusion, but there was a question in my previous comment, something that i would like to add to the PR before we merge it. Could you please take a look at it and let me know if you can maybe point me towards a solution.

Thanks

andreaturli:
Not familiar with the oneandone API but looks like `baremetal` is another servertype for server creation.

If this is correct, then we could either introduce a OneAndOneTemplateOption for serverType or try to infer it from hardwareProfiles if this additional flag is available. Makes sense?

alibazlamit:
I have implemented the baremetal creation in the compute service adapter, based on the hardware flavor chosen we can determine the server type.

Here is the [JIRA ](https://issues.apache.org/jira/browse/JCLOUDS-1386)ticket.

alibazlamit:
@andreaturli a friendly reminder, this is a review ready PR

andreaturli:
Thanks @alibazlamit, we'll have asap. Thanks

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/compute/OneandoneComputeServiceAdapter.java
```diff
@@ -65,6 +66,7 @@
    private final OneAndOneApi api;
    private final Predicate<Server> waitServerUntilAvailable;
    private final PasswordGenerator.Config passwordGenerator;
```
andreaturli: it seems not used

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/compute/OneandoneComputeServiceAdapter.java
```diff
@@ -83,56 +85,112 @@
...
-               }
-            }
-            Hdd.CreateHdd hdd = Hdd.CreateHdd.create(minHddSize, volume.isBootDevice());
-            hdds.add(hdd);
-         } catch (Exception ex) {
-            throw Throwables.propagate(ex);
-
+      String imageId = image.getId();
+      HardwareFlavour hardwareModel = isFlavor(hardware.getId());
+      boolean isBaremetal = hardware.getName() == null ? false : hardware.getName().contains(baremetalModelsKey);
```
andreaturli: I think worth introducing a `OneandoneTemplateOptions extends TemplateOptions` to deal with specific requirements a user may specify like bare metal or even `hardwareModel` above. wdyt?

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/compute/OneandoneComputeServiceAdapter.java
```diff
@@ -83,56 +85,112 @@
...
       }
 
+      //
+      // provision server
+      Server server = null;
+      org.apache.jclouds.oneandone.rest.domain.Hardware.CreateHardware hardwareRequest = null;
       try {
-         org.apache.jclouds.oneandone.rest.domain.Hardware.CreateHardware hardwareRequest
-                 = org.apache.jclouds.oneandone.rest.domain.Hardware.CreateHardware.create(cores, 1, ram, hdds);
+         if (hardwareModel != null) {
```
andreaturli: this logic to build the hardwareRequest is quite complex. would it make sense to extract it in its own method, and add some unit tests, ideally

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/compute/OneandoneComputeServiceAdapter.java
```diff
@@ -179,9 +220,54 @@
...
    @Override
    public List<HardwareFlavour> listHardwareProfiles() {
-      return api.serverApi().listHardwareFlavours();
+      List<HardwareFlavour> result = new ArrayList<HardwareFlavour>();
+      List<HardwareFlavour> cloudModels = api.serverApi().listHardwareFlavours();
+      for (HardwareFlavour flavor : cloudModels) {
+         result.add(flavor);
+      }
+      List<BareMetalModel> baremetalModels = api.serverApi().listBaremetalModels();
+      for (BareMetalModel model : baremetalModels) {
```
andreaturli: maybe you can consider a `Function<BareMetalModel, HardwareFlavour>` for better readability and testability?

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/compute/config/OneAndOneComputeServiceContextModule.java
```diff
@@ -125,4 +126,28 @@ public boolean apply(Server server) {
                  || server.status().percent() != 0);
       }
    }
+
+   @Provides
+   @Singleton
+   @Named(POLL_PREDICATE_SERVER_SUSPENDED)
```
andreaturli: why a not using `ComputeServiceProperties.TIMEOUT_NODE_SUSPENDED` so that you can use a quite adopted pattern like in openstack,digitalocean,packet
```
   @Provides
   @com.google.inject.name.Named(TIMEOUT_NODE_TERMINATED)
   protected Predicate<RegionAndId> provideServerTerminatedPredicate(final NovaApi api, ComputeServiceConstants.Timeouts timeouts,
                                                                ComputeServiceConstants.PollPeriod pollPeriod) {
      return retry(new ServerTerminatedPredicate(api), timeouts.nodeTerminated, pollPeriod.pollInitialPeriod,
              pollPeriod.pollMaxPeriod);
   }
```
?

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/compute/function/HardwareFlavourToHardware.java
```diff
@@ -53,7 +53,7 @@ public Hardware apply(HardwareFlavour from) {
       }
       final HardwareBuilder builder;
       builder = new HardwareBuilder()
-              .ids(from.id())
+              .ids(from.name())
```
andreaturli: maybe better to use 
```
.id(from.id())
.name(from.name())
```
if `.ids()` is not what you want here

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/config/OneAndOneProperties.java
```diff
@@ -19,6 +19,7 @@
 public final class OneAndOneProperties {
 
    public static final String POLL_PREDICATE_SERVER = "jclouds.oneandone.rest.predicate.server";
+   public static final String POLL_PREDICATE_SERVER_SUSPENDED = "jclouds.oneandone.rest.predicate.serverdelete";
```
andreaturli: oh now I see, it's a predicate to wait for server deletion not really suspension. ok maybe my comment above is still valid for other predicates though

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/domain/BareMetalModel.java
```diff
@@ -0,0 +1,70 @@
...
+
+      public abstract double core();
+
+      public abstract double coresPerProcessor();
+
+      public abstract double ram();
+
+      public abstract String unit();
+
+      public abstract Hdd hdds();
```
andreaturli: out of curiosity, why it's `hdds` if it is not a collection?

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/compute/OneandoneComputeServiceAdapter.java
alibazlamit: I need to combine `BareMetalModel` with `HardwareFlavour` as one list, to mix the baremetal flavors with the cloud flavors, do you think that could be achieved using the function you suggested?

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/compute/function/HardwareFlavourToHardware.java
alibazlamit: My first approach was what you have suggested, for some reason the API has similiar Id's between the baremetal flavors and the cloud flavors, the only thing to make them unique is the name so far, this might be an issue on the API side that might get changed in the future, if that ever happens we can go with what you suggested

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/compute/OneandoneComputeServiceAdapter.java
alibazlamit: My intention was to allow the users to just pass in the flavor name i.e. `BMC_XL` or `XL` as a `hardwareId` to choose between baremetal and cloud servers.

Maybe the `OneandoneTemplateOptions` would be an overkill since the only thing users will need to provide is the different hardware flavor name which can be achieved using the `public TemplateBuilder hardwareId(String string);` from the `TeamplateBuilder` class

alibazlamit:
@andreaturli Thanks for the review, i have applied most of the changes you requested.
I wanted to run the live tests, but four of them fail on when trying to ssh into the server, for example `testCreateAndRunAService` fails with this error 
```
(web:rsa[fingerprint(37:18:5e:f4:ce:d4:95:c4:02:10:8c:01:b9:b9:d3:4b),sha1(c0:81:47:7b:a1:88:47:2d:e0:1c:e4:6a:28:d6:fc:ab:c3:6b:44:16)]@50.21.186.61:22) (web:rsa[fingerprint(37:18:5e:f4:ce:d4:95:c4:02:10:8c:01:b9:b9:d3:4b),sha1(c0:81:47:7b:a1:88:47:2d:e0:1c:e4:6a:28:d6:fc:ab:c3:6b:44:16)]@50.21.186.61:22) error acquiring {hostAndPort=50.21.186.61:22, loginUser=web, ssh=null, connectTimeout=300000, sessionTimeout=300000} (out of retries - max 7): Exhausted available authentication methods
org.jclouds.rest.AuthorizationException
	at org.jclouds.sshj.SshjSshClient.propagate(SshjSshClient.java:394)
```

I am not sure why is it trying to ssh into the server with loginUser=web, there is not such user at the created server and it is not going to be able to ssh, any ideas?

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/compute/OneandoneComputeServiceAdapter.java
```diff
@@ -62,77 +61,95 @@
...
+      this.generateHardwareRequest = generateHardwareRequest;
       this.waitServerUntilAvailable = waitServerUntilAvailable;
-      this.passwordGenerator = passwordGenerator;
    }
 
    @Override
    public NodeAndInitialCredentials<Server> createNodeWithGroupEncodedIntoName(String group, String name, Template template) {
 
-      final String dataCenterId = template.getLocation().getId();
+      final String dataCenterId ="908DC2072407C94C8054610AD5A53B8C";// template.getLocation().getId();
```
andreaturli: why this hardcoded value?

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/compute/OneandoneComputeServiceAdapter.java
```diff
@@ -62,77 +61,95 @@
...
    private final OneAndOneApi api;
    private final Predicate<Server> waitServerUntilAvailable;
-   private final PasswordGenerator.Config passwordGenerator;
+   private final String baremetalModelsKey = "BMC";
 
    @Inject
    OneandoneComputeServiceAdapter(OneAndOneApi api, CleanupResources cleanupResources,
+           GenerateHardwareRequest generateHardwareRequest,
            @Named(POLL_PREDICATE_SERVER) Predicate<Server> waitServerUntilAvailable,
            PasswordGenerator.Config passwordGenerator) {
```
andreaturli: as you have removed it from the constructor, why do you still need to inject it?

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/compute/OneandoneComputeServiceAdapter.java
```diff
@@ -62,77 +61,95 @@
...
    }
 
    @Override
    public NodeAndInitialCredentials<Server> createNodeWithGroupEncodedIntoName(String group, String name, Template template) {
 
-      final String dataCenterId = template.getLocation().getId();
+      final String dataCenterId ="908DC2072407C94C8054610AD5A53B8C";// template.getLocation().getId();
       Hardware hardware = template.getHardware();
       TemplateOptions options = template.getOptions();
       Server updateServer = null;
```
andreaturli: null not needed, maybe remove it?

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/compute/OneandoneComputeServiceAdapter.java
```diff
@@ -62,77 +61,95 @@
...
-            }
-            Hdd.CreateHdd hdd = Hdd.CreateHdd.create(minHddSize, volume.isBootDevice());
-            hdds.add(hdd);
-         } catch (Exception ex) {
-            throw Throwables.propagate(ex);
-
+      String imageId = image.getId();
+      Hardware hardwareModel = generateHardwareRequest.isFlavor(hardware.getId());
+      boolean isBaremetal = hardware.getName() == null ? false : hardware.getName().contains(baremetalModelsKey);
+      ServerAppliance workingImage = null;
```
andreaturli: null not needed, maybe remove it?

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/compute/OneandoneComputeServiceAdapter.java
```diff
@@ -62,77 +61,95 @@
...
+         FirewallPolicy.Rule.CreatePayload rule = FirewallPolicy.Rule.CreatePayload.builder()
+                 .portFrom(range.getKey())
+                 .portTo(range.getValue())
+                 .protocol(Types.RuleProtocol.TCP)
+                 .build();
+         rules.add(rule);
+      }
+      String firewallPolicyId = "";
+      if (inboundPorts.length > 0) {
+         FirewallPolicy firewallPolicy = api.firewallPolicyApi().create(FirewallPolicy.CreateFirewallPolicy.create(name + " firewall policy", "desc", rules));
```
andreaturli: what happens when you ask to create multiple nodes? does `firewallPolicyApi` handles the concurrency correctly?

jclouds usually takes care of fw rules for a group of node extending `CreateNodesWithGroupEncodedIntoNameThenAddToSet` -- see `ApplyNovaTemplateOptionsCreateNodesWithGroupEncodedIntoNameThenAddToSet` in openstack-nova api for an example

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/compute/OneandoneComputeServiceAdapter.java
```diff
@@ -62,77 +61,95 @@
...
-            hdds.add(hdd);
-         } catch (Exception ex) {
-            throw Throwables.propagate(ex);
-
+      String imageId = image.getId();
+      Hardware hardwareModel = generateHardwareRequest.isFlavor(hardware.getId());
+      boolean isBaremetal = hardware.getName() == null ? false : hardware.getName().contains(baremetalModelsKey);
+      ServerAppliance workingImage = null;
+
+      //choose the correct image based on the server type baremetal or cloud
```
andreaturli: sorry can't understand this.

Can you elaborate a bit more the rationale of this `findWorkingImage` method?

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/compute/OneandoneComputeServiceAdapter.java
```diff
@@ -62,77 +61,95 @@
...
-      // provision server
-      Server server = null;
-      Double cores = ComputeServiceUtils.getCores(hardware);
-      Double ram = (double) hardware.getRam();
-      if (ram < 1024) {
-         ram = 0.5;
-      } else {
-         ram = ram / 1024;
+      //configuring Firewall rules
+      Map<Integer, Integer> portsRange = getPortRangesFromList(inboundPorts);
```
andreaturli: Is this change strictly related to bare metal support? seems to me that this PR is trying to change too many things, would it be possible for you to split up into multiple smaller PR so to have a better/easier review?

nacx:
>I am not sure why is it trying to ssh into the server with loginUser=web, there is not such user at the created server and it is not going to be able to ssh, any ideas?

There is a test that installs a Jetty Server and creates an admin ("web") user for that service to test end-to-end a typical use case. It is using that "web" user to try access your node. If the login fails, it may be due to the previous script failing or because the user could not be properly configured in the OS. You could check that.
Anyway, let's first address Andrea's comments and once everything is clean we can investigate this failure.

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/compute/OneandoneComputeServiceAdapter.java
alibazlamit: Some of the images are of Type ISO and type Image, the ones that are of type ISO are basically an installation of the Linux system and not needed here, this function makes sure to choose the correct boot-able image based on the image from the template at hand

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/compute/OneandoneComputeServiceAdapter.java
alibazlamit: Mistake, fixed.

alibazlamit: Mistake, fixed.

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/compute/OneandoneComputeServiceAdapter.java
alibazlamit: It is still in the constructor as i need to use it.

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/compute/OneandoneComputeServiceAdapter.java
alibazlamit: removed

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/compute/OneandoneComputeServiceAdapter.java
alibazlamit: removed

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/compute/OneandoneComputeServiceAdapter.java
```diff
@@ -143,23 +160,6 @@
 
          updateServer = api.serverApi().get(server.id());
 
-         Map<Integer, Integer> portsRange = getPortRangesFromList(inboundPorts);
```
alibazlamit: This bulk of code was moved up

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/compute/OneandoneComputeServiceAdapter.java
alibazlamit: It does handle the creation correctly as all the live tests passes, if you think there is a necessary change here, we can do that in another PR that addresses improvements.

This bulk of code was moved up just to organize the process more. it has been merged before since the first PR

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/compute/OneandoneComputeServiceAdapter.java
alibazlamit: This has just been moved to the GenerateHardwareRequest.java to organize code just as you request to separate the creation of the hardware request.

alibazlamit:
> andreaturli commented on Feb 14  
Not familiar with the oneandone API but looks like baremetal is another servertype for server creation.
> 
> If this is correct, then we could either introduce a OneAndOneTemplateOption for serverType or try to infer it from hardwareProfiles if this additional flag is available. Makes sense?

@andreaturli I am looking into that option just in case, could you please point me to an example where one of the current providers actually use and TemplateOption, i would like to investigate that and see if it would fit the case here.

alibazlamit:
@andreaturli @nacx just a friendly reminder, i have addresses all of the comments either by applying the appropriate changes or be replying to them here, could we please move on with this PR?

alibazlamit:
@andreaturli Can you please take a look at this PR and let me know when can we move on with it?

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/compute/OneandoneComputeServiceAdapter.java
alibazlamit: done [here](https://github.com/StackPointCloud/jclouds-labs/blob/oneandone-baremetal-update/oneandone/src/main/java/org/apache/jclouds/oneandone/rest/compute/strategy/GenerateHardwareRequest.java)

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/config/OneAndOneProperties.java
alibazlamit: I don't quite get what is required here could you please explain more?

alibazlamit:
@andreaturli Could you please let me know which issues are not addressed, i replied to every you comment you had, some changes are applied the others are explained as you can see.

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/compute/config/OneAndOneComputeServiceContextModule.java
andreaturli: @alibazlamit why do we need this `POLL_PREDICATE_SERVER_SUSPENDED` when jclouds has `TIMEOUT_NODE_SUSPENDED` ?

andreaturli:
@alibazlamit I've added a question to help the discussion. Can we start from there?

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/compute/config/OneAndOneComputeServiceContextModule.java
alibazlamit: I have changed it to `TIMEOUT_NODE_SUSPENDED`

alibazlamit:
@andreaturli  I have addressed the issues you mentioned, could you now please take a look?

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/compute/OneandoneComputeServiceAdapter.java
andreaturli: `PasswordGenerator.Config passwordGenerator` is not used, remove it

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/domain/Types.java
```diff
@@ -253,7 +253,7 @@ public static ApplianceType fromValue(String v) {
    }
 
    public enum OSImageType {
-      Standard, Minimal, Personal, ISO_OS, ISO_TOOL, NULL, UNRECOGNIZED;
+     STANDARD, MINIMAL, Personal, ISO_OS, ISO_TOOL, NULL, UNRECOGNIZED;
```
andreaturli: please make it all capital

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/domain/Types.java
```diff
@@ -283,4 +283,22 @@ public String toString() {
...
+      cloud, baremetal, UNRECOGNIZED;
+
+      public static ServerType fromValue(String v) {
+         return Enums.getIfPresent(ServerType.class, v).or(UNRECOGNIZED);
+      }
+
+   }
+
+   public enum ServerTypeCompatibility {
+      vps, cloud, baremetal, UNRECOGNIZED;
```
andreaturli: upper case?

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/domain/Types.java
```diff
@@ -283,4 +283,22 @@ public String toString() {
          return value;
       }
    }
+
+   public enum ServerType {
+      cloud, baremetal, UNRECOGNIZED;
```
andreaturli: why not upper case?

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/compute/OneandoneComputeServiceAdapter.java
```diff
@@ -179,9 +179,46 @@
       return new NodeAndInitialCredentials<Server>(updateServer, updateServer.id(), serverCredentials);
    }
 
+   private ServerAppliance findWorkingImage(String imageId, Types.ServerTypeCompatibility serverType) {
```
andreaturli: why this name? it seems that it tries to find a `ServerAppliance`, doesn't it?

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/compute/OneandoneComputeServiceAdapter.java
```diff
@@ -82,57 +83,73 @@
...
-               }
-            }
-            Hdd.CreateHdd hdd = Hdd.CreateHdd.create(minHddSize, volume.isBootDevice());
-            hdds.add(hdd);
-         } catch (Exception ex) {
-            throw Throwables.propagate(ex);
-
+      String imageId = image.getId();
+      Hardware hardwareModel = generateHardwareRequest.isFlavor(hardware.getId());
+      boolean isBaremetal = hardware.getName() == null ? false : hardware.getName().contains(baremetalModelsKey);
```
andreaturli: wouldn't be better to introduce a custom `TemplateOptions` where you can pass in `bareMetal` boolean parameter? see https://github.com/jclouds/jclouds/blob/master/providers/digitalocean2/src/main/java/org/jclouds/digitalocean2/compute/options/DigitalOcean2TemplateOptions.java#L38 for an example

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/compute/OneandoneComputeServiceAdapter.java
```diff
@@ -82,57 +83,73 @@
...
+      ServerAppliance workingImage;
+
+      //choose the correct image based on the server type baremetal or cloud
```
andreaturli: move this business logic into its own method

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/compute/OneandoneComputeServiceAdapter.java
```diff
@@ -82,57 +83,73 @@
...
-      // provision server
-      Server server = null;
-      Double cores = ComputeServiceUtils.getCores(hardware);
-      Double ram = (double) hardware.getRam();
-      if (ram < 1024) {
-         ram = 0.5;
-      } else {
-         ram = ram / 1024;
+      //configuring Firewall rules
+      Map<Integer, Integer> portsRange = getPortRangesFromList(inboundPorts);
```
andreaturli: all the reasonings for the firewalls / inbound ports shouldn't probably happen in the adapter: firewalls are created once and attached to all the nodes of a group. See `CreateResourcesThenCreateNodes` as a reference at https://github.com/jclouds/jclouds/blob/master/providers/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/strategy/CreateResourcesThenCreateNodes.java#L138

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/compute/OneandoneComputeServiceAdapter.java
andreaturli: I've now realised I already commented this section: I'm afraid you need to change the code now otherwise we can't merge this PR. The reason for that is to have homogenous code, easier to maintain and debug.
if each contributions start deviating from jclouds's best practices it would become impossible for maintainers to help out. I'm sure you understand

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/compute/OneandoneComputeServiceAdapter.java
```diff
@@ -259,19 +289,25 @@ public void destroyNode(String id) {
...
+      waitServerUntilAvailable.apply(getNode(id));
+      api.serverApi().updateStatus(id, Server.UpdateStatus.create(Types.ServerAction.REBOOT, Types.ServerActionMethod.HARDWARE, false, null));
       waitServerUntilAvailable.apply(getNode(id));
-      api.serverApi().updateStatus(id, Server.UpdateStatus.create(Types.ServerAction.REBOOT, Types.ServerActionMethod.HARDWARE));
    }
 
    @Override
    public void resumeNode(String id) {
-      api.serverApi().updateStatus(id, Server.UpdateStatus.create(Types.ServerAction.POWER_ON, Types.ServerActionMethod.HARDWARE));
+      api.serverApi().updateStatus(id, Server.UpdateStatus.create(Types.ServerAction.POWER_ON, Types.ServerActionMethod.HARDWARE, false, null));
```
andreaturli: what happens if the node is already started?

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/compute/OneandoneComputeServiceAdapter.java
```diff
@@ -259,19 +289,25 @@ public void destroyNode(String id) {
...
+      waitServerUntilAvailable.apply(getNode(id));
    }
 
    @Override
    public void suspendNode(String id) {
       waitServerUntilAvailable.apply(getNode(id));
-      api.serverApi().updateStatus(id, Server.UpdateStatus.create(Types.ServerAction.POWER_OFF, Types.ServerActionMethod.HARDWARE));
+      api.serverApi().updateStatus(id, Server.UpdateStatus.create(Types.ServerAction.POWER_OFF, Types.ServerActionMethod.HARDWARE, false, null));
```
andreaturli: what if the node is already powered off?

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/compute/function/ServerToNodeMetadata.java
```diff
@@ -78,26 +81,43 @@ public ServerToNodeMetadata(Function<Hdd, Volume> fnVolume,
    @Override
    public NodeMetadata apply(final Server server) {
       checkNotNull(server, "Null server");
+      String hardwareId = null;
 
       DataCenter dataCenter = api.dataCenterApi().get(server.datacenter().id());
       Location location = find(locations.get(), idEquals(dataCenter.id()));
```
andreaturli: you already have the `datacenter.id` you don't need to call the api to get the datacenter object

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/compute/function/ServerToNodeMetadata.java
```diff
@@ -78,26 +81,43 @@ public ServerToNodeMetadata(Function<Hdd, Volume> fnVolume,
...
+      Float size = null;
+      Integer minRam = 0;
+      List<Hdd> hdds = null;
+      List<Volume> volumes = Lists.newArrayList();
+      List<Processor> processors = new ArrayList<Processor>();
+
       //check if the server was built on a hardware flavour(Fixed instance)
-      if (server.hardware().fixedInstanceSizeId() != null && !"0".equals(server.hardware().fixedInstanceSizeId())) {
-         hardware = hardwareFlavors.get().get(server.hardware().fixedInstanceSizeId());
+      if (server.serverType() == Types.ServerType.baremetal) {
```
andreaturli: reuse `HardwareFlavourToHardware` to get the same information, but without calling the `api` again

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/compute/function/ServerToNodeMetadata.java
```diff
@@ -78,26 +81,43 @@ public ServerToNodeMetadata(Function<Hdd, Volume> fnVolume,
...
+         hdds = server.hardware().hdds();
 
          if (server.hardware().hdds().isEmpty()) {
             hdds = api.serverApi().getHardware(server.id()).hdds();
          }
 
-         size = getHddSize(hdds);
+         size = (float) getHddSize(hdds);
+         if (size == 0) {
+            Server refreshedServer = api.serverApi().get(server.id());
```
andreaturli: you should avoid the extra `api` calls inside `ServerToNodeMetadata` - please remove `api` injection and refactor this class to not use it

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/compute/strategy/GenerateHardwareRequest.java
```diff
@@ -0,0 +1,103 @@
...
+import org.apache.jclouds.oneandone.rest.OneAndOneApi;
+import org.apache.jclouds.oneandone.rest.domain.Hdd;
+import org.apache.jclouds.oneandone.rest.domain.SingleServerAppliance;
+import org.jclouds.compute.domain.Hardware;
+import org.jclouds.compute.domain.Template;
+import org.jclouds.compute.domain.Volume;
+import org.jclouds.compute.util.ComputeServiceUtils;
+
+@Singleton
+public class GenerateHardwareRequest {
```
andreaturli: can you add some javadoc for this class? I don't understand its usage

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/compute/OneandoneComputeServiceAdapter.java
alibazlamit: nothing happens

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/compute/OneandoneComputeServiceAdapter.java
alibazlamit: it will remain powered off

alibazlamit:
@andreaturli a few of your comments do make sense about this PR, i will address those issues so we can move on.

I don't see why i would separate Recovery Images it is a very small feature and i am not sure that will really affect reviewing the PR. In my opinion i think that most of the things that you are asking me to change will make reviewing this PR difficult, like changing firewall policies and moving few lines of code into a separate method, changes like that made this PR look messy and increase the diff.

I agree with you about being committed to jClouds best practices, but from my experience this is not the right PR to go back and review things that were already accepted by jClouds.

I have addressed all the small changes that make sense, i can post the live test results yet again.I hope you agree with me about closing this PR before its gets even worst

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/domain/Types.java
alibazlamit: done

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/domain/Types.java
alibazlamit: done

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/domain/Types.java
alibazlamit: done

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/compute/OneandoneComputeServiceAdapter.java
alibazlamit: changed

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/compute/OneandoneComputeServiceAdapter.java
alibazlamit: no need as the PR will get even more complicated to review

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/compute/OneandoneComputeServiceAdapter.java
alibazlamit: I suggest we open a separate PR addressing any backwards concern you might have

oneandone/src/main/java/org/apache/jclouds/oneandone/rest/compute/function/ServerToNodeMetadata.java
alibazlamit: changed

PR #433, opened 2018-04-12
JCLOUDS-1406 - Add default location configuration to dimension data provider

trevorflanagan:


dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/config/DimensionDataCloudControlHttpApiModule.java
```diff
@@ -40,6 +43,17 @@
...
 
+   @Override
+   protected void installLocations() {
+      super.installLocations();
+      //      bind(RegionIdToZoneIdsSupplier.class).to(ZonesForRegion.class).in(Scopes.SINGLETON);
+      bind(RegionIdToURISupplier.class).to(RegionsToApiEndpoints.class).in(Scopes.SINGLETON);
+      //      bind(ZoneIdsSupplier.class).to(ZoneIdsFromRegionIdToZoneIdsValues.class).in(Scopes.SINGLETON);
+      //      bind(RegionIdsSupplier.class).to(RegionIdsFromRegionIdToURIKeySet.class).in(Scopes.SINGLETON);
+      //      bind(ZoneIdToURISupplier.class).to(ZoneIdToURIFromJoinOnRegionIdToURI.class).in(Scopes.SINGLETON);
+      //      bind(ImplicitLocationSupplier.class).to(OnlyLocationOrFirstZone.class).in(Scopes.SINGLETON);
```
nacx: I'd say you need this four supplier bindings, as the endpoints are "per-region", not "per-zone".

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/filters/DatacenterIdFilter.java
```diff
@@ -17,31 +17,51 @@
...
 
    @Override
    public HttpRequest filter(HttpRequest request) throws HttpException {
-      Set<String> datacenterIds = datacenterIdsSupplier.get();
       if (datacenterIds != null && !datacenterIds.isEmpty()) {
-         return request.toBuilder().addQueryParam("datacenterId", datacenterIds).build();
+         return request.toBuilder().addQueryParam(queryParam, datacenterIds).build();
       } else {
          return request;
       }
```
nacx: Having this as a filter might be problematic in the future, as the filter works with the "default region". When users start using the ComputeService and configure the TempalteOptions to deploy to a concrete region, there would be no way to call the InfraApi (or any API using this filter) and make it filter by the configured region.
It would be better to remove this filter and add the datacenterId as method parameters in the API.

trevorflanagan: @nacx as agreed I have removed both of the HTTP filters.

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/config/DimensionDataCloudControlHttpApiModule.java
```diff
@@ -40,6 +51,16 @@
 @ConfiguresHttpApi
 public class DimensionDataCloudControlHttpApiModule extends HttpApiModule<DimensionDataCloudControlApi> {
 
+   @Override
+   protected void installLocations() {
+      super.installLocations();
+      bind(RegionIdToURISupplier.class).to(RegionsToApiEndpoints.class).in(Scopes.SINGLETON);
+      bind(ZoneIdsSupplier.class).to(ZoneIdsFromRegionIdToZoneIdsValues.class).in(Scopes.SINGLETON);
+      bind(RegionIdsSupplier.class).to(RegionIdsFromRegionIdToURIKeySet.class).in(Scopes.SINGLETON);
```
nacx: Remove this binding, as it is the one causing the circular dependency.

The `RegionsToApiEndpoints` supplier that provides the endpoint for each region uses the `@Region Supplier<Set<String>>` (the `RegionIdsSupplier`). but with this binding it was being configured to get the regions from the `Map<String, Supplier<URI>>` generated by the `RegionIdToURISupplier`, thus the circular dependency.

The default implementation of the `RegionIdsSupplier` just populates the list of regions based on the value of the `jclouds.regions` property, which is what we want here, so removing the binding should leave all good and the regions properly configured.

nacx: With this bind removed the locations look properly populated. Tested with:
```java
public static void main(String[] args) {
      try (ApiContext<DimensionDataCloudControlApi> ctx =
              ContextBuilder.newBuilder("dimensiondata-cloudcontrol")
                 .credentials("foo", "bar") // Not needed. We won't call the API
                 .build()) {
         Supplier<Set<? extends Location>> locations = ctx.utils().injector()
               .getInstance(Key.get(new TypeLiteral<Supplier<Set<? extends Location>>>() {
               }, Memoized.class));
         
         for (Location l : locations.get()) {
            System.out.println(l);
         }
      }
   }
```
Prints:
```
{scope=REGION, id=na, description=na, parent=dimensiondata-cloudcontrol, iso3166Codes=[NA]}
{scope=REGION, id=eu, description=eu, parent=dimensiondata-cloudcontrol, iso3166Codes=[EU]}
{scope=REGION, id=au, description=au, parent=dimensiondata-cloudcontrol, iso3166Codes=[AU]}
{scope=REGION, id=mea, description=mea, parent=dimensiondata-cloudcontrol, iso3166Codes=[AF]}
{scope=REGION, id=ap, description=ap, parent=dimensiondata-cloudcontrol, iso3166Codes=[AP]}
{scope=REGION, id=canada, description=canada, parent=dimensiondata-cloudcontrol, iso3166Codes=[CA]}
{scope=ZONE, id=NA9, description=NA9, parent=na, iso3166Codes=[NA]}
{scope=ZONE, id=NA12, description=NA12, parent=na, iso3166Codes=[NA]}
{scope=ZONE, id=EU6, description=EU6, parent=eu, iso3166Codes=[EU]}
{scope=ZONE, id=EU7, description=EU7, parent=eu, iso3166Codes=[EU]}
{scope=ZONE, id=EU8, description=EU8, parent=eu, iso3166Codes=[EU]}
{scope=ZONE, id=AU9, description=AU9, parent=au, iso3166Codes=[AU]}
{scope=ZONE, id=AU10, description=AU10, parent=au, iso3166Codes=[AU]}
{scope=ZONE, id=AU11, description=AU11, parent=au, iso3166Codes=[AU]}
{scope=ZONE, id=AF3, description=AF3, parent=mea, iso3166Codes=[AF]}
{scope=ZONE, id=AP4, description=AP4, parent=ap, iso3166Codes=[AP]}
{scope=ZONE, id=AP5, description=AP5, parent=ap, iso3166Codes=[AP]}
{scope=ZONE, id=CA2, description=CA2, parent=canada, iso3166Codes=[CA]}
```

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/DimensionDataCloudControlProviderMetadata.java
```diff
@@ -42,12 +47,26 @@ public DimensionDataCloudControlProviderMetadata() {
...
+      properties.setProperty(PROPERTY_REGION + ".eu." + ISO3166_CODES, "EU");
+      properties.setProperty(PROPERTY_REGION + ".eu.zones", "EU6,EU7,EU8");
+      properties.setProperty(PROPERTY_REGION + ".au." + ISO3166_CODES, "AU");
+      properties.setProperty(PROPERTY_REGION + ".au.zones", "AU9,AU10,AU11");
+      properties.setProperty(PROPERTY_REGION + ".mea." + ISO3166_CODES, "AF");
+      properties.setProperty(PROPERTY_REGION + ".mea.zones", "AF3");
+      properties.setProperty(PROPERTY_REGION + ".ap." + ISO3166_CODES, "AP");
+      properties.setProperty(PROPERTY_REGION + ".ap.zones", "AP4,AP5");
+      properties.setProperty(PROPERTY_REGION + ".canada." + ISO3166_CODES, "CA");
+      properties.setProperty(PROPERTY_REGION + ".canada.zones", "CA2");
```
nacx: Consider fixing the ISO3166 code values and put the right ones for each physical region and zone: https://en.wikipedia.org/wiki/ISO_3166-2

You can configure ISO3166 codes for each zone by adding properties such as:
```java
properties.setProperty(PROPERTY_ZONE + "EU7" + ISO3166_CODES, "DE-BE");
```

nacx:
FTR, there was a small bug in the `LocationsSupplier` that did not properly filter the zones by the values configured in the `jclouds.zones` property. It is fixed in https://github.com/jclouds/jclouds/pull/1195.

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/config/DimensionDataCloudControlHttpApiModule.java
trevorflanagan: Excellent, now the structure becomes very clear to me.

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/DimensionDataCloudControlProviderMetadata.java
trevorflanagan: thanks @nacx I can try figure those out.

trevorflanagan:
> Change looks good. Would it make sense, though, to create a couple filter objects ListServerOptions and ListDatacenterOptions that extend PaginationOptions and introduce the datacenter query parameters? if the datacenter filter parameter is optional, then it would be better and cleaner for the API to just hide it in the "options" object passed to the list method?

@nacx I gave the change you suggested a go, but I don't think this solution will work cleanly as you or I might have hoped. I face an issue in the [`org.jclouds.dimensiondata.cloudcontrol.domain.PaginatedCollection#toPaginationOptions`](https://github.com/DimensionDataDublin/jclouds-labs/blob/zones/dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/domain/PaginatedCollection.java#L101) method as it returns a `PaginationOptions` instead of the Subtype eg ListDatacenterOptions. We need the correct subtype for the cast to work in the Parsers. To resolve the issue we would also need to us having to create subclasses of  `org.jclouds.dimensiondata.cloudcontrol.domain.PaginatedCollection` also for both of the filters we have already removed. 

Maybe an alternative approach could be to modify and rename `org.jclouds.dimensiondata.cloudcontrol.options.PaginationOptions` so that it has the following ability:

``` 
public PaginationOptions queryParameters(String queryParameter, Set<String> values) {
      for (String value : values) {
         queryParameters.put(queryParameter, value);
      }
      return this;
   }
``` 
This means we have the extra flexibility without impacting other usages of `PaginationOptions`. This would be my preferred approach.

Another way to resolve could be to re-build query class each time we iterate, but I dont really like this option:
```
   final class ParseCustomerImages extends ParseJson<CustomerImages> {

      @Inject
      ParseCustomerImages(Json json) {
         super(json, TypeLiteral.get(CustomerImages.class));
      }

      private static class ToPagedIterable extends Arg0ToPagedIterable<CustomerImage, ToPagedIterable> {

         private DimensionDataCloudControlApi api;

         @Inject
         ToPagedIterable(final DimensionDataCloudControlApi api) {
            this.api = api;
         }

         @Override
         protected Function<Object, IterableWithMarker<CustomerImage>> markerToNextForArg0(
               final Optional<Object> arg0) {
            return new Function<Object, IterableWithMarker<CustomerImage>>() {
               @Override
               public IterableWithMarker<CustomerImage> apply(Object input) {
                  ListWithDatacenterIdFilteringAndPaginationOptions options = new ListWithDatacenterIdFilteringAndPaginationOptions(PaginationOptions.class
                                          .cast(input),(Set<String>) arg0.get() ) ;
                  return api.getServerImageApi().listCustomerImages(options);
               }
            };
         }
      }
   }
```

nacx:
I see what you mean.

Is possible, I'd like to hide such generic `queryParameter` methods to let users have a high-level API that is easily understandable.

What about this:

Now we are using the `Arg0ToPagedIterable` class because we only need the first parameter passed to the method when creating a request for the next page. The second argument (the query parameters holder), is just the same object we use as a "marker" in the `IterableWithMarker`, so we pass it directly. This is convenient, but it does not have to be this way.

We could have just a plain object `PaginationOptions` that did **not** extend the `BaseHttpRequestOptions`, just a plain object we use as the "marker to next page" and let the `ItetableWithMarker` return that obejct.

Then we could have specific filter objects just like `DatacenterListFilters` with a method `pagination(PaginationOptions)` that configures the query parameters accoding to the pagination options. Up to this point is just the same as the last proposal you mentioned, but there is one small thing to change. We could change the `ToPagedIterable` to extend `ArgsToPagedIterable` instead. That would change the method to `markerToNextForArgs` and it would recieve not only the first argument passed to the method, but all arguments.

There you could take the method arguments and also the `DatacenterListFilters` argument as passed by the user and just override in the latter the pagination options as they come from the `IterableWithMarker`. This is pretty much the same than your last proposal but without having to revuild the filter options object. In each iteration we should be just modifying its pagination parameters.

WDYT?

trevorflanagan:
@nacx - I have taken a look at what you are proposing. I do see the value in what you suggest. However I am slightly confused about how I might complete this. I have [pushed a new branch](https://github.com/DimensionDataDublin/jclouds-labs/tree/zones_review_updates_for_nacx) that maybe you could look at to help you understand where I am at.

First thing, I have created a new class `PaginationFilterOptions` which is more or less the same as existing `PaginationOptions` but does not extend `BaseHttpRequestOptions`. The reason for the new class is that we use `PaginationOptions` on other APIs that at the moment I do not want to change since we do have the same requirements for datacenter filtering. I have also introduced a `DatacenterIdListFilters` which does extend `BaseHttpRequestOptions` and wraps the `PaginationFilterOptions` so that it can be used for pagination.

Taking the example of [ServerApi ](https://github.com/DimensionDataDublin/jclouds-labs/blob/zones/dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/features/ServerApi.java) as it looks on this PR we expose two methods for `listServers`. The first one is what I think is the user facing API one with the Set of `datacenterIds` we want to query on:
```
   @Named("server:list")
   @GET
   @Path("/server")
   @Transform(ParseServers.ToPagedIterable.class)
   @ResponseParser(ParseServers.class)
   @Fallback(Fallbacks.EmptyPagedIterableOnNotFoundOr404.class)
   PagedIterable<Server> listServers(@QueryParam("datacenterId") Set<String> datacenterId);
```
The second is the one as used in order to iterate over the pages returned:
```
   @Named("server:list")
   @GET
   @Path("/server")
   @ResponseParser(ParseServers.class)
   @Fallback(Fallbacks.EmptyIterableWithMarkerOnNotFoundOr404.class)
   PaginatedCollection<Server> listServers(@QueryParam("datacenterId") Set<String> datacenterId,
         PaginationOptions options);
```
With the changes you propose I can see this working as below. I'm hoping this is the parameters of the two listServer methods.

```
   @Named("server:list")
   @GET
   @Path("/server")
   @ResponseParser(ParseServers.class)
   @Fallback(Fallbacks.EmptyIterableWithMarkerOnNotFoundOr404.class)
   PaginatedCollection<Server> listServers(DatacenterIdListFilters datacenterIdListFilters);

   @Named("server:list")
   @GET
   @Path("/server")
   @Transform(ParseServers.ToPagedIterable.class)
   @ResponseParser(ParseServers.class)
   @Fallback(Fallbacks.EmptyPagedIterableOnNotFoundOr404.class)
   PagedIterable<Server> listServers(@QueryParam("datacenterId") Set<String> datacenterId);

   @Singleton
   final class ParseServers extends ParseJson<Servers> {

      @Inject
      ParseServers(final Json json) {
         super(json, TypeLiteral.get(Servers.class));
      }

      static class ToPagedIterable extends Arg0ToPagedIterable<Server, ToPagedIterable> {

         private DimensionDataCloudControlApi api;

         @Inject
         ToPagedIterable(final DimensionDataCloudControlApi api) {
            this.api = api;
         }

         @Override
         protected Function<Object, IterableWithMarker<Server>> markerToNextForArg0(final Optional<Object> arg0) {
            return new Function<Object, IterableWithMarker<Server>>() {
               @Override
               public IterableWithMarker<Server> apply(Object input) {
                  return api.getServerApi().listServers(
                        DatacenterIdListFilters.Builder.datacenterId((Set<String>) arg0.get())
                              .paginationOptions(PaginationFilterOptions.class.cast(input)));
               }
            };
         }
      }

   }
```

nacx:
The idea is to make "optional" parameters or filters really optional and have a clean API. That's the reasoning behind refactoring them into an options object. With this in mind, the code in your branch looks good, but I'd change the signatures to:

```java
@Named("server:list")
@GET
@Path("/server")
@ResponseParser(ParseServers.class)
@Fallback(Fallbacks.EmptyIterableWithMarkerOnNotFoundOr404.class)
PaginatedCollection<Server> listServers(DatacenterIdListFilters datacenterIdListFilters);

@Named("server:list")
@GET
@Path("/server")
@Transform(ParseServers.ToPagedIterable.class)
@ResponseParser(ParseServers.class)
@Fallback(Fallbacks.EmptyPagedIterableOnNotFoundOr404.class)
PagedIterable<Server> listServers();
```

I mean, if the filter is optional, just provide a method with no arguments (no filters). If users want to filter they can use the method with the options parameter. You can then have the function that advances take into account the presence or absence of the filters parameter:

```java
@Override
protected Function<Object, IterableWithMarker<Server>> markerToNextForArg0(final Optional<Object> arg0) {
   return new Function<Object, IterableWithMarker<Server>>() {
      @Override
      public IterableWithMarker<Server> apply(Object input) {
         DatacenterIdListFilters filters = arg0.isPresent() ? (DatacenterIdListFilters) input : new DatacenterIdListFilters();
         filters.paginationOptions((PaginationFilterOptions) input);
         return api.getServerApi().listServers(filters);
      }
   };
}
```

There is also no need to duplicate the `PaginationOptions` class if it is still used in other APIs (making it not extend the base options class was just a suggestion to illustrate it is not a need for a "marker" to the next page). Let's just remove the new Pagination options class and let the Ddatacenter filters builder accept the existing one. You can easily add all the pagination query params to the current filter query parameter map and there won't be duplicated code.

trevorflanagan:
rebuild please

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/options/DatacenterIdListFilters.java
```diff
@@ -0,0 +1,68 @@
...
+      if (paginationOptions.orderBy() != null) {
+         this.queryParameters.put("orderBy", paginationOptions.orderBy());
+      }
+      if (paginationOptions.pageNumber() != null) {
+         this.queryParameters.put("pageNumber", paginationOptions.pageNumber());
+      }
+      if (paginationOptions.pageSize() != null) {
+         this.queryParameters.put("pageSize", paginationOptions.pageSize());
+      }
+      return this;
```
nacx: Just:

```java
public DatacenterIdListFilters paginationOptions(final PaginationOptions paginationOptions) {
   this.queryParameters.putAll(paginationOptions.buildQueryParameters());
   return this;
}
```

To avoid duplicating logic.

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/options/DatacenterIdListFilters.java
```diff
@@ -0,0 +1,68 @@
...
+import java.util.Set;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+public class DatacenterIdListFilters extends BaseHttpRequestOptions {
+
+   private DatacenterIdListFilters() {
+   }
+
+   private DatacenterIdListFilters datacenterIds(final Set<String> datacenterIds) {
```
nacx: If the common use case will be that there is just one ID to pass, consider providing a varargs version of the method too, to make it cleaner and easier to consume.

dimensiondata/src/test/java/org/jclouds/dimensiondata/cloudcontrol/features/InfrastructureApiLiveTest.java
```diff
@@ -52,7 +54,8 @@ public void testListOperatingSystems() {
       for (Datacenter dc : getDatacenters()) {
          datacenterIds.add(dc.id());
       }
-      FluentIterable<OperatingSystem> operatingSystems = api().listOperatingSystems(datacenterIds).concat();
+      ImmutableList<OperatingSystem> operatingSystems = api()
+            .listOperatingSystems(DatacenterIdListFilters.Builder.datacenterId(datacenterIds)).toList();
```
nacx: [minor] static importing the `datacenterId` method makes the invocation cleaner. Not something to be changed, just a comment on how, in general, we use this kind of option objects in jclouds.

trevorflanagan:
Suggested changes made @nacx, thanks. I have also re-based onto master and squashed commits.

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/DimensionDataCloudControlProviderMetadata.java
```diff
@@ -42,12 +48,38 @@ public DimensionDataCloudControlProviderMetadata() {
...
 
-   public DimensionDataCloudControlProviderMetadata(Builder builder) {
+   public DimensionDataCloudControlProviderMetadata(final Builder builder) {
       super(builder);
    }
 
    public static Properties defaultProperties() {
       Properties properties = DimensionDataCloudControlApiMetadata.defaultProperties();
+      properties.setProperty(PROPERTY_REGIONS, "na,eu,au,mea,ap,canada");
+      properties.setProperty(PROPERTY_ISO3166_CODES, "NA,EU,AU,AF,AP,CA");
```
nacx: Since the regions expand several countries I think we can't really put a proper ISO code for them (like "US", "DE", etc). I'd remove the ISO code properties for the regions and just leave them for the zones. Makes sense?

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/options/DatacenterIdListFilters.java
```diff
@@ -0,0 +1,58 @@
...
+import org.jclouds.http.options.BaseHttpRequestOptions;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+public class DatacenterIdListFilters extends BaseHttpRequestOptions {
+
+   private DatacenterIdListFilters() {
+   }
+
+   private DatacenterIdListFilters datacenterIds(final String... datacenterIds) {
```
nacx: Make this public, otherwise, if you use the builder to configure the pagination first you won't be able to configure the ids.

Also consider providing the method that accepts a collection of IDs too, as that's convenient since the APIs usually return collections. In my previous comment, I pretended to suggest to "add" the varargs method, not to replace the current one. If you add ti back, consider using `Collection` in the signature instead of `Set`.

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/options/IdListFilters.java
```diff
@@ -0,0 +1,59 @@
...
+import org.jclouds.http.options.BaseHttpRequestOptions;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+public class IdListFilters extends BaseHttpRequestOptions {
+
+   private IdListFilters() {
+   }
+
+   private IdListFilters ids(final String... ids) {
```
nacx: Same comments here.

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/DimensionDataCloudControlProviderMetadata.java
trevorflanagan: Makes sense I can do this update

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/options/DatacenterIdListFilters.java
trevorflanagan: Ah I misunderstood. Thanks, I will make this change.

trevorflanagan:
Appreciate all the help with this one @nacx , I've learnt alot about jclouds working on it.

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/DimensionDataCloudControlProviderMetadata.java
```diff
@@ -42,12 +47,37 @@ public DimensionDataCloudControlProviderMetadata() {
...
    }
 
    public static Properties defaultProperties() {
       Properties properties = DimensionDataCloudControlApiMetadata.defaultProperties();
+      properties.setProperty(PROPERTY_REGIONS, "na,eu,au,mea,ap,canada");
+      properties.setProperty(PROPERTY_REGION + ".na." + ISO3166_CODES, "NA");
+      properties.setProperty(PROPERTY_REGION + ".na.zones", "NA9,NA12");
+      properties.setProperty(PROPERTY_ZONE + "NA9" + ISO3166_CODES, "US-VA");
+      properties.setProperty(PROPERTY_ZONE + "NA12" + ISO3166_CODES, "US-CA");
+      properties.setProperty(PROPERTY_REGION + ".eu." + ISO3166_CODES, "EU");
```
nacx: Remove

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/DimensionDataCloudControlProviderMetadata.java
```diff
@@ -42,12 +47,37 @@ public DimensionDataCloudControlProviderMetadata() {
...
+      properties.setProperty(PROPERTY_REGION + ".eu.zones", "EU6,EU7,EU8");
+      properties.setProperty(PROPERTY_ZONE + "EU6" + ISO3166_CODES, "DE-HE");
+      properties.setProperty(PROPERTY_ZONE + "EU7" + ISO3166_CODES, "NL-NH");
+      properties.setProperty(PROPERTY_ZONE + "EU8" + ISO3166_CODES, "BE-BRU");
+      properties.setProperty(PROPERTY_REGION + ".au." + ISO3166_CODES, "AU");
```
nacx: Remove

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/DimensionDataCloudControlProviderMetadata.java
```diff
@@ -42,12 +47,37 @@ public DimensionDataCloudControlProviderMetadata() {
...
+      properties.setProperty(PROPERTY_REGION + ".au.zones", "AU9,AU10,AU11");
+      properties.setProperty(PROPERTY_ZONE + "AU9" + ISO3166_CODES, "AU-NSW");
+      properties.setProperty(PROPERTY_ZONE + "AU10" + ISO3166_CODES, "AU-VIC");
+      properties.setProperty(PROPERTY_ZONE + "AU11" + ISO3166_CODES, "NZ-WKO");
+      properties.setProperty(PROPERTY_REGION + ".mea." + ISO3166_CODES, "AF");
```
nacx: Remove

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/DimensionDataCloudControlProviderMetadata.java
```diff
@@ -42,12 +47,37 @@ public DimensionDataCloudControlProviderMetadata() {
...
+      properties.setProperty(PROPERTY_REGION + ".mea.zones", "AF3");
+      properties.setProperty(PROPERTY_ZONE + "AF3" + ISO3166_CODES, "ZA-GT");
+      properties.setProperty(PROPERTY_REGION + ".ap." + ISO3166_CODES, "AP");
```
nacx: Remove

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/DimensionDataCloudControlProviderMetadata.java
```diff
@@ -42,12 +47,37 @@ public DimensionDataCloudControlProviderMetadata() {
...
+      properties.setProperty(PROPERTY_REGION + ".ap.zones", "AP4,AP5");
+      properties.setProperty(PROPERTY_ZONE + "AP4" + ISO3166_CODES, "JP-13");
+      properties.setProperty(PROPERTY_ZONE + "AP5" + ISO3166_CODES, "HK");
+      properties.setProperty(PROPERTY_REGION + ".canada." + ISO3166_CODES, "CA");
```
nacx: Remove

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/DimensionDataCloudControlProviderMetadata.java
```diff
@@ -42,12 +47,37 @@ public DimensionDataCloudControlProviderMetadata() {
...
 
-   public DimensionDataCloudControlProviderMetadata(Builder builder) {
+   public DimensionDataCloudControlProviderMetadata(final Builder builder) {
       super(builder);
    }
 
    public static Properties defaultProperties() {
       Properties properties = DimensionDataCloudControlApiMetadata.defaultProperties();
+      properties.setProperty(PROPERTY_REGIONS, "na,eu,au,mea,ap,canada");
+      properties.setProperty(PROPERTY_REGION + ".na." + ISO3166_CODES, "NA");
```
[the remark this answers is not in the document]
trevorflanagan: Ahh thanks. I have removed.

nacx:
Pushed to [master](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/60c3323c) and [2.1.x](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/5c97ea45). Thanks @trevorflanagan!

PR #434, opened 2018-04-13
JCLOUDS-1407 - Dimensiondata Server API 2.6 support

btrishkin:
Server API 2.6 support implementation


dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/domain/IdeController.java
```diff
@@ -0,0 +1,72 @@
...
+
+import com.google.auto.value.AutoValue;
+import com.google.common.collect.ImmutableList;
+import org.jclouds.javax.annotation.Nullable;
+import org.jclouds.json.SerializedNames;
+
+import java.util.List;
+
+@AutoValue
+public abstract class IdeController extends AbstractBaseController {
```
nacx: Inheritance with Google auto value is not supported. You already have to define the parent class fields in the child class factory methods and builders, wo there is little gain abstracting the common accessors in the base class. There are just a few, so consider removing the base calss and putting all information here.

Also, inheritance is tricky: we annotate the "factory methods" with the json field names, not the individual fields. This is good for deserialization but when serializing we need to match each object field to its json name. Ideally we would annotate each field method, but that would mean duplicating the information from the factory method annotation. Instead, assuming the fact that Google Auto does not support inheritance, in jclouds we took the convention that object fields (well, abstract field accessors) would be decalred *in the same order* than the one of the factory method, and jclouds uses that convention when serializing. When there is a class hierarchy, field order is not that clear and more error-prone.

nacx: Just seeing the classes below, you need to configure the order of the parameters in the factory methods matching the order in which the fields are declared. Double check that for all classes.

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/features/ServerApi.java
```diff
@@ -61,7 +61,7 @@
 
 @RequestFilters({ BasicAuthentication.class, OrganisationIdFilter.class })
 @Consumes(MediaType.APPLICATION_JSON)
-@Path("/{jclouds.api-version}/server")
+@Path("/2.6/server")
```
nacx: Change the default version in the API metadata instead of doing this. This way, if users want to use the old path they can just override the property when creating the context. All new fields in the Server object are optional, so that shouldn't be an issue.

nacx: Do all APIs need to us ethe same version? If so, youd it make sense to store the version in a constant and reference it from all APIs?

btrishkin: Well, we implement them in chunks. This PR covers only ServerApi feature which will be updated to new version. Everything else remain old 2.4 for now. We can introduce constant with newest API version supported, but outdated features will have to have 2.4 version hardcoded in them until we implement latest version of them. Would that be ok?

nacx: Understood. Let's leave it as-is then for now.

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/domain/IdeController.java
```diff
@@ -18,42 +18,50 @@
...
 
 @AutoValue
-public abstract class IdeController extends AbstractBaseController {
+public abstract class IdeController {
 
-   public abstract Integer channel();
+   public abstract String id();
 
-   @Nullable
-   public abstract List<AbstractDrive> deviceOrDisks();
```
nacx: Oh, I see now that inheritance here was there not only to reuse fields but because the API returns collections with different object types in it.
To avoid having a `List<Object>`, which makes the API hard to consume, would it make sense to have just one object with all fields? I see the child classes just had one or two different fields. Would that make sense?
If not, I'd consider getting the abstract classes back and just be extra-careful in variable order. take into account that the fields in the base class are serialized first, then the ones in the child class (and if at some point the base class is modified, the annotations in all child classes need to be updated accordingly).
Apologies for the noise if you consider the second option to be better, but till now it wasn't clear for me the purpose of the abstract classes.

btrishkin: Yes, AbstractDrive will help a lot in ServerToHardware function implementation for a new ServerApi version. Gonna keep it then.

btrishkin:
New PR (https://github.com/jclouds/jclouds-labs/pull/447) created instead of this one, as it's quite outdated

PR #436, opened 2018-05-19
Upgrade to java8

andreaturli:
- bump guava version to 21.0
- fix DimensionData test

this is required for https://github.com/jclouds/jclouds/pull/1197

### DIFF
--- a/dimensiondata/src/test/java/org/jclouds/dimensiondata/cloudcontrol/internal/BaseDimensionDataCloudControlApiLiveTest.java
+++ b/dimensiondata/src/test/java/org/jclouds/dimensiondata/cloudcontrol/internal/BaseDimensionDataCloudControlApiLiveTest.java
@@ -40,7 +40,7 @@ import org.testng.annotations.Test;
 import java.util.Properties;
 import java.util.Set;
 
-import static com.google.common.util.concurrent.MoreExecutors.sameThreadExecutor;
+import static com.google.common.util.concurrent.MoreExecutors.newDirectExecutorService;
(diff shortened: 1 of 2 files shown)


andreaturli:
cool, thanks @nacx and @gaul merging now

andreaturli:
merged at [master](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/1b5dd9fe)

demobox:
> bump guava version to 21.0

@andreaturli Just to clarify: is this part of this PR? I didn't see any code change related to that, but it may have been removed or after an underlying merge.

dimensiondata/src/test/java/org/jclouds/dimensiondata/cloudcontrol/internal/BaseDimensionDataCloudControlApiLiveTest.java
```diff
@@ -40,7 +40,7 @@
 import java.util.Properties;
 import java.util.Set;
 
-import static com.google.common.util.concurrent.MoreExecutors.sameThreadExecutor;
+import static com.google.common.util.concurrent.MoreExecutors.newDirectExecutorService;
```
andreaturli: @demobox I think this is required because of guava 21 inherited from jclouds/jclouds

andreaturli:
@demobox I think guava21 updates are incidentally required as we updated guava21 in jclouds/jclouds, so technically the PR doesn't bump guava version for jclouds-labs in its pom.xml but it happens because of the upstream changes. makes sense?

demobox:
> makes sense?

Yes, clear, thanks. Just wanted to make sure we weren't missing any code that we were intending to add ;-)

andreaturli:
sure thanks for double-checking it

PR #439, opened 2018-07-04
JCLOUDS-1432 - handle RESOURCE_NOT_FOUND,

FileIOUtility:
do not expect 404 when assets not found,
initially done for Server API only

### DIFF
--- a/dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/features/ServerApi.java
+++ b/dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/features/ServerApi.java
@@ -68,7 +68,6 @@ public interface ServerApi {
    @GET
    @Path("/server")
    @ResponseParser(ParseServers.class)
-   @Fallback(Fallbacks.EmptyIterableWithMarkerOnNotFoundOr404.class)
    PaginatedCollection<Server> listServers(DatacenterIdListFilters datacenterIdListFilters);
 
    @Named("server:list")
@@ -76,7 +75,6 @@ public interface ServerApi {
    @Path("/server")
    @Transform(ParseServers.ToPagedIterable.class)
    @ResponseParser(ParseServers.class)
-   @Fallback(Fallbacks.EmptyPagedIterableOnNotFoundOr404.class)
    PagedIterable<Server> listServers();
 
    @Named("server:get")
--- a/dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/handlers/DimensionDataCloudControlErrorHandler.java
+++ b/dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/handlers/DimensionDataCloudControlErrorHandler.java
@@ -26,6 +26,8 @@ import org.jclouds.rest.ResourceNotFoundException;
 
 import javax.inject.Singleton;
 
+import java.util.ConcurrentModificationException;
+
 import static org.jclouds.http.HttpUtils.closeClientButKeepContentStream;
 
 /**
@@ -47,16 +49,22 @@ public class DimensionDataCloudControlErrorHandler implements HttpErrorHandler {
             String.format("%s -> %s", command.getCurrentRequest().getRequestLine(), response.getStatusLine());
       switch (response.getStatusCode()) {
          case 400:
-            if (message.contains("RESOURCE_NOT_FOUND") || message.contains("OPERATION_NOT_SUPPORTED")) {
+            if (message.contains("RESOURCE_NOT_FOUND")) {
                exception = new ResourceNotFoundException(message, exception);
-            } else if (message.contains("INVALID_INPUT_DATA") || message.contains("ORGANIZATION_NOT_VERIFIED")
-                  || (message.contains("SYSTEM_ERROR") && !message.contains("RETRYABLE_SYSTEM_ERROR")) || message
-                  .contains("CPU_SPEED_NOT_AVAILABLE") || message.contains("CONFIGURATION_NOT_SUPPORTED")) {
+            } else if (message.contains("OPERATION_NOT_SUPPORTED")) {
+               exception = new UnsupportedOperationException(message, exception);
+            } else if (message.contains("RESOURCE_BUSY")) {
+               exception = new ConcurrentModificationException(message, exception);
+            } else if (message.contains("RESOURCE_LOCKED")) {
                exception = new IllegalStateException(message, exception);
-            } else if (message.contains("RESOURCE_BUSY") || message.contains("UNEXPECTED_ERROR")) {
-               exception = new ResourceNotFoundException(message, exception);
             } else if (message.contains("NAME_NOT_UNIQUE")) {
                exception = new ResourceAlreadyExistsException(message, exception);
+            } else if (message.contains("UNEXPECTED_ERROR")
+                  || message.contains("RETRYABLE_SYSTEM_ERROR")
+                  || message.contains("SYSTEM_ERROR")) {
+               break;
+            } else {
+               exception = new IllegalArgumentException(message, exception);
             }
             break;
          case 401:
@@ -66,9 +74,9 @@ public class DimensionDataCloudControlErrorHandler implements HttpErrorHandler {
             exception = new AuthorizationException(message, exception);
             break;
          case 404:
-            if (!command.getCurrentRequest().getMethod().equals("DELETE")) {
-               exception = new ResourceNotFoundException(message, exception);
-            }
+            // CloudControl uses error code 400 with RESOURCE_NOT_FOUND to report missing assets
+            // 404 means malformed URI only
+            exception = new IllegalArgumentException(message, exception);
             break;
       }
       command.setException(exception);
--- a/dimensiondata/src/test/java/org/jclouds/dimensiondata/cloudcontrol/features/ServerApiMockTest.java
+++ b/dimensiondata/src/test/java/org/jclouds/dimensiondata/cloudcontrol/features/ServerApiMockTest.java
@@ -26,11 +26,10 @@ import org.jclouds.dimensiondata.cloudcontrol.domain.Server;
 import org.jclouds.dimensiondata.cloudcontrol.domain.options.CloneServerOptions;
 import org.jclouds.dimensiondata.cloudcontrol.domain.options.CreateServerOptions;
 import org.jclouds.dimensiondata.cloudcontrol.internal.BaseAccountAwareCloudControlMockTest;
+import org.jclouds.http.HttpResponseException;
 import org.jclouds.http.Uris;
-import org.jclouds.rest.ResourceNotFoundException;
 import org.testng.annotations.Test;
 
-import javax.ws.rs.HttpMethod;
 import java.util.List;
 
 import static javax.ws.rs.HttpMethod.GET;
@@ -39,6 +38,7 @@ import static org.assertj.core.api.Assertions.failBecauseExceptionWasNotThrown;
 import static org.jclouds.dimensiondata.cloudcontrol.options.DatacenterIdListFilters.Builder.datacenterId;
 import static org.testng.Assert.assertEquals;
 import static org.testng.Assert.assertNotNull;
+import static org.testng.Assert.assertNull;
 import static org.testng.Assert.assertTrue;
 
 /**
@@ -55,8 +55,8 @@ public class ServerApiMockTest extends BaseAccountAwareCloudControlMockTest {
       try {
          serverApi().deployServer(ServerApiMockTest.class.getSimpleName(), "imageId", true, networkInfo,
                "administratorPassword");
-         failBecauseExceptionWasNotThrown(ResourceNotFoundException.class);
-      } catch (ResourceNotFoundException e) {
+         failBecauseExceptionWasNotThrown(HttpResponseException.class);
+      } catch (HttpResponseException e) {
          assertNotNull(e);
          assertSent(POST, "/caas/2.4/6ac1e746-b1ea-4da5-a24e-caf1a978789d/server/deployServer");
       }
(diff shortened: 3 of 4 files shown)


trevorflanagan:
This is @FileIOUtility first PR for Apache jclouds but he has been working on several aspects of the project for a while now.

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/features/ServerApi.java
```diff
@@ -68,21 +68,19 @@
    @GET
    @Path("/server")
    @ResponseParser(ParseServers.class)
-   @Fallback(Fallbacks.EmptyIterableWithMarkerOnNotFoundOr404.class)
```
nacx: This is a behavior change since this method will now fail upon 4xx responses, when previously it just returned an empty list. Consider adding the corresponding fallback and keep returning empty lists.

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/features/ServerApi.java
```diff
@@ -68,21 +68,19 @@
...
    PaginatedCollection<Server> listServers(DatacenterIdListFilters datacenterIdListFilters);
 
    @Named("server:list")
    @GET
    @Path("/server")
    @Transform(ParseServers.ToPagedIterable.class)
    @ResponseParser(ParseServers.class)
-   @Fallback(Fallbacks.EmptyPagedIterableOnNotFoundOr404.class)
```
nacx: Same here

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/handlers/DimensionDataCloudControlErrorHandler.java
```diff
@@ -66,9 +66,9 @@ public void handleError(HttpCommand command, HttpResponse response) {
             exception = new AuthorizationException(message, exception);
             break;
          case 404:
-            if (!command.getCurrentRequest().getMethod().equals("DELETE")) {
-               exception = new ResourceNotFoundException(message, exception);
-            }
+            // CloudControl uses error code 400 with RESOURCE_NOT_FOUND to report missing assets
+            // 404 means malformed URI only
+            exception = new IllegalStateException(message, exception);
```
nacx: A malformed URL in a request is better represented as an `IllegalArgumentException`. Illegal sates are more for 409 (Conflict) errors.

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/handlers/ResponseFallbacks.java
```diff
@@ -0,0 +1,46 @@
...
+
+   public static final class VoidOnResourceNotFound implements Fallback<Void> {
+      public Void createOrPropagate(Throwable t) {
+         if (t instanceof ResourceNotFoundException) {
+            return null;
+         } else {
+            throw Throwables.propagate(t);
+         }
+      }
+   }
```
nacx: You'd better copy the [default behavior](https://github.com/jclouds/jclouds/blob/master/core/src/main/java/org/jclouds/Fallbacks.java#L132) but changing the code to 400, to make sure you just capture the right responses.

dimensiondata/src/test/java/org/jclouds/dimensiondata/cloudcontrol/features/ServerApiMockTest.java
```diff
@@ -121,10 +121,9 @@ public void testListServersWithDatacenterFiltering() throws Exception {
       return Uris.uriBuilder("/caas/2.4/6ac1e746-b1ea-4da5-a24e-caf1a978789d/server/server");
    }
 
-   public void testListServers_404() throws Exception {
-      server.enqueue(response404());
-      assertTrue(serverApi().listServers().concat().isEmpty());
-      assertSent(HttpMethod.GET, getListServerUriBuilder().toString());
+   public void testListServers_NoServersFound() {
+      server.enqueue(emptyListResponse("server"));
```
nacx: This test does not exercise the fallback, as it returns a 200 response. Change it to return a 400 error code (you'll need the fallbacks for the empty list mentioned previously).

dimensiondata/src/test/java/org/jclouds/dimensiondata/cloudcontrol/features/ServerApiMockTest.java
```diff
@@ -135,10 +134,10 @@ public void testGetServer() throws Exception {
...
    }
 
-   public void testGetServer_404() throws Exception {
-      server.enqueue(response404());
-      serverApi().getServer("12345");
-      assertSent(GET, "/caas/2.4/6ac1e746-b1ea-4da5-a24e-caf1a978789d/server/server/12345");
+   public void testGetServer_NotFound() {
+      server.enqueue(responseResourceNotFound());
+      Server foundServer = serverApi().getServer("12345");
+      assertNull(foundServer, "should return null when Server was not found");
```
nacx: Same here. Get back the test that exercises the 400 fallback.

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/features/ServerApi.java
FileIOUtility: EmptyIterableWithMarkerOnNotFoundOr404 only affects responses with 404 code or triggering ResourceNotFoundException, none of which if possible for CloudControl list APIs.
This means that previously the method was failing (as expected) for all 4xx responses, ie. there is no behavior change.

dimensiondata/src/test/java/org/jclouds/dimensiondata/cloudcontrol/features/ServerApiMockTest.java
FileIOUtility: Yes, this test is testing a 200 response. This was intended as an empty list response test.
404 is not possible, as previously mentioned, hence the rewrite.

For error code 400 - there is already testDeployServerReturnsUnexpectedError
(Actually it is a preexisting bug in DimensionDataCloudControlErrorHandler that responseCode = UNEXPECTED_ERROR results in ResourceNotFoundException)
Maybe we need 400 test for list and get Server. I can add these, but I am not sure if it belongs here.

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/handlers/ResponseFallbacks.java
FileIOUtility: I see now that NullOnNotFoundOr404 / VoidOnNotFoundOr404 might do the job here (they also trigger on ResourceNotFoundException). Let me check, maybe I can revert to using them.

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/handlers/DimensionDataCloudControlErrorHandler.java
FileIOUtility: thanks, done

dimensiondata/src/test/java/org/jclouds/dimensiondata/cloudcontrol/features/ServerApiMockTest.java
nacx: It is OK now that there is no fallback defined for those methods.

FileIOUtility:
Is it OK to squash these 3 commits before you merge?

nacx:
Thanks, @FileIOUtility! LGTM

From the discussion about response codes, I've revisited the error handler and there are some errors we could consider propagating better:

```java
case 400:
   if (message.contains("RESOURCE_NOT_FOUND")
       || message.contains("OPERATION_NOT_SUPPORTED")) {
      exception = new ResourceNotFoundException(message, exception);
   } else if (message.contains("INVALID_INPUT_DATA")
              || message.contains("ORGANIZATION_NOT_VERIFIED")
              || (message.contains("SYSTEM_ERROR")
                  && !message.contains("RETRYABLE_SYSTEM_ERROR"))
              || message.contains("CPU_SPEED_NOT_AVAILABLE")
              || message.contains("CONFIGURATION_NOT_SUPPORTED")) {
      exception = new IllegalStateException(message, exception);
   } else if (message.contains("RESOURCE_BUSY")
              || message.contains("UNEXPECTED_ERROR")) {
      exception = new ResourceNotFoundException(message, exception);
   } else if (message.contains("NAME_NOT_UNIQUE")) {
      exception = new ResourceAlreadyExistsException(message, exception);
   }
   break;
```
I'd say it makes more sense to remap the following error codes:

* `RESOURCE_BUSY` -> `IllegalStateException`
* `UNEXPECTED_ERROR`, `SYSTEM_ERROR` and `RETRYABLE_SYSTEM_ERROR` -> We don't know what's going on. Let's keep the default `HttpResponseException` instead of remapping them.
* `OPERATION_NOT_SUPPORTED` -> An `UnsupportedOperationException` looks better?
* `INVALID_INPUT_DATA`, `CPU_SPEED_NOT_AVAILABLE`, `CONFIGURATION_NOT_SUPPORTED` -> `IllegalArgumentException`

Since this PR is about fixinf the handling of an error code, I think we could add these changes in this PR. WDYT?

FileIOUtility:
I think this is correct.
I will add this as 4th commit (before squashing).
RESOURCE_BUSY - Another operation is in progress on an asset. Could alternatively be mapped to ConcurrentModificationException. But it is not on the list of exceptions allowed to propagate.

nacx:
>RESOURCE_BUSY - Another operation is in progress on an asset. Could alternatively be mapped to ConcurrentModificationException. But it is not on the list of exceptions allowed to propagate.

Looks like a good candidate to be added to the list. Wanna add it too and use that exception here?

FileIOUtility:
Yes, I will add it to Throwables2.

dimensiondata/src/main/java/org/jclouds/dimensiondata/cloudcontrol/handlers/DimensionDataCloudControlErrorHandler.java
```diff
@@ -47,16 +47,20 @@ public void handleError(HttpCommand command, HttpResponse response) {
...
-            } else if (message.contains("RESOURCE_BUSY") || message.contains("UNEXPECTED_ERROR")) {
-               exception = new ResourceNotFoundException(message, exception);
             } else if (message.contains("NAME_NOT_UNIQUE")) {
                exception = new ResourceAlreadyExistsException(message, exception);
+            } else if (message.contains("UNEXPECTED_ERROR")
+                  || message.contains("RETRYABLE_SYSTEM_ERROR")
+                  || message.contains("SYSTEM_ERROR")) {
+               break;
+            } else {
+               exception = new IllegalArgumentException(message, exception);
```
FileIOUtility: There are too many responseCode value to list here. Rather than that, it is better to default to IllegalArgumentException

dimensiondata/src/test/java/org/jclouds/dimensiondata/cloudcontrol/features/ServerApiMockTest.java
```diff
@@ -56,7 +57,7 @@ public void testDeployServerReturnsUnexpectedError() throws InterruptedException
          serverApi().deployServer(ServerApiMockTest.class.getSimpleName(), "imageId", true, networkInfo,
                "administratorPassword");
          failBecauseExceptionWasNotThrown(ResourceNotFoundException.class);
-      } catch (ResourceNotFoundException e) {
+      } catch (HttpResponseException e) {
```
nacx: Why does this need to be changed?

nacx:
>EDIT: in Java language level 6, ConcurrentModificationException cannot take cause exception, so after all, I don't want to use it.

jclouds needs at least Java 7 to work. We've been trying to keep backward compatibility in our builds, but we are about to move to Java 8 in the next major release. I don't think it makes sense to keep worrying about Java 6 compat.

dimensiondata/src/test/java/org/jclouds/dimensiondata/cloudcontrol/features/ServerApiMockTest.java
FileIOUtility: On UNEXPECTED_ERROR, deployServer API will now throw HttpResponseException.
ResourceNotFound did not make any sense in that case.
HttpResponseException means that we don't know that is going on - error was unexpected.

FileIOUtility:
FYI: build was aborted during "jclouds jdbc core" phase.

dimensiondata/src/test/java/org/jclouds/dimensiondata/cloudcontrol/features/ServerApiMockTest.java
nacx: Then the assertion just before the `catch` needs to be modified, right?

nacx:
rebuild please

FileIOUtility:
build failed again after I committed the suggested change that needs at least Java 7 to work:

"Undefined reference: void java.util.ConcurrentModificationException.<init>(String, Throwable)"
https://jclouds.ci.cloudbees.com/job/jclouds-labs-pull-requests/1994/org.apache.jclouds.labs$dimensiondata-cloudcontrol/consoleFull

nacx:
It's because the default build config. No need to close the PR. I'll bump the source and target versions in the build config and it should be fine.

nacx:
FTR: https://github.com/jclouds/jclouds/pull/1224 should fix this build once merged.

There is one checkstyle violation in this branch though:
```
[WARNING] src/test/java/org/jclouds/dimensiondata/cloudcontrol/features/ServerApiMockTest.java[31:8] (imports) UnusedImports: Unused import - org.jclouds.rest.ResourceNotFoundException.
```

nacx:
Everything should be in place now in jclouds-core: rebuild please

nacx:
And build is back to green! 😄 @FileIOUtility mind squashing the commits so I can cleanly merge the PR?

nacx:
No need for a new PR. Just squash and push+force here.

nacx:
merged to master as [2d9cb407](http://git-wip-us.apache.org/repos/asf/jclouds-labs/commit/2d9cb407). Thanks @FileIOUtility!

PR #449, opened 2018-10-15
Remove deprecated Azure provider

nacx:
Azure ARM has been there for a while and it does not make sense to keep this in labs.


nacx:
Testing new PR build

PR #452, opened 2019-01-11
Use State enum instead of String state in Domain classes

btrishkin:


nacx:
Merged to master and 2.1.x. Thanks! See; https://github.com/apache/jclouds-labs

